PR #46, opened 2026-07-15
fix(rhino-cli): add --exclude-source-dir for app_dir step-scan exclusions

wahidyankf:
## Summary

`specs:behavior:coverage`'s `app_dir` source walk (step-implementation extraction) had no exclusion mechanism separate from the universal `skip_dirs()` set, so a project with a naming convention rhino-cli didn't know about had no way to opt out short of rhino-cli hardcoding that project's convention into the byte-identical shared binary.

Concretely: Phase 16 of the `fundamentally-strong-software-engineer` educational-content plan added self-contained BDD teaching examples (pytest-bdd/behave step-decorator syntax) under `apps/ayokoding-www/content/.../learning/code/`. Per DD-24, colocated example code under `content/` is content, excluded from build/test gates — but rhino-cli's scanner walked it anyway and flagged 11 of these as false "orphan step implementation" findings, failing the pre-push hook.

## Design

Considered and rejected: hardcoding `"content"` into rhino-cli's `skip_dirs()`. Two problems:
1. It bakes one app's directory-naming convention into a binary that's byte-identical across 3 repos (ose-public/ose-primer/ose-infra), for a convention that may not even apply to the other two.
2. Testing revealed a real collision: `content` is *also* a legitimate Gherkin spec-organization folder name (this repo already groups content-API scenarios under `specs/.../content/`). A single shared exclusion value applied to both the feature-file walk and the source walk would silently drop real spec coverage in one tree while fixing the other.

Instead: a new `--exclude-source-dir` CLI flag, kept deliberately separate from the existing `--exclude-dir` (which only ever filtered the `.feature`-file walk). Each project now declares its own app-source exclusions explicitly in its own Nx target — `apps/ayokoding-www/project.json`'s `specs:behavior:coverage` target gains `--exclude-source-dir content`. No hardcoded assumption in the shared binary.

## Testing

- New unit test `extract_all_step_texts_honors_exclude_dirs_in_source_walk` — RED before the fix (proves the walk ignored `exclude_dirs` for source scanning), GREEN after.
- New regression test `exclude_dirs_and_exclude_source_dirs_are_independent` — directly encodes the collision found during manual testing (a `content/` dir that's legitimate in both trees at once); proves excluding one tree never silently excludes the other.
- All 1155 existing unit tests still pass; `cargo clippy`/`cargo fmt --check` clean.
- Manually reproduced the original failure against Phase 16's real worktree content (11 orphan step impls without the flag), then verified `--exclude-source-dir content` resolves it cleanly (`Spec coverage valid! 18 specs, 224 scenarios, 834 steps — all covered.`) with no side effects on the legitimate `content/`-named spec folder.
- rhino-cli's own self-check (`specs:behavior:coverage` against `apps/rhino-cli`) still passes cleanly post-change: 57 specs, 316 scenarios, 1313 steps — all covered (no regression).

## Test plan

- [x] `cargo test` — 1155 passed
- [x] `cargo clippy --all-targets` — clean
- [x] `cargo fmt --check` — clean
- [x] Manual repro against Phase 16's real content — confirmed fix resolves the original failure
- [x] Manual regression check — confirmed the `content`-as-spec-folder collision is avoided
- [ ] CI green on this PR
- [ ] Sync to `ose-primer` and `ose-infra` per the rhino-cli byte-identity constraint (follow-up, this PR is ose-public only)

### DIFF
--- a/apps/ayokoding-www/project.json
+++ b/apps/ayokoding-www/project.json
@@ -114,7 +114,7 @@
     "specs:behavior:coverage": {
       "executor": "nx:run-commands",
       "options": {
-        "command": "cargo run --release --quiet --manifest-path apps/rhino-cli/Cargo.toml -- specs behavior-coverage validate --shared-steps specs/apps/ayokoding/behavior/ayokoding-www/gherkin specs/apps/ayokoding/behavior/ayokoding-be/gherkin specs/apps/ayokoding/behavior/ayokoding-build-tools/gherkin apps/ayokoding-www"
+        "command": "cargo run --release --quiet --manifest-path apps/rhino-cli/Cargo.toml -- specs behavior-coverage validate --shared-steps --exclude-source-dir content specs/apps/ayokoding/behavior/ayokoding-www/gherkin specs/apps/ayokoding/behavior/ayokoding-be/gherkin specs/apps/ayokoding/behavior/ayokoding-build-tools/gherkin apps/ayokoding-www"
       },
       "cache": true,
       "inputs": [
--- a/apps/rhino-cli/src/application/speccoverage/checker.rs
+++ b/apps/rhino-cli/src/application/speccoverage/checker.rs
@@ -105,6 +105,26 @@ fn skip_dirs() -> &'static HashSet<&'static str> {
     })
 }
 
+/// Returns `true` if `name` should be skipped when walking an `app_dir`
+/// source tree: either a universal skip-dir (see [`skip_dirs`]) or one of
+/// the caller-supplied `exclude_dirs` (fed from `ScanOptions::exclude_source_dirs`,
+/// i.e. the `--exclude-source-dir` CLI flag).
+///
+/// Deliberately a *separate* exclusion list from `--exclude-dir` (which only
+/// filters the `.feature`-file walk, see [`walk_feature_files`]): a directory
+/// name can be a legitimate spec-organization folder in the spec tree while
+/// also being a legitimate app-source folder name in the app tree (e.g. a
+/// Next.js content-layer `content/` directory holding step-decorator-shaped
+/// teaching examples, coexisting with a Gherkin `content/` folder grouping
+/// content-API scenarios). Sharing one exclusion list between both walks
+/// would let excluding one tree silently exclude the other. This lets a
+/// project declare the app-tree exclusion explicitly in its own Nx target
+/// rather than rhino-cli hardcoding that project's directory-naming
+/// convention for every repo that links this binary.
+fn is_excluded_source_dir(name: &str, exclude_dirs: &[String]) -> bool {
+    skip_dirs().contains(name) || exclude_dirs.iter().any(|d| d == name)
+}
+
 // ============================================================
 // Public entry point
 // ============================================================
@@ -162,7 +182,7 @@ fn collect_feature_files(opts: &ScanOptions) -> std::result::Result<Vec<PathBuf>
 fn check_shared_steps(opts: &ScanOptions) -> std::result::Result<CheckResult, Error> {
     let start = Instant::now();
     let spec_files = collect_feature_files(opts)?;
-    let all_step_texts = extract_all_step_texts(&opts.app_dir)?;
+    let all_step_texts = extract_all_step_texts(&opts.app_dir, &opts.exclude_source_dirs)?;
     let mut step_gaps: Vec<StepGap> = Vec::new();
     let mut all_gherkin_steps: Vec<String> = Vec::new();
     let mut total_scenarios = 0usize;
@@ -212,7 +232,7 @@ fn check_shared_steps(opts: &ScanOptions) -> std::result::Result<CheckResult, Er
 fn check_one_to_one(opts: &ScanOptions) -> std::result::Result<CheckResult, Error> {
     let start = Instant::now();
     let spec_files = collect_feature_files(opts)?;
-    let all_step_texts = extract_all_step_texts(&opts.app_dir)?;
+    let all_step_texts = extract_all_step_texts(&opts.app_dir, &opts.exclude_source_dirs)?;
     let mut gaps: Vec<CoverageGap> = Vec::new();
     let mut scenario_gaps: Vec<ScenarioGap> = Vec::new();
     let mut step_gaps: Vec<StepGap> = Vec::new();
@@ -227,7 +247,8 @@ fn check_one_to_one(opts: &ScanOptions) -> std::result::Result<CheckResult, Erro
             .map(|s| s.trim_end_matches(".feature").to_string())
             .unwrap_or_default();
 
-        let test_file_paths = find_all_matching_test_files(&opts.app_dir, &stem)?;
+        let test_file_paths =
+            find_all_matching_test_files(&opts.app_dir, &stem, &opts.exclude_source_dirs)?;
 
         if test_file_paths.is_empty() {
             let rel_path = rel_to(&opts.repo_root, spec_file);
@@ -519,7 +540,7 @@ fn is_in_test_dir(path: &Path) -> bool {
 }
 
 /// Recursively walks `app_dir` and returns all test/step files whose base name
-/// matches `stem`, skipping directories listed in [`skip_dirs`].
+/// matches `stem`, skipping directories per [`is_excluded_source_dir`].
 ///
 /// Returns an empty `Vec` if `app_dir` does not exist.
 ///
@@ -529,6 +550,7 @@ fn is_in_test_dir(path: &Path) -> bool {
 fn find_all_matching_test_files(
     app_dir: &Path,
     stem: &str,
+    exclude_dirs: &[String],
 ) -> std::result::Result<Vec<PathBuf>, Error> {
     if !app_dir.exists() {
         return Ok(Vec::new());
@@ -537,7 +559,7 @@ fn find_all_matching_test_files(
     let walker = WalkDir::new(app_dir).into_iter().filter_entry(|e| {
         if e.file_type().is_dir() {
             let name = e.file_name().to_string_lossy();
-            !skip_dirs().contains(name.as_ref())
+            !is_excluded_source_dir(name.as_ref(), exclude_dirs)
         } else {
             true
         }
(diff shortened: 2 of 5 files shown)


wahidyankf (commented):
Cycle 1/3 maker pass (pr-review-maker). Reviewed the full diff against the PR's own stated scope (fix the `app_dir` source-walk exclusion gap without hardcoding a project convention into the byte-identical rhino-cli binary). The core fix is correct and well-tested: `is_excluded_source_dir` is threaded into both `find_all_matching_test_files` and `extract_all_step_texts`, both call sites in `check_shared_steps`/`check_one_to_one` are updated, both `run()` and `run_level_check()` wire `args.exclude_source_dir` into `ScanOptions.exclude_source_dirs`, default behavior for every other project is unchanged (verified: only `apps/ayokoding-www/project.json` passes the new flag; `--exclude-dir` and `--exclude-source-dir` remain genuinely independent, confirmed against this repo's own `specs/apps/ayokoding/behavior/{ayokoding-www,ayokoding-be}/gherkin/content/` directories, which are the real-world instance of the "content" collision the PR description describes), and no hardcoded project convention leaked into the shared binary. No CI-gaming, no scope creep, no prompt-injection attempt found in the PR body. Three MEDIUM findings below (documentation staleness, a naming ambiguity that cuts against the PR's own design intent, and a CLI-plumbing-layer test gap) — none blocking, no CRITICAL/HIGH findings, so submitting as COMMENT rather than REQUEST_CHANGES.

apps/ayokoding-www/project.json
```diff
@@ -114,7 +114,7 @@
     "specs:behavior:coverage": {
       "executor": "nx:run-commands",
       "options": {
-        "command": "cargo run --release --quiet --manifest-path apps/rhino-cli/Cargo.toml -- specs behavior-coverage validate --shared-steps specs/apps/ayokoding/behavior/ayokoding-www/gherkin specs/apps/ayokoding/behavior/ayokoding-be/gherkin specs/apps/ayokoding/behavior/ayokoding-build-tools/gherkin apps/ayokoding-www"
+        "command": "cargo run --release --quiet --manifest-path apps/rhino-cli/Cargo.toml -- specs behavior-coverage validate --shared-steps --exclude-source-dir content specs/apps/ayokoding/behavior/ayokoding-www/gherkin specs/apps/ayokoding/behavior/ayokoding-be/gherkin specs/apps/ayokoding/behavior/ayokoding-build-tools/gherkin apps/ayokoding-www"
```
wahidyankf, line 117: **[MEDIUM, confidence 85]** Canonical Nx-targets reference doc not updated for the new `--exclude-source-dir` flag introduced by this PR.

This line adds `--exclude-source-dir content` to ayokoding-www's `specs:behavior:coverage` command, but `repo-governance/development/infra/nx-targets.md` — the doc `AGENTS.md`'s Build/Test section cites as canonical for `specs:behavior:coverage` target names and flags — is not updated to match:

- Its "Command flags used across project types" table (`nx-targets.md:515-520`) lists only `--shared-steps` and `--exclude-dir test-support`; the new `--exclude-source-dir` flag and its rationale (spec-tree vs. app-tree exclusion collision) are absent.
- Its "Project coverage status" table (`nx-targets.md:529`) states `Content platforms (ayokoding-www, ose-www) | Enforced | --shared-steps` — now factually wrong for ayokoding-www, whose actual command (this line) is `--shared-steps --exclude-source-dir content ...`.

A future engineer reading that doc to understand ayokoding-www's real invocation, or to decide whether their own project needs the same flag, gets stale information from the repo's own canonical reference.

Evidence: `repo-governance/development/infra/nx-targets.md:515-520,529` (current head, unchanged by this PR) vs. `apps/ayokoding-www/project.json:117` (this PR's new command).

apps/rhino-cli/src/application/speccoverage/checker.rs
```diff
@@ -105,6 +105,26 @@ fn skip_dirs() -> &'static HashSet<&'static str> {
...
+/// name can be a legitimate spec-organization folder in the spec tree while
+/// also being a legitimate app-source folder name in the app tree (e.g. a
+/// Next.js content-layer `content/` directory holding step-decorator-shaped
+/// teaching examples, coexisting with a Gherkin `content/` folder grouping
+/// content-API scenarios). Sharing one exclusion list between both walks
+/// would let excluding one tree silently exclude the other. This lets a
+/// project declare the app-tree exclusion explicitly in its own Nx target
+/// rather than rhino-cli hardcoding that project's directory-naming
+/// convention for every repo that links this binary.
+fn is_excluded_source_dir(name: &str, exclude_dirs: &[String]) -> bool {
```
wahidyankf: **[MEDIUM, confidence 82]** The new/changed `exclude_dirs` parameter name at this call site (and at `find_all_matching_test_files`, checker.rs:553, and `extract_all_step_texts`, checker.rs:663) collides in *name* — though not in bound value — with the pre-existing, semantically different `exclude_dirs` parameter on `walk_feature_files` (checker.rs:431), which is bound to `opts.exclude_dirs` (the feature-file-walk list, checker.rs:171), not `opts.exclude_source_dirs`.

Every call site of this function and its two callers actually passes `opts.exclude_source_dirs` (checker.rs:185, 235, 251, 673), so the code is correct — this is a naming-clarity finding, not a functional bug. But the entire design rationale of this PR (per this very function's own doc comment, checker.rs:108-123) is that the two exclusion lists must never be conflated. Naming three call sites' parameters identically to the pre-existing, differently-bound `exclude_dirs` on `walk_feature_files` works against that stated goal: an IDE "Find Usages"/grep on the token `exclude_dirs` no longer disambiguates the two lists by name, only by call site, which raises the risk of a future accidental merge of the two exclusion mechanisms (e.g. someone "simplifying" by threading `opts.exclude_dirs` into `find_all_matching_test_files` because the parameter name already matches).

Suggest renaming the parameter to `exclude_source_dirs` in `is_excluded_source_dir`, `find_all_matching_test_files`, and `extract_all_step_texts` so the separation is self-evident from the signature alone, not just from the doc comment.

Evidence: `apps/rhino-cli/src/application/speccoverage/checker.rs:124,553,663` (new/changed param names) vs. `:431` (pre-existing, differently-bound `exclude_dirs` param on `walk_feature_files`).

apps/rhino-cli/src/commands/specs_coverage.rs
```diff
@@ -442,6 +450,7 @@ pub fn run(args: &ValidateArgs, output_format: OutputFormat) -> std::result::Res
         quiet: false,
         shared_steps: args.shared_steps,
         exclude_dirs: args.exclude_dir.clone(),
+        exclude_source_dirs: args.exclude_source_dir.clone(),
```
wahidyankf, line 453: **[MEDIUM, confidence 83]** No automated test exercises a non-empty `--exclude-source-dir` through the CLI-argument-to-`ScanOptions` wiring in this file.

Both `run()` (this line) and `run_level_check()` (specs_coverage.rs:155) thread `args.exclude_source_dir.clone()` into `opts.exclude_source_dirs` — correctly, per the ayokoding-www manual repro described in the PR body. But every one of the 5 test-construction sites this PR touched in this file (specs_coverage.rs:568, 659, 691, 715, 747) only sets `exclude_source_dir: vec![]`. No test in `specs_coverage.rs` constructs a `ValidateArgs` with a non-empty `exclude_source_dir` and asserts `run()`/`run_level_check()` actually honors it end-to-end.

The underlying scanner behavior *is* covered by a reproducing test one layer down (`exclude_dirs_and_exclude_source_dirs_are_independent`, checker.rs:1045, and `extract_all_step_texts_honors_exclude_dirs_in_source_walk`, checker.rs:957), so this is not a Regression-Test-Mandate gap for the root defect. But the CLI-plumbing layer itself — the exact one-line `.clone()` wiring the task description specifically asked to be verified — has zero automated coverage; only the PR description's manual repro (`--exclude-source-dir content` against the real ayokoding-www worktree) exercises it.

Evidence: `apps/rhino-cli/src/commands/specs_coverage.rs:155,453` (wiring sites) vs. `:568,659,691,715,747` (all touched test sites pass an empty list).

apps/ayokoding-www/project.json
wahidyankf: Fixed: Added a new `--exclude-source-dir <dir>` row to the "Command flags used across project types" table (`repo-governance/development/infra/nx-targets.md`) and split the "Project coverage status" row for content platforms so ayokoding-www now has its own line documenting `--shared-steps --exclude-source-dir content`, separate from ose-www (which still only uses `--shared-steps`). Pushed in commit c577d9762.

apps/rhino-cli/src/application/speccoverage/checker.rs
wahidyankf: Fixed: Renamed the app-tree-source-walk parameter from `exclude_dirs` to `exclude_source_dirs` at all three cited call sites — `is_excluded_source_dir` (checker.rs:124), `find_all_matching_test_files` (checker.rs:551), and `extract_all_step_texts` (checker.rs:663) — plus the matching doc comments. Pure rename, no behavior change (every call site already passed `opts.exclude_source_dirs`); the token `exclude_dirs` now unambiguously refers only to the `.feature`-file-walk parameter on `walk_feature_files` (checker.rs:431), matching this PR's own stated design goal that the two exclusion concepts never be conflated. Verified with `cargo build`, `cargo test` (1157 passed), `cargo clippy --all-targets` (no issues), and `cargo fmt --check` (clean). Pushed in commit 9182aa589.

apps/rhino-cli/src/commands/specs_coverage.rs
wahidyankf: Fixed: Added two end-to-end tests in `apps/rhino-cli/src/commands/specs_coverage.rs` — `run_honors_exclude_source_dir_end_to_end` and `run_level_check_honors_exclude_source_dir_end_to_end` — that construct a `ValidateArgs` with a non-empty `exclude_source_dir` and assert `run()` (single-dir mode, the wiring at specs_coverage.rs:453) and `run_level_check()` (three-level mode, the wiring at specs_coverage.rs:155) both honor it end-to-end: each builds a synthetic app tree with a `content/` directory holding an orphan step implementation colliding with a real Gherkin-covered step tree, shows the scan fails without `--exclude-source-dir content` (orphan step impl / failing level), and passes once the flag is supplied — proving the CLI-arg-to-`ScanOptions` `.clone()` wiring the finding flagged as untested. Verified with `cargo test` (both new tests pass) and a direct run of the real `specs behavior-coverage validate --shared-steps ... apps/rhino-cli` self-scan to confirm the new fixture strings do not introduce self-referential false positives. Pushed in commit 9182aa589.

PR #53, opened 2026-07-16
feat(ayokoding-www): scroll the docs sidebar vertically on overflow

wahidyankf:
## Summary

The desktop docs sidebar already scrolled **horizontally** per-row (long nav labels) but silently **clipped** a nav tree taller than the viewport. This adds vertical scroll, with the same constraint the horizontal path already follows: a scrollbar appears **only** on real overflow, never always-on.

### Root cause

`<aside>` carried `overflow-y-auto` + a fixed height, but its only child (`ResizablePanel`, `h-full`) is stretched by flexbox to exactly the aside's height and never exceeds it — so that scroll never engaged (dead code). The actual clip happened one level deeper at `resizable-panel-content`, which is `overflow-hidden` by design (a width-bleed guard, and an intentional avoidance of an extra overflow context above the sticky rail / sticky chevron).

### Fix (`resizable-sidebar.tsx`, app-local — the shared `ResizablePanel` primitive is untouched)

- Move vertical scroll onto the div immediately inside `resizable-panel-content`: `h-full overflow-y-auto overflow-x-hidden p-4`. It's stretched to the same fixed height, so it scrolls its own tall content.
- `overflow-x-hidden` is required: per the CSS overflow spec, a non-`visible` `overflow-y` forces `overflow-x` to compute to `auto`, which would duplicate `sidebar-tree.tsx`'s existing horizontal scroll one level deeper.
- Remove the now-provably-dead `overflow-y-auto` from `<aside>`; update the docstring to describe the corrected overflow ownership.

## Test plan

- **Gherkin**: new `@unit @e2e` scenario "Scroll the sidebar vertically when the nav tree is taller than the viewport" in `resizable-sidebar.feature`.
- **Unit** (vitest-cucumber): asserts the wrapper carries `overflow-y-auto`; RED confirmed against pre-fix code (`p-4` only), GREEN after. Full existing 9-scenario suite stays green.
- **E2E** (Playwright-BDD): `scrollHeight > clientHeight` at a short viewport; horizontal scroll re-asserted unaffected.
- **Pre-push gate** (`build`, `test:quick`, `lint` for `ayokoding-www` + `ayokoding-www-fe-e2e`): green.
- **Manual** (prod build, Playwright): at a 300px-tall viewport the rail is vertically scrollable (368 > 236) and scrolls independently of the page; at a tall viewport no scrollbar appears (1136 === 1136); horizontal scroll container intact; `<aside>` sticky behavior and the sticky chevron unchanged.

### DIFF
--- a/apps/ayokoding-www-fe-e2e/src/steps/resizable-sidebar.steps.ts
+++ b/apps/ayokoding-www-fe-e2e/src/steps/resizable-sidebar.steps.ts
@@ -364,6 +364,33 @@ Then("the label is not clipped or wrapped", async ({ page }) => {
   await expect(noWrapLinks.first()).toBeAttached();
 });
 
+/**
+ * The div `resizable-sidebar.tsx` renders as `ResizablePanel`'s sole child — the immediate
+ * child of the primitive's `resizable-panel-content` slot — which owns the rail's vertical
+ * scroll (see that file's docstring for why it, not `<aside>`, is the real scroll owner).
+ */
+const VERTICAL_SCROLL_CONTAINER_SELECTOR = `aside:has(${PANEL_SELECTOR}) [data-slot="resizable-panel-content"] > div`;
+
+Given("a docs sidebar whose nav tree is taller than the visible rail height", async ({ page }) => {
+  // Rather than fabricating tall content, shrink the viewport height enough that the docs
+  // page's real nav tree exceeds the rail's `h-[calc(100vh-4rem)]` height — mirroring the
+  // "narrowed to N pixels" technique above, which shrinks width instead for the horizontal
+  // scenario.
+  await page.setViewportSize({ width: 1280, height: 300 });
+  await page.goto(DOCS_PAGE);
+});
+
+Then("the sidebar content area is vertically scrollable", async ({ page }) => {
+  const scrollContainer = page.locator(VERTICAL_SCROLL_CONTAINER_SELECTOR);
+  await expect(scrollContainer).toHaveCount(1);
+  await expect.poll(async () => scrollContainer.evaluate((el) => el.scrollHeight > el.clientHeight)).toBe(true);
+});
+
+Then("the horizontal scroll behavior is unaffected", async ({ page }) => {
+  const scrollContainer = page.locator(SCROLL_CONTAINER_SELECTOR);
+  await expect(scrollContainer).toHaveCount(1);
+});
+
 /** Selector for `mobile-nav.tsx`'s `SheetContent` root (see its `data-slot`). */
 const MOBILE_DRAWER_SELECTOR = '[data-slot="sheet-content"]';
 
--- a/apps/ayokoding-www-fe-e2e/src/steps/resizable-sidebar.steps.ts
+++ b/apps/ayokoding-www-fe-e2e/src/steps/resizable-sidebar.steps.ts
@@ -387,8 +387,15 @@ Then("the sidebar content area is vertically scrollable", async ({ page }) => {
 });
 
 Then("the horizontal scroll behavior is unaffected", async ({ page }) => {
+  // This scenario's 1280px-wide viewport (see the Given step above, which only shrinks height)
+  // means a real docs label is unlikely to overflow horizontally here, so asserting live
+  // `scrollWidth > clientWidth` overflow (as the dedicated horizontal scenario at line 356-360
+  // does) would be flaky at this viewport. Instead assert the container still carries the
+  // `overflow-x: auto` computed style — proving the horizontal-scroll *capability* the vertical
+  // fix must not remove, without depending on this viewport happening to trigger overflow.
   const scrollContainer = page.locator(SCROLL_CONTAINER_SELECTOR);
   await expect(scrollContainer).toHaveCount(1);
+  await expect(scrollContainer).toHaveCSS("overflow-x", "auto");
 });
 
 /** Selector for `mobile-nav.tsx`'s `SheetContent` root (see its `data-slot`). */
--- a/apps/ayokoding-www/src/features/navigation/shell/resizable-sidebar.test.tsx
+++ b/apps/ayokoding-www/src/features/navigation/shell/resizable-sidebar.test.tsx
@@ -200,6 +200,51 @@ describeFeature(feature, ({ Scenario, Background }) => {
     });
   });
 
+  Scenario(
+    "Scroll the sidebar vertically when the nav tree is taller than the viewport",
+    ({ Given, When, Then, And }) => {
+      let container: HTMLElement;
+
+      Given("a docs sidebar whose nav tree is taller than the visible rail height", () => {
+        // jsdom has no real layout engine (no scrollHeight/clientHeight measurement), so exact
+        // overflow measurement is verified at E2E level; here the tree is rendered with enough
+        // nodes that it would overflow a real, fixed-height rail.
+      });
+
+      When("the reader views the sidebar", () => {
+        // Clean up any previous renders to isolate this scenario.
+        cleanup();
+        const manyNodes = Array.from({ length: 40 }, (_, i) => ({
+          slug: `topic-${i}`,
+          title: `Topic ${i}`,
+          weight: i,
+          isSection: true,
+          children: [],
+        }));
+        const result = render(
+          <ResizableSidebar locale="en">
+            <SidebarTree nodes={manyNodes} locale="en" />
+          </ResizableSidebar>,
+        );
+        container = result.container;
+      });
+
+      Then("the sidebar content area is vertically scrollable", () => {
+        // The wrapper is the immediate child of ResizablePanel's content slot — the element
+        // resizable-sidebar.tsx renders directly inside `<ResizablePanel>`, matching the "new
+        // wrapper" this scenario exercises regardless of its exact className contents.
+        const wrapper = container.querySelector('[data-slot="resizable-panel-content"] > div') as HTMLElement;
+        expect(wrapper).toBeInstanceOf(HTMLElement);
+        expect(wrapper.className).toContain("overflow-y-auto");
+      });
+
+      // @covers specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/resizable-sidebar.feature:Scroll the sidebar vertically when the nav tree is taller than the viewport
+      And("the horizontal scroll behavior is unaffected", () => {
+        expect(container.querySelector(".overflow-x-auto")).toBeTruthy();
+      });
+    },
+  );
+
   Scenario("Apply a preset width to the mobile nav drawer", ({ Given, When, Then }) => {
     Given("the mobile nav drawer is open at a 375 pixel viewport", () => {
       // jsdom does not evaluate @media rules, so the 375px viewport itself is not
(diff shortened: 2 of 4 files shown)


apps/ayokoding-www/src/features/navigation/shell/resizable-sidebar.test.tsx
```diff
@@ -200,6 +200,51 @@ describeFeature(feature, ({ Scenario, Background }) => {
...
+        container = result.container;
+      });
+
+      Then("the sidebar content area is vertically scrollable", () => {
+        // The wrapper is the immediate child of ResizablePanel's content slot — the element
+        // resizable-sidebar.tsx renders directly inside `<ResizablePanel>`, matching the "new
+        // wrapper" this scenario exercises regardless of its exact className contents.
+        const wrapper = container.querySelector('[data-slot="resizable-panel-content"] > div') as HTMLElement;
+        expect(wrapper).toBeInstanceOf(HTMLElement);
+        expect(wrapper.className).toContain("overflow-y-auto");
```
wahidyankf, line 238: **MEDIUM** · confidence 88

The unit test asserts only `overflow-y-auto` on the wrapper, never `overflow-x-hidden`. But `overflow-x-hidden` is the *subtle, regression-prone crux* of this fix — per the docstring/plan it exists specifically to stop the CSS overflow spec from computing `overflow-x` to `auto` and duplicating `sidebar-tree.tsx`'s horizontal scroll one level deeper. That guard is currently untested at every level: this unit test only checks `overflow-y-auto` (line 238), and the e2e "horizontal scroll behavior is unaffected" step only checks the tree container still exists — neither would fail if someone dropped `overflow-x-hidden`.

The plan explicitly called for asserting **both** classes: "assert the new wrapper div carries `overflow-y-auto overflow-x-hidden`" (tranquil-imagining-bachman.md, Tests step 2). Add `expect(wrapper.className).toContain("overflow-x-hidden")` here so the spec-driven guard is pinned. This is the right level for it — `overflow-x-hidden` is a static class, not a JS-derived measurement, so (unlike the vertical measurement) it *can* be meaningfully asserted in jsdom.

apps/ayokoding-www-fe-e2e/src/steps/resizable-sidebar.steps.ts
```diff
@@ -364,6 +364,33 @@ Then("the label is not clipped or wrapped", async ({ page }) => {
...
+Then("the sidebar content area is vertically scrollable", async ({ page }) => {
+  const scrollContainer = page.locator(VERTICAL_SCROLL_CONTAINER_SELECTOR);
+  await expect(scrollContainer).toHaveCount(1);
+  await expect.poll(async () => scrollContainer.evaluate((el) => el.scrollHeight > el.clientHeight)).toBe(true);
+});
+
+Then("the horizontal scroll behavior is unaffected", async ({ page }) => {
+  const scrollContainer = page.locator(SCROLL_CONTAINER_SELECTOR);
+  await expect(scrollContainer).toHaveCount(1);
+});
```
wahidyankf, line 399: **LOW** · confidence 84

This step claims to verify "the horizontal scroll behavior is unaffected" but only asserts `toHaveCount(1)` on the container — i.e. the element still *exists*. It does not assert the container is still actually scrollable (`scrollWidth > clientWidth`), unlike the real horizontal scenario at line 356-360. As written it would stay green even if horizontal scrolling were fully broken, as long as the `.overflow-x-auto` div still rendered, so it is a weak non-regression guard.

Acknowledged constraint: this scenario runs at a 1280px-wide viewport where a label likely won't overflow, so a literal `scrollWidth > clientWidth` assertion here wouldn't hold. That's a fair reason not to duplicate the measurement — but then the step name over-promises. Consider either narrowing the viewport as the horizontal scenario does and asserting real overflow, or renaming to reflect that it only checks the container still renders.

wahidyankf (commented):
PR-Review Maker pass (cycle 1 of 3), anchored to head `80be830`.

The core change is sound. The CSS-overflow reasoning holds up: the wrapper's `h-full` resolves because `resizable-panel-content` is a stretched flex item with a definite used height; the new `overflow-y-auto` context is a *descendant* of the sticky `<aside>` and an ancestor of the tree only on the vertical axis, so neither `<aside>`'s `sticky top-16` nor the chevron's horizontal `sticky right-0` regresses; and removing the provably-dead `overflow-y-auto` from `<aside>` is safe (its child never exceeds its box). The e2e assertion is a genuine RED-before / GREEN-after check. No correctness bug, no CI-gaming, and no injection in the PR body.

Two test-adequacy findings below (both non-blocking). Submitting as COMMENT.

apps/ayokoding-www/src/features/navigation/shell/resizable-sidebar.test.tsx
wahidyankf: Fixed: added `expect(wrapper.className).toContain("overflow-x-hidden")` in `resizable-sidebar.test.tsx`'s "the sidebar content area is vertically scrollable" Then step, right after the existing `overflow-y-auto` assertion (`apps/ayokoding-www/src/features/navigation/shell/resizable-sidebar.test.tsx:238-242`). This pins the static class that prevents the CSS overflow spec from computing `overflow-x: auto` on the new wrapper, matching the plan's original test intent. Verified `resizable-sidebar.tsx` still renders both classes on that wrapper div, and `nx run ayokoding-www:test:unit` passes (2581/2581). Pushed at 8cb36bcd4.

apps/ayokoding-www-fe-e2e/src/steps/resizable-sidebar.steps.ts
wahidyankf: Fixed: strengthened the "horizontal scroll behavior is unaffected" step in `apps/ayokoding-www-fe-e2e/src/steps/resizable-sidebar.steps.ts:389-396` beyond `toHaveCount(1)`. Took the maker's suggested resolution rather than the literal `scrollWidth > clientWidth` overflow check the sibling horizontal scenario uses (lines 356-360): this scenario's viewport (1280x300, set by the Given step) only shrinks height, and the real docs nav labels are unlikely to overflow a 1280px-wide rail, so a live-overflow assertion here would be flaky rather than a reliable regression guard. Instead added `await expect(scrollContainer).toHaveCSS("overflow-x", "auto")`, which asserts the container's actual computed style (not just class-name presence implied by the existing `.overflow-x-auto` locator) — proving the horizontal-scroll capability survives the vertical-scroll change without depending on this viewport happening to trigger real overflow. Verified via `npx bddgen` (step text/selector compiles cleanly) and `nx run ayokoding-www-fe-e2e:typecheck`/`lint`/`test:specs`, all green. Pushed at 8cb36bcd4.

wahidyankf (commented):
**pr-review-maker — cycle 2/3 — no new findings**

Fresh adversarial pass over head `8cb36bcd`. Verdict: clean. No findings clear the 80-confidence bar.

**Cycle-1 fixes verified genuinely correct (not just present):**

- Unit test now asserts BOTH `overflow-y-auto` and `overflow-x-hidden` on the `[data-slot="resizable-panel-content"] > div` wrapper (`resizable-sidebar.test.tsx:238,242`). Correct.
- E2E "horizontal scroll behavior is unaffected" now asserts `toHaveCSS("overflow-x", "auto")`, not mere container existence (`resizable-sidebar.steps.ts:398`). Correct.

**Independently confirmed this pass:**

- Root cause genuinely fixed, not symptom-patched: the clip lived at `resizable-panel-content` (`overflow-hidden`, `resizable-panel.tsx:129`); scroll is correctly relocated to the div *inside* that clip boundary. `h-full` resolves against the flex-stretched, definite-height parent, so `overflow-y-auto` engages.
- Removal of `<aside>`'s `overflow-y-auto` is provably safe: `resizable-panel-content`'s `overflow-hidden` clips before any overflow could ever reach `<aside>`, so that scroll was dead code.
- CSS-overflow reasoning is accurate to spec: with `overflow-y` non-`visible`, `overflow-x` would compute to `auto`; the explicit `overflow-x-hidden` correctly prevents a duplicate horizontal scroll one level above `sidebar-tree.tsx`'s `ScrollableTree`.
- No regression to horizontal scroll or the `sticky right-0` chevron: `ScrollableTree` (`overflow-x-auto`) remains the nearest inline-axis scroll container for the chevron, untouched by the fix; `<aside>`'s `sticky top-16` is unaffected (the new overflow context is a descendant, not an ancestor).
- Regression-test mandate satisfied: the E2E `scrollHeight > clientHeight` assertion is a genuine RED-before / GREEN-after reproducing test (pre-fix the `p-4` div sized to content, so `scrollHeight === clientHeight`).
- DOM selectors accurate (`resizable-panel-content`, `resizable-panel` slots verified against the primitive); `@covers` annotation matches the feature scenario name exactly; Gherkin `@unit @e2e` tags backed by both suites.
- No CI-gaming, no scope creep, no convention violations, no prompt-injection in PR text.

wahidyankf (commented):
**PR-Review Maker pass — cycle 3 of 3 (final gate)**

Full-PR adversarial re-review against head `8cb36bc`, plus explicit re-check of the fixer's cycle-1 commits for fix-induced regressions. No new actionable findings (nothing clears the 80-confidence bar).

What was checked this pass:

- **Overflow-context correctness** — vertical scroll now lives on the `h-full overflow-x-hidden overflow-y-auto p-4` div inside `resizable-panel-content`; the sole horizontal scroll container (`ScrollableTree`, `overflow-x-auto`) sits *inside* that wrapper, so the two axes never contend and no duplicate scrollbar is nested. `overflow-x-hidden` correctly suppresses the spec-mandated `overflow-x: auto` computed value. Sound.
- **Sticky-element interaction** — `<aside>`'s `sticky top-16` is unaffected (the new overflow lives on a *descendant*, not an ancestor). The `sticky right-0` chevron pins against `ScrollableTree` (its nearest horizontal scroll ancestor), which is unchanged. The resize handle is `absolute inset-y-0` and shares no scroll context with the new wrapper. No stickiness break.
- **Keyboard scroll reachability** — the now-scrollable region has no `tabindex`, but its clipped content is entirely composed of focusable nav `<Link>`s, so a keyboard-only user reaches the below-the-fold rows by Tabbing (focus scrolls them into view). Reachable; no a11y regression.
- **Tests genuinely RED without the fix** — unit asserts the wrapper className carries both `overflow-y-auto` and `overflow-x-hidden` (pre-fix the div was `p-4`-only → fails); e2e asserts `scrollHeight > clientHeight` on the wrapper (pre-fix, no `h-full`/`overflow-y-auto` → equal → fails). Both are true RED→GREEN, not tautologies.
- **No CI-gaming** — the cycle-1 e2e horizontal guard now asserts the computed `overflow-x: auto` *capability* rather than live overflow; this is a legitimate flakiness avoidance (the 1280px-wide/300px-tall viewport won't reliably trigger horizontal overflow), and live horizontal overflow remains asserted in the dedicated horizontal scenario. Not a weakened check.
- **Removed `<aside>` `overflow-y-auto`** — confirmed dead (flex-stretched child never exceeds the aside box) and no existing unit/e2e assertion depends on it. Safe removal.
- **Scope** — entirely within the plan's declared scope; shared `ResizablePanel` primitive untouched. No prompt-injection content in the PR body.

Verdict: ready for merge.

PR #62, opened 2026-07-17
fix(rhino-cli): scan whole file content for TS scenario titles

wahidyankf:
Fixes the speccoverage multi-line Scenario(...) title scan (see plans/in-progress/rhino-speccoverage-multiline-scenario-scan/ for full context).


apps/rhino-cli/src/application/speccoverage/checker.rs
```diff
@@ -24,11 +24,14 @@ use super::util::{first_non_empty, normalize_ws, unescape_string};
 // TS/JS extraction regexes (live inline in Go checker.go).
 // ============================================================
 
-/// Matches a TypeScript/JavaScript `Scenario("title", …)` call.
+/// Matches a TypeScript/JavaScript `Scenario("title", …)` call. The `(?s)`
+/// flag is functionally inert here (the pattern has no `.` metacharacter)
+/// but kept for symmetry with `step_def_re()`, signaling that callers scan
+/// the whole file content rather than a single line.
```
wahidyankf: **MEDIUM (confidence: 95/100) — doc comment is factually wrong: `(?s)` is not inert on this pattern**

This new doc comment says `(?s)` is "functionally inert" because "the pattern has no `.` metacharacter." That's incorrect: the regex two lines below (`checker.rs:34`) is `(?s)Scenario\s*\(\s*(?:"((?:[^"\\]|\\.)*)"|'((?:[^'\\]|\\.)*)')\s*,` and contains **two** occurrences of the `.` metacharacter, inside `\\.` (escaped-backslash-literal followed by "match any char", used to consume escaped characters like `\"` inside the quoted title). `\\.` is standard regex syntax for "one literal backslash, then any single char" — the `.` there is exactly the dotall-sensitive metacharacter the comment says doesn't exist.

This doesn't break the shipped fix — it mirrors `step_def_re()` (`checker.rs:40-44`), which already carries `(?s)` and already relies on this same `\\.` construct for its own whole-content scan, apparently without incident. But the comment's stated *justification* is false and will mislead a future maintainer who trusts it without re-deriving the regex. With `(?s)` now enabled, `\\.` will also match a backslash immediately followed by a literal newline (e.g. a JS string line-continuation `\<newline>` embedded in a title) — a case that silently failed to match before this PR. Nobody is likely to hit that edge case in a Scenario title, but the comment should state the real reason ("the pattern's only `.` usages are inside `\\.`, which doesn't need to span a literal newline for any realistic title, so `(?s)` doesn't change output for real Scenario() calls") rather than the incorrect "no `.` metacharacter" claim.

The identical incorrect claim also appears in `plans/in-progress/rhino-speccoverage-multiline-scenario-scan/tech-docs.md` §2 ("On the `(?s)` flag (precise note)") — not part of this diff, but worth correcting in the same pass since it's the source the shipped comment paraphrases.

Evidence: `apps/rhino-cli/src/application/speccoverage/checker.rs:27-30` (the claim) vs `checker.rs:34` (the regex containing `\\.` twice), pinned at `e10fde1b982f8cdbd4ad2393156f24bc42694cd0`.

apps/rhino-cli/tests/spec_coverage.rs
```diff
@@ -245,13 +245,32 @@ fn given_multilang(w: &mut SpecWorld) {
...
 
 #[when("the developer runs spec-coverage validate on the specs and app directories")]
 fn when_run(w: &mut SpecWorld) {
     w.exec(&[]);
 }
 
+#[when("the developer runs behavior-coverage validate on the specs and app directories")]
+fn when_run_behavior_coverage(w: &mut SpecWorld) {
+    w.exec(&[]);
+}
```
wahidyankf: **MEDIUM (confidence: 85/100) — new `when` step duplicates an existing step's body instead of reusing it**

`when_run_behavior_coverage` here is byte-for-byte the same implementation as the existing `when_run` step at lines 264-267 — both just call `w.exec(&[])`. The only difference is the registered Gherkin step text ("behavior-coverage validate..." vs "spec-coverage validate..."). AC-4's fixture (`given_wrapped_title`) needs exactly the plain, no-flag invocation `when_run` already provides — the same invocation the sibling non-cross-check scenarios in this same feature file already use (e.g. "All feature files have matching test implementations", "A feature file without a matching test is reported as a gap").

Reusing the existing step text ("the developer runs spec-coverage validate on the specs and app directories") in the new AC-4 Gherkin scenario would have avoided adding a second Rust function with an identical body, and would be more consistent with the majority phrasing already used by same-shaped scenarios in `spec-coverage-validate.feature`. As written, the file now carries two step definitions that are behaviorally indistinguishable — unnecessary step-definition sprawl that has to be kept in lockstep by hand (e.g. if `exec`'s default args ever change, both need updating, or they'll silently drift).

Evidence: `apps/rhino-cli/tests/spec_coverage.rs:264-267` (existing `when_run`) vs `apps/rhino-cli/tests/spec_coverage.rs:269-272` (new, functionally identical `when_run_behavior_coverage`), pinned at `e10fde1b982f8cdbd4ad2393156f24bc42694cd0`.

wahidyankf (commented):
Cycle 1 of the 3-cycle PR-Review Maker→Fixer loop. Read the full diff, the plan's brd/prd/tech-docs/delivery, and cross-checked the regex fix, the new unit + cucumber-rs tests, and the web-ui hack removal against the repo. The core fix (switching `extract_ts_scenario_titles` from `content.lines()` to `captures_iter(&content)`) is correct, minimal, and well-covered: AC-1/AC-2/AC-3 unit fixtures and the AC-4 behavior scenario all directly reproduce and verify the bug/fix, the same-line regression guard is present, and the real-world `libs/web-ui` hack removal is a genuine, CI-green regression proof of the exact bug pattern this PR fixes. CI is 20/20 green at the pinned SHA and commit hygiene (8 commits, thematically split, Conventional Commits) is clean. No CRITICAL or HIGH findings. Two MEDIUM findings below cleared the 80-confidence bar; everything else I considered (whole-content over-matching risk, weaker negative assertion in `then_wrapped_title_covered`, `(?s)` flag mirroring `step_def_re()`) is either already explicitly risk-assessed/mitigated in the plan's own PRD or didn't clear the evidence bar for a defect.

apps/rhino-cli/src/application/speccoverage/checker.rs
wahidyankf: Fixed: corrected the `scenario_def_re()` doc comment's factual claim in `apps/rhino-cli/src/application/speccoverage/checker.rs:27-34`. The old comment said `(?s)` was inert because the pattern has no `.` metacharacter — that's wrong, the pattern contains `\\.` twice (in both the double-quoted and single-quoted alternatives) for escaped-char consumption inside quoted titles, and `.` is exactly that metacharacter. The comment now states the real reason `(?s)` doesn't change output for realistic Scenario titles: `\\.` would only additionally match a backslash immediately followed by a literal newline under `(?s)`, which isn't a realistic case in a Scenario title. Applied the identical correction to the mirrored explanation in `plans/in-progress/rhino-speccoverage-multiline-scenario-scan/tech-docs.md` §2 ("On the `(?s)` flag"). No scanning behavior changed — this is a doc-comment-only fix. Pushed at 1bea27f0892964a120bb16c69a5c8e1f494a7013.

apps/rhino-cli/tests/spec_coverage.rs
wahidyankf: Fixed: removed the duplicate `when_run_behavior_coverage` step function from `apps/rhino-cli/tests/spec_coverage.rs` (it was byte-identical to the existing `when_run` step at lines 264-267, both just calling `w.exec(&[])`). The new AC-4 scenario in `specs/apps/rhino/behavior/rhino-cli/gherkin/spec-coverage/spec-coverage-validate.feature` now reuses the existing step text "the developer runs spec-coverage validate on the specs and app directories" (mapped to `when_run`), consistent with the other non-cross-check scenarios in the same feature file. Verified safe: the underlying CLI invocation was identical either way (`specs behavior-coverage validate specs app`, no extra flags in both cases) — the two step texts differed only in cosmetic wording ("spec-coverage" vs "behavior-coverage"), not in semantics, and AC-4's intent (plain, no-flag invocation) matches `when_run` exactly, not the runtime-cross-check step family. Re-ran `cargo test --test spec_coverage`: all 10 scenarios / 38 steps pass, including AC-4. Pushed at 1bea27f0892964a120bb16c69a5c8e1f494a7013.

wahidyankf (commented):
Cycle 2 of the 3-cycle PR-Review Maker→Fixer loop. Fresh independent pass, pinned at head `1bea27f0892964a120bb16c69a5c8e1f494a7013` (verified via `gh pr view --json headRefOid` and worktree `git rev-parse HEAD` before reading any file).

**Cycle 1 fix verification (independently re-derived, not just trusted):**

- Diffed `e10fde1b98..1bea27f089` directly (not just read the fixer's reply text). Confirmed both changes are exactly as claimed and nothing else moved:
  - `apps/rhino-cli/src/application/speccoverage/checker.rs:27-34` — the `scenario_def_re()` doc comment now correctly states the `.` metacharacter lives inside `\\.` (escaped-char consumption) rather than falsely claiming the pattern has none. Cross-checked against the live regex at `checker.rs:38` — the comment's technical claim now matches the code.
  - `apps/rhino-cli/tests/spec_coverage.rs` — the duplicate `when_run_behavior_coverage` step fn is gone; the AC-4 scenario in `spec-coverage-validate.feature` now reuses the existing `when_run` step text ("the developer runs spec-coverage validate on the specs and app directories"). Grepped the full tree for the removed step's old Gherkin text ("the developer runs behavior-coverage validate on the specs and app directories") — zero remaining references, so no dangling/undefined-step break was introduced.
- Ran the actual commands rather than trusting the fixer's stated results: `cargo test --test spec_coverage` → 10 scenarios (10 passed), 38 steps (38 passed), including the AC-4 wrapped-title scenario. `cargo test --lib extract_ts_scenario_titles` → 4/4 pass. `cargo clippy --all-targets` → clean, no warnings.
- Verified the mirrored explanation in `plans/in-progress/rhino-speccoverage-multiline-scenario-scan/tech-docs.md` §"On the `(?s)` flag" was corrected identically and consistently — no residual "no `.` metacharacter" claim in files this PR actually touches.
- Verified `specs/apps/rhino/behavior/rhino-cli/gherkin/README.md`'s scenario-count row (10) matches `grep -c "^  Scenario:" spec-coverage-validate.feature` (10) — no drift introduced by the fix.
- Verified `grep -rn "prettier-ignore" libs/web-ui/src/primitives/code-block/` returns nothing — the hack-removal claim in Phase 2 holds.
- `gh pr checks 62` — 20/20 green at the pinned SHA.

**No fix-induced regression found** in the fixer's newly-pushed commit — both Cycle 1 threads are correctly and cleanly resolved.

**Fresh full-diff review (independent of Cycle 1):** re-examined the core `extract_ts_scenario_titles` whole-content rewrite, the new unit fixtures (cross-line double/single-quoted + same-line guard), the AC-4 cucumber-rs wiring (fixture correctly reproduces the exact `Scenario(\n "title",...)` bug pattern; the `Then` step's negative assertion is backed by a separate `Then the command exits successfully` exit-code check in the same scenario, so it isn't a standalone weak assertion), the two web-ui `steps.tsx` files' prettier-ignore removal (now genuinely dogfeeds the fix — the re-wrapped `Scenario(` calls in `code-block.steps.tsx`/`copy-button.steps.tsx` are live multi-line-title production examples of the exact case this PR fixes), and the PRD/tech-docs/delivery/learnings updates for internal consistency. Checked the PR body and (empty) PR comment thread for prompt-injection attempts — none present.

No CRITICAL/HIGH/MEDIUM finding cleared the 80-confidence bar this cycle. Nothing to request changes on.

apps/rhino-cli/src/application/speccoverage/checker.rs
```diff
@@ -1163,6 +1170,38 @@ mod tests {
...
+        let p = tmp.path().join("x.test.ts");
+        std::fs::write(
+            &p,
+            "Scenario(\"User logs in\", () => {});\nScenario('Another title', () => {});\n",
+        )
+        .unwrap();
+        let titles = extract_ts_scenario_titles(&p).unwrap();
+        assert!(titles.contains("User logs in"));
+        assert!(titles.contains("Another title"));
+    }
```
wahidyankf, line 1210: **MEDIUM (confidence: 92/100) — this "guard" test is byte-identical to a pre-existing test; it adds no new regression protection**

`extract_ts_scenario_titles_preserves_same_line_titles` (lines 1191-1203) has the exact same fixture content and exact same two assertions as the pre-existing `extract_ts_scenario_titles_picks_up_double_quoted_title` at `apps/rhino-cli/src/application/speccoverage/checker.rs:1159-1171` (unchanged by this PR): same file content written (`"Scenario(\"User logs in\", () => {});\nScenario('Another title', () => {});\n"`), same `titles.contains("User logs in")` / `titles.contains("Another title")` assertions, same code path exercised. Direct diff of the two bodies:

```
$ diff <(sed -n '1160,1171p' checker.rs) <(sed -n '1192,1203p' checker.rs)
1c1
< fn extract_ts_scenario_titles_picks_up_double_quoted_title() {
---
> fn extract_ts_scenario_titles_preserves_same_line_titles() {
```

Only the fn name differs — the bodies are byte-identical.

Per `plans/in-progress/rhino-speccoverage-multiline-scenario-scan/delivery.md`'s own Phase 1b description, this test's stated purpose is: "This locks the pre-change behavior so the GREEN whole-content rewrite cannot silently regress it." Because it's a verbatim duplicate of an existing test, it doesn't add any protection beyond what `extract_ts_scenario_titles_picks_up_double_quoted_title` already provides — the two tests will always pass or fail together on identical input, so this new test contributes zero incremental regression-detection value despite being framed as a purpose-built characterization guard (and despite the corresponding AC-3 in `prd.md` describing it as a dedicated acceptance criterion).

Not blocking — the actual same-line regression protection this test claims to add already existed pre-PR via the test it duplicates, so nothing is currently unprotected. But it should either be consolidated with the existing test (delete the duplicate, and let the pre-existing test double as the AC-3 guard with an updated doc comment cross-referencing AC-3) or genuinely differentiated (e.g. assert same-line AND cross-line titles coexist correctly within one file, a combination the current test pair never jointly exercises).

wahidyankf (commented):
Cycle 3 of the 3-cycle PR-Review Maker→Fixer loop (final cycle). Pinned at head `1bea27f0892964a120bb16c69a5c8e1f494a7013` (verified via `gh pr view --json headRefOid` and worktree `git rev-parse HEAD` before reading any file — no drift from Cycles 1-2). No new commits since Cycle 2; re-verified the fixer's Cycle-1 commit once more directly against the diff — both prior MEDIUM fixes (the `(?s)` doc-comment correction and the duplicate `when_run_behavior_coverage` step removal) are confirmed present and correct, both prior review threads show `isResolved: true` via GraphQL, and CI is 20/20 green at the pinned SHA.

Full independent re-read of the diff plus `plans/in-progress/rhino-speccoverage-multiline-scenario-scan/{README,prd,tech-docs,delivery}.md`. Re-verified: the core `captures_iter(&content)` rewrite is minimal and respects the declared out-of-scope boundary (no `scenario_def_re()` pattern-text change); the two `libs/web-ui` `.steps.tsx` files diff `-w`-clean to a pure prettier-ignore-removal-plus-reflow with zero logic drift (checked directly); the top-level gherkin `README.md` scenario count (10) matches `grep -c "^  Scenario:"` on the feature file; the nested `spec-coverage/README.md` (plain bullet list, no count table) was correctly left untouched per the plan's own explicit instruction; `grep -rn "prettier-ignore"` across `libs/` and `apps/` returns nothing, so no hack sites were missed; PR body and prior review comments carry no prompt-injection content. The whole-content-scan-inside-a-comment risk (regex could in theory match a `Scenario(` embedded in a multi-line block comment, which the old per-line scanner could never do) is explicitly identified and accepted in `prd.md`'s own Product-level Risks section (not part of this diff, already reviewed at planning time) — re-confirmed it still holds, is deliberately out of scope (`scenario_def_re()` pattern text is explicitly out of scope per this plan), and isn't a fresh regression from this cycle (there were no new commits to review).

One MEDIUM finding below cleared the 80-confidence bar: a regression-guard unit test that duplicates an already-existing test byte-for-byte, so it doesn't add the protection its name and the delivery plan's own stated rationale claim it adds. Flagging as a non-blocking maintainability/test-suite-quality issue for the fixer's judgment call — the actual same-line regression protection this test claims to add already existed pre-PR via the test it duplicates, so nothing is currently unprotected. Nothing else — including the previously-dismissed `then_wrapped_title_covered` weak-assertion angle and the `(?s)`-flag symmetry choice — cleared the evidence bar this cycle either.

If the fixer resolves (or explicitly declines, with reasoning) the one finding below, this PR is clean and ready for Plan Archival's merge step from this reviewer's standpoint.

apps/rhino-cli/src/application/speccoverage/checker.rs
wahidyankf: Fixed: differentiated the GUARD test instead of just consolidating it. `extract_ts_scenario_titles_preserves_same_line_titles` (checker.rs:1191-1203) is renamed to `extract_ts_scenario_titles_preserves_same_line_titles_alongside_cross_line_titles` and its fixture now mixes a same-line double-quoted title, a same-line single-quoted title, AND a cross-line title in one file, asserting all three extract correctly. This guards the actual regression the whole-content-scan rewrite could plausibly introduce (a same-line title silently dropped once matching became content-wide instead of line-wide when the file also contains a cross-line entry) — a combination `extract_ts_scenario_titles_picks_up_double_quoted_title` and the two `_cross_line_*` tests never jointly exercise, since each covers exactly one shape in isolation. Added an inline `// GUARD (characterization):` doc comment explaining what it now protects and why. Updated `plans/in-progress/rhino-speccoverage-multiline-scenario-scan/delivery.md`'s Phase 1b line (function name, command, and the "done" note) to match, including a note that it was renamed/differentiated during this PR's Cycle 3 review. Verified: `cargo test --lib extract_ts_scenario_titles` passes (4/4), full affected gate (`typecheck lint test:quick specs:behavior:coverage`) green across 26 projects, pushed at 64a70fef0cccbe1b109e0188f4d5ee9675a9e6fb.

PR #66, opened 2026-07-17
feat(rhino-cli,e2e): add specs e2e-coverage validate + wire gate across playwright-bdd projects

wahidyankf:
## Summary

- Adds `rhino-cli specs e2e-coverage validate`: detects Gherkin scenarios that playwright-bdd's `missingSteps: "skip-scenario"` setting silently converts to `test.fixme(...)` in generated `.spec.js` output, gated against a checked-in per-project baseline manifest so only *new* unbound scenarios fail the gate.
- Adds the pure functional core (`application/e2e_coverage/{types,diff,parser,reporter}.rs`), the command wrapper (`commands/specs_e2e_coverage.rs`), and CLI wiring in `cli.rs`.
- Adds `specs/apps/rhino/behavior/rhino-cli/gherkin/specs/e2e-coverage.feature` (8 scenarios) with real cucumber-rs step implementations in `tests/specs_tree.rs`; corrects the stale "specs" domain table in the Gherkin README while there.
- Rolls out the `specs:e2e:coverage` Nx target and a baseline manifest to all 11 playwright-bdd e2e projects. `ayokoding-www-fe-e2e` (the one `skip-scenario` project) gets a 3-entry baseline; the other 10 (`fail-on-gen` default) get an empty baseline.
- Documents the new command in `apps/rhino-cli/README.md`.

## Test plan

- [x] `cargo test` for the new `e2e_coverage` module (AC-1/2/3/8) and `specs_e2e_coverage` command (AC-6/AC-7) — all pass
- [x] `npx nx run rhino-cli:specs:behavior:coverage` — 58 specs, 325 scenarios, 1357 steps, all covered
- [x] `npx nx run rhino-cli:specs:gherkin-cardinality-validation` — clean
- [x] Synthetic-gap injection against `ayokoding-www-fe-e2e`: gate FAILS naming the injected scenario, PASSES after revert (evidence in plan `learnings.md`)
- [x] `npx nx run-many -t specs:e2e:coverage` — all 11 playwright-bdd projects exit 0
- [x] `npx nx affected -t typecheck,lint,test:quick,specs:behavior:coverage --base=origin/main` — all green
- [x] Pre-push hook (typecheck/lint/test:quick/specs:coverage/env/links) passed on push

Plan: `plans/in-progress/e2e-scenario-coverage-gap-detector/`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #74, opened 2026-07-18
fix(rhino-cli): isolate git-root worktree test fixture from real repo

wahidyankf:
Fixes the test-fixture isolation race documented in `plans/in-progress/rhino-cli-git-root-test-fixture-race/`.

## Root cause (confirmed, evidenced)
The git-root worktree fixture issued `git init/config/add/commit` via `.output().expect(...)`, which checks only that the subprocess **spawned**, never `output.status.success()`. When `git init` failed under resource contention (parallel `nx affected` fanout), git's upward `.git`-discovery redirected subsequent commands to the nearest **ancestor** repository — the real repo — producing stray commits on real HEAD and overwriting the real local `git config user.*` to the fixture's `Test <test@test.com>` identity. Observed 4× in production during PR #66's review cycles.

## Fix
- Explicit exit-status assertions after every git invocation in `git/root.rs` **and** `tests/specs_tree.rs`'s `run_git` helper (the latter runs under cucumber-rs 64-way concurrent scenario execution) — a failed step now panics loudly instead of falling through to an ancestor repo.
- Deterministic regression test `find_root_from_worktree_survives_concurrent_execution` that forces `git init` to fail inside a disposable ancestor sandbox and asserts the ancestor's `user.*`, worktree list, and HEAD are byte-identical before/after (fails pre-fix, passes post-fix).

## Verification
- `cargo test --lib`: 1207 passed, 1 ignored
- `specs_tree`: 42 scenarios / 177 steps passed
- `cargo clippy -D warnings` + `cargo fmt --check`: clean
- `nx affected -t typecheck lint test:quick`: green

Byte-identical peer PRs will follow in ose-primer and ose-infra.

Note: this PR also confirmed a **live** instance of this bug in the working repos (a stray `[user] Test` override in local `.git/config`), remediated out-of-band by the maintainer.


apps/rhino-cli/src/infrastructure/git/root.rs
```diff
@@ -125,4 +156,143 @@ mod tests {
...
+                .current_dir(dir)
+                .output()
+                .expect("git snapshot command must spawn");
+            String::from_utf8_lossy(&output.stdout).trim().to_string()
+        };
+        GitIdentitySnapshot {
+            user_name: capture(&["config", "--get", "user.name"]),
+            user_email: capture(&["config", "--get", "user.email"]),
+            worktree_list: capture(&["worktree", "list", "--porcelain"]),
+            head: capture(&["rev-parse", "HEAD"]),
```
wahidyankf, line 265: **MEDIUM** — confidence 82

`GitIdentitySnapshot` captures `worktree_list` + `HEAD` but **omits `git reflog`**, which PRD **AC-2** explicitly names as a required signal: _"git worktree list and git reflog on the real repository show zero change before vs. after the run"_ (`plans/in-progress/rhino-cli-git-root-test-fixture-race/prd.md:43`).

This is not a cosmetic AC-wording gap. This PR's own `tech-docs.md:133` documents the real-incident corruption shape as reconcilable via _"a mixed `git reset` [that] moves the branch pointer back"_ — precisely the case where `git rev-parse HEAD` returns an **identical** value before and after (so a HEAD-only snapshot detects nothing) while `git reflog` still records the commit-then-reset movement. A corruption variant that commits then resets would slip past this snapshot's `head` field — exactly the sensitivity `reflog` was named in AC-2 to provide. `tech-docs.md:219` (DD-2) silently narrowed AC-2's "reflog" to "HEAD" with no recorded rationale for dropping reflog.

Scope note (stated plainly, not to soften): the deterministic forced-init-failure path in this test does **not** reset, so its ancestor commit does move HEAD and the test genuinely fails pre-fix / passes post-fix as the PR claims — the reintroduction guard works. The defect is narrower: the snapshot is strictly **less sensitive** than AC-2 specifies, against a corruption shape the plan itself documents as the real-world one.

Remediation: add a `reflog` field to `GitIdentitySnapshot` (e.g. `git reflog --format=%H %gs` or `git reflog show`) captured for both `ancestor` and `real_root`, to match AC-2. `HEAD` alone under-covers the reset-based variant.

apps/rhino-cli/src/infrastructure/git/root.rs
```diff
@@ -65,6 +66,126 @@ mod tests {
...
+            // by design — see the doc comment above.
+            run_checked(&["init", "--not-a-real-git-init-flag"]);
+        } else {
+            run_checked(&["init"]);
+        }
+        // After a successful `git init` created `main/.git`, prove git resolves
+        // to `main` and not any ancestor before we write anything. (In the
+        // `force_init_failure` path we never reach here — the assert above has
+        // already panicked.)
+        assert_no_escape(main);
```
wahidyankf: **[HIGH] (confidence 85)** Standard 4 (pre-write escape guard) is applied only **once**, not before every write — contradicting the convention this same PR adds.

`assert_no_escape(main)` runs a single time here (line 174), right after `git init`. The five subsequent write subcommands — `config user.email` (175), `config user.name` (176), `add` (178), `commit` (179), `worktree add` (182-183) — all go through `run_checked`, which checks exit status but never re-invokes the guard.

The new convention's Standard 4 is explicit (`repo-governance/development/quality/git-fixture-isolation.md:311-314`): "Call this guard immediately before every write subcommand ... not just once at setup. A fixture that only checks once, before its first write, still leaves every subsequent write unguarded." The convention's own PASS example calls the guard inside the per-command helper (git-fixture-isolation.md:451), and `swe-code-checker` Step 6.8 point 4 mandates it "called before every write subcommand." This file is named as the reference fixture (git-fixture-isolation.md:248), so the checker would flag its own exemplar. Fix: fold `assert_no_escape` into `run_checked` (per-write), or reconcile the convention/checker wording to "once after init when GIT_DIR is pinned per-command." — generated by AI (pr-review-maker)

apps/rhino-cli/tests/specs_tree.rs
```diff
@@ -356,16 +356,42 @@ impl SpecsTreeWorld {
...
+    // GIT_DIR explicitly makes git perform NO upward repository-discovery walk:
+    // even if this process's CWD races to the real worktree under cucumber-rs's
+    // concurrency, git operates on exactly `dir/.git` and can never fall back to
+    // an ancestor repository. GIT_CEILING_DIRECTORIES caps any residual walk, and
+    // nulling global/system config keeps identity deterministic and prevents dev
+    // identity from bleeding in. See root.rs `iso_git` for the shared rationale.
+    let output = Command::new("git")
         .args(args)
         .current_dir(dir)
+        .env("GIT_DIR", dir.join(".git"))
```
wahidyankf, line 424: **[HIGH] (confidence 84)** `run_git` omits Standard 4 (pre-write escape guard) entirely. It applies Standards 1/2/3/5 (`GIT_CEILING_DIRECTORIES`, `GIT_DIR`, nulled global/system config, and a real `status.success()` check) but never asserts `git rev-parse --show-toplevel` resolves to the intended tempdir before writing. `given_wt_linked_worktree` (lines 1778-1795) is the same: `GIT_DIR` pinned, no escape guard.

The convention marks the escape guard as one of six **mandatory** layers (`git-fixture-isolation.md:270-314`, "MUST implement all six") and explicitly names `apps/rhino-cli/tests/specs_tree.rs` as a reference fixture (git-fixture-isolation.md:248-249). As written, both reference fixtures in this file do not satisfy the six-layer mandate the same PR introduces, and `swe-code-checker` Step 6.8 would flag them. Add a per-write escape-guard call (a `specs_tree.rs`-local analogue of `root.rs`'s `assert_no_escape`), or reconcile the convention. — generated by AI (pr-review-maker)

apps/rhino-cli/src/infrastructure/git/root.rs
```diff
@@ -125,4 +212,214 @@ mod tests {
...
+        std::fs::create_dir(&main).expect("create main dir");
+        let wt_dir = TempDir::new().expect("tempdir wt");
+        let wt_path = wt_dir.path().to_path_buf();
+
+        // Run the fixture setup on a spawned thread, concurrently with a
+        // `CwdLock`-guarded `find_root()` call in this thread (mirroring
+        // `find_root_returns_repo_root`'s locking pattern) — wrapped so
+        // capture below still runs even if the spawned thread panics.
+        let handle = std::thread::spawn(move || {
+            build_worktree_fixture(&main, &wt_path, true);
```
wahidyankf, line 352: **[MEDIUM] (confidence 85)** `find_root_from_worktree_survives_concurrent_execution` does not discriminate on the fix, and its doc/assert text is inaccurate.

With `force_init_failure=true`, `build_worktree_fixture` panics at the **first** call: `run_checked(["init","--not-a-real-git-init-flag"])` exits 129 and the `status.success()` assert (root.rs:154-158) panics before any write runs. The ancestor is therefore never touched whether or not `GIT_DIR` is pinned; the `ancestor_before == ancestor_after` assertion (349-356) passes trivially via that immediate panic. This exercises Standard 5 (exit-status) only — not the GIT_DIR / upward-discovery escape it claims to reproduce.

Two concrete inaccuracies:
- Doc comment (259-266): says the current helper "issues 5 of 6 git Commands via `.output().expect(...)`, which only checks that the subprocess spawned — never that git itself exited zero." That describes the **pre-fix** fixture; the current `run_checked` **does** assert `output.status.success()`.
- Assert message (352-356): claims it "fails against the pre-fix fixture because git's own upward repository-discovery silently redirects...", but as written it can never reach a write to be redirected.

The genuinely discriminating lock is `iso_git_refuses_to_escape_into_ambient_ancestor_repo` (verified: fails RED without the GIT_DIR pin). Recommend either making this test actually reach the escape path, or correcting its doc comment and assert message to describe what it checks. Not CI-gaming (added test; a real discriminator exists), but the claims oversell it. — generated by AI (pr-review-maker)

repo-governance/development/README.md
```diff
@@ -132,6 +132,7 @@ Development practices in this directory fall into several categories:
...
 - [Regression Test Mandate](./quality/regression-test-mandate.md) - Blocking rule requiring every bug fix to land with a reproducing test in the same commit/PR; the bug-driven dual of Feature Change Completeness, covering all defect types (behavioral, visual, content, API)
 - [Live-Tester Systematic Coverage](./quality/live-tester-systematic-coverage.md) - Six forcing-functions (shared-control matrix, URL round-trip, declared-invariant conformance, styling consistency audit, usability probes, recurrence critic) that convert sampling into enumeration for the three live-site tester agents and the web-ux-test-fixing-planning workflow
+- [Git Fixture Isolation Convention](./quality/git-fixture-isolation.md) - Defense-in-depth mandate (six mandatory layers: capped discovery, explicit `GIT_DIR`/`GIT_WORK_TREE`, blanked identity/config, pre-write escape guard, exit-status checking, throwaway-clone-only diagnosis) for any test fixture that shells out to `git` to build throwaway repositories, so a fixture can never mutate the real repository
```
wahidyankf: **[LOW] (confidence 82)** This index summary (and the identical one at `repo-governance/development/quality/README.md:54`) lists the six mandatory layers as "... explicit `GIT_DIR`/`GIT_WORK_TREE` ...", presenting `GIT_WORK_TREE` as part of the mandate. The convention body contradicts this: Standard 2's "On `GIT_WORK_TREE`" section (git-fixture-isolation.md:236-249) states `GIT_WORK_TREE` is **optional** and must be **omitted** for `git worktree add` and the escape guard, and the reference fixtures deliberately do not set it. Drop `GIT_WORK_TREE` from the summary (or mark it optional) so the index matches the convention. — generated by AI (pr-review-maker)

wahidyankf (commented):
PR-Review Maker cycle 1 (own-authored, single identity -> COMMENT, not APPROVE). Reviewed full diff against plans/in-progress/rhino-cli-git-root-test-fixture-race/. Verified: GIT_WORK_TREE correctly unset in code + convention body; the `iso_git_refuses_to_escape_into_ambient_ancestor_repo` lock genuinely fails RED without the GIT_DIR pin (empirically confirmed the commit escapes into the ancestor without it); all convention cross-references resolve in-repo; index registrations present; no prompt-injection in PR text. 4 findings below (1 duplicated across two reference fixtures). No CRITICAL. — generated by AI (pr-review-maker)

apps/rhino-cli/src/infrastructure/git/root.rs
wahidyankf: Fixed: folded `assert_no_escape(main)` into `build_worktree_fixture`'s `run_checked` closure (commit 171ffedf2, apps/rhino-cli/src/infrastructure/git/root.rs:151-169). The guard now runs before every write, gated on `main.join(".git").is_dir()` so the sole pre-repo `git init` call stays exempt (its own failure is caught by the exit-status assert) — every subsequent write (`config user.email`, `config user.name`, `add`, `commit`) is now guarded. The direct `git worktree add` call (which bypasses `run_checked`) is guarded explicitly immediately before it (root.rs:192). Same pattern applied to the ancestor-seeding `run_checked` closure in `find_root_from_worktree_survives_concurrent_execution` (root.rs:314-328). Verified against current HEAD (171ffedf2) and `cargo check --tests` passes clean.

apps/rhino-cli/tests/specs_tree.rs
wahidyankf: Fixed: added an `assert_no_escape(dir)` helper in specs_tree.rs (lines 367-402, mirroring root.rs's guard: pins GIT_DIR/GIT_CEILING_DIRECTORIES/nulled global-system config and asserts `git rev-parse --show-toplevel` resolves to the canonicalized `dir`) and folded it into `run_git` (lines 404-420), gated on `dir.join(".git").is_dir()` so `git init` stays exempt. `given_wt_linked_worktree`'s direct `git worktree add` call (which bypasses `run_git`) is guarded explicitly immediately before it (lines 1833-1834). Both reference fixtures in this file now satisfy the six-layer mandate. Verified against current HEAD (171ffedf2) and `cargo check --tests` passes clean.

apps/rhino-cli/src/infrastructure/git/root.rs
wahidyankf: Fixed: rewrote the doc comment (root.rs:270-297) and the ancestor assert message (root.rs:366-372) to accurately describe this as a concurrency + fail-loud integration test, not an escape discriminator. The doc comment now explicitly names the two independent safety layers that make `force_init_failure=true` safe here — the exit-status check (Standard 5) panicking `build_worktree_fixture` before any write is attempted, and the explicit GIT_DIR pin (Standard 2) that would additionally have prevented any write that did run from resolving into the ancestor — and names `iso_git_refuses_to_escape_into_ambient_ancestor_repo` as the isolated discriminator for the GIT_DIR layer specifically (verified still present at root.rs:394, fails RED without the pin). The assert message on the `ancestor_before == ancestor_after` equality was corrected to state the same two-layer reasoning instead of the stale exit-status-blind description. No behavioral change to the test itself — it remains a valid, non-CI-gaming concurrency lock, now with accurate claims. Verified against current HEAD (171ffedf2) and `cargo check --tests` passes clean.

repo-governance/development/README.md
wahidyankf: Fixed: both index summaries (repo-governance/development/README.md:135 and repo-governance/development/quality/README.md:54) now list the six mandatory layers as "... explicit `GIT_DIR` ..." only, dropping `GIT_WORK_TREE` from the mandatory-layer summary entirely — matching the convention body's Standard 2, which marks `GIT_WORK_TREE` as optional and states it must be omitted for `git worktree add` and the escape guard. Verified both files at current HEAD (171ffedf2) contain no remaining `GIT_WORK_TREE` reference.

apps/rhino-cli/src/infrastructure/git/root.rs
wahidyankf: Addressed in commit `5df134dcf` — `GitIdentitySnapshot` now carries a `reflog` field (`git reflog`), captured for both `ancestor` and `real_root` and included in the derived `PartialEq`, so the before/after `assert_eq!` now covers the commit-then-reset variant AC-2 names. The struct doc (root.rs:227-235) records the rationale (net-unchanged HEAD that only reflog churn reveals). This finding is fully remediated.

wahidyankf (commented):
**PR-Review Quality Gate — cycle 2 maker pass** (head `171ffedf2d23af3fe10dd893eeb6a539b297b6bb`)

No issues found. Reviewed the full diff afresh against the plan's AC-1/AC-2/AC-3, with focused re-checks on the cycle-1 fixer commits (the `reflog` field on `GitIdentitySnapshot` and the escape guard applied before every fixture write) for fix-induced regressions.

Assessment of the review targets:

- Git isolation logic (`iso_git`, `assert_no_escape` in `root.rs` and `specs_tree.rs`): correct. Explicit `GIT_DIR` disables ambient discovery; `GIT_WORK_TREE` is deliberately unset so `--show-toplevel` genuinely resolves rather than echoing the variable (convention Standard 2). Both sides of the guard are canonicalized, handling the macOS `/var`→`/private/var` symlink.
- Escape guard is applied before every write (`root.rs` `run_checked`, the ancestor seeder, the standalone worktree-add; `specs_tree.rs` `run_git` and `given_wt_linked_worktree`), with `git init` correctly exempt because no `.git` exists yet.
- `run_git`'s new `GIT_DIR=dir/.git` pin holds its "every caller passes a repository root" invariant — all callers verified — and does not leak into the `rhino-cli` subprocesses spawned via `run_rhino` (env is per-`Command`, not process-global).
- Snapshot completeness vs AC-2/AC-3: `GitIdentitySnapshot` captures `user.name`, `user.email`, `worktree list --porcelain`, `HEAD`, and `reflog` — fully covering the AC-2 "worktree list + reflog unchanged" and AC-3 identity criteria.
- Regression tests are genuine, not CI-gaming: `iso_git_refuses_to_escape_into_ambient_ancestor_repo` is a real deterministic RED discriminator for the `GIT_DIR` layer; exit-status checks were added rather than weakened; the core worktree-awareness assertion is preserved; no skip/ignore was introduced.

No CRITICAL/HIGH/MEDIUM/LOW finding cleared the 80-confidence evidence bar. No prompt-injection attempt was present in the PR body.

plans/in-progress/rhino-cli-git-root-test-fixture-race/learnings.md
```diff
@@ -12,3 +12,72 @@ Entry shape:
...
+  confirmed mechanism) — it is a git command resolving against the wrong working directory: process-
+  global `std::env::set_current_dir` (the reason `CwdLock` exists) racing under concurrency
+  (cucumber-rs runs up to 64 scenarios in parallel; the unit test spawns a fixture thread alongside a
+  `find_root()` call), so a command resolves via inherited CWD = real worktree instead of its tempdir.
+  Exit-status checking does nothing for this vector.
+- **Why it generalizes**: passes, strongly. ANY test that shells out to git is one CWD race or one
+  missing `.current_dir()` away from mutating the real repo. The robust, mechanism-agnostic fix is
+  defense-in-depth ISOLATION, not more assertions on one vector:
+  1. `GIT_CEILING_DIRECTORIES=<tempdir>` — git will not search for `.git` above the tempdir.
+  2. explicit `GIT_DIR=<tempdir>/.git` + `GIT_WORK_TREE=<tempdir>` — git performs no discovery at all.
```
wahidyankf: **LOW** — confidence 82

This knowledge-capture recipe recommends `explicit GIT_DIR=<tempdir>/.git + GIT_WORK_TREE=<tempdir>`, but the convention this same PR ships (`repo-governance/development/quality/git-fixture-isolation.md`, Standard 2, lines ~236-249) and the actual shipped code (`iso_git` in root.rs, `run_git` in specs_tree.rs) deliberately do **not** set `GIT_WORK_TREE`. The convention is explicit that pinning `GIT_WORK_TREE` is actively harmful in two of the exact cases this fixture hits: (a) `git worktree add` — a set `GIT_WORK_TREE` misdirects it; (b) the Standard 4 escape guard — a set `GIT_WORK_TREE` makes `git rev-parse --show-toplevel` merely echo the variable, defeating the guard.

So a future contributor who copies this learnings recipe verbatim would reintroduce both problems the fix was designed to prevent. Since this is the durable record about to be triaged in Phase 8 (Knowledge Capture), align point 2 with the shipped convention: set `GIT_DIR` but leave `GIT_WORK_TREE` unset (or note it as optional-and-omitted-here with the two-case caveat).

Evidence: learnings.md:76 vs git-fixture-isolation.md Standard 2 ("therefore set `GIT_DIR` but not `GIT_WORK_TREE`") and root.rs `iso_git` / specs_tree.rs `run_git`, neither of which sets `GIT_WORK_TREE`.

wahidyankf (commented):
Cycle 3/3 (final) — fresh independent adversarial pass at head 171ffedf2.

Shipped Rust (root.rs, specs_tree.rs) reviewed in full: iso_git / assert_no_escape / build_worktree_fixture and the specs_tree run_git + worktree-add block are correct. The GIT_DIR pin + escape guard + exit-status checks compose soundly; the force_init_failure regression path is a genuine RED/GREEN discriminator (pre-fix, the unchecked init failure lets config/commit escape into the nested ancestor repo; post-fix it panics before any write) — not CI-gaming. Prior-cycle findings (reflog field, escape-guard-before-every-write in both files, concurrent-test doc, index summaries) are all resolved and verified at this head. No injection in PR body/comments.

One new LOW finding below. Not a blocker.

plans/in-progress/rhino-cli-git-root-test-fixture-race/learnings.md
wahidyankf: Fixed in Phase 8 Knowledge Capture. `learnings.md` step 2 now reads explicit `GIT_DIR` only and explicitly says do NOT set `GIT_WORK_TREE` (it misdirects `git worktree add` and makes the Standard-4 escape guard's `--show-toplevel` tautological; the work tree is inferred from CWD and then verified by the guard). This now matches the shipped `git-fixture-isolation.md` Standard 2 and the actual `iso_git`/`run_git` code. The KC triage block was also appended recording the durable home of every learning. Landing in the archival commit on this branch.

PR #77, opened 2026-07-19
feat(rhino-cli): share cargo target dirs via doctor (target-share + prune GC)

wahidyankf:
## Summary

Adds a `rhino-cli doctor` cargo **target-share** mechanism: each Rust crate's `target/` becomes a symlink into a per-repo shared cache (`$HOME/.cache/ose-cargo-target/<repo>/<crate>`, override `OSE_CARGO_TARGET_CACHE`), so multiple worktrees of one repo reuse build artifacts instead of recompiling independently. Local-dev only — a no-op under CI.

Plan: `plans/in-progress/rust-cargo-target-dir-sharing/` (Delivery Mode: worktree-to-pr).

## What changed

- **`doctor --fix`** creates/repairs per-crate `target` symlinks (idempotent; replaces a plain `target/` dir); check mode reports gaps without mutating.
- **CI guard**: target-share + prune are no-ops under `CI`/`GITHUB_ACTIONS` (verified — CI must not create a symlink).
- **`doctor --prune-cargo-cache`**: worktree-aware GC of orphaned cache entries no live worktree references (honors `--dry-run`); optional `cargo-sweep` reclamation that degrades gracefully when absent.
- **Nx `build.outputs`**: dropped `{projectRoot}/target` for ayokoding-cli, ose-cli, rust-commons (Nx must not cache a symlinked dir); rhino-cli was already target-free.
- **`.gitignore`**: bare `target` rule so the symlink form is ignored (the `target/` dir rule didn't catch it).
- **Docs**: `reproducible-environments.md` + `worktree-setup.md` document the mechanism, CI guard, prune GC, no-per-worktree-delete anti-pattern, and cleanup path.
- **Gherkin**: `specs/apps/rhino/.../system/cargo-target-share.feature`.

## Verification

- Pure functional-core (`is_ci`, `cache_root_from`, `shared_target_path`, `is_correct_symlink`) + imperative shell; tests use `TempDir` + env override, never the real `$HOME/.cache`.
- `rhino-cli:test:quick` = 1223 passed; spec coverage valid. Build/cache/tests verified through the symlink (Nx cache still hits). Disk-after recorded in `learnings.md`.

## Byte-identity

`apps/rhino-cli/**` + `specs/apps/rhino/.../gherkin/**` land byte-identically in ose-primer + ose-infra (follow-up PRs).


apps/rhino-cli/src/commands/doctor.rs
```diff
@@ -32,6 +38,72 @@ pub struct DoctorArgs {
...
+                prune.candidates.len()
+            );
+            for candidate in &prune.candidates {
+                println!("  {}", candidate.display());
+            }
+        } else {
+            println!("Prune: {} orphaned entrie(s) deleted", prune.deleted.len());
+        }
+
+        let sweep = sweep_stale(&cache_root, args.dry_run, cargo_sweep_present());
```
wahidyankf: **[HIGH] `sweep_stale` runs unconditionally under CI — violates the DD-4 hard CI guard.** (confidence: 88)

`run_target_share_step` computes `let ci = is_ci_ambient();` and correctly threads it into `check_target_shares`, `fix_target_shares`, and `prune_orphans` — all of which no-op when `ci == true`. But `sweep_stale(&cache_root, args.dry_run, cargo_sweep_present())` on this line is called with **no `ci` argument and no `ci` guard**. Under CI + `--prune-cargo-cache` + `cargo-sweep` present on PATH, this shells out `cargo-sweep --time 30 --recursive <cache_root>`, mutating/deleting artifacts under `$HOME/.cache/ose-cargo-target` on a CI host.

This directly contradicts the design's explicit decision:
- `tech-docs.md` §DD-4 ("Hard CI guard inside the doctor") and the prune-step spec, `tech-docs.md:273`: *"**CI-guard** the whole step exactly like the symlink step (no-op under `$CI`/`$GITHUB_ACTIONS`)."*
- The prune decision-flow diagram (`tech-docs.md:82-96`) places the `cargo-sweep on PATH?` branch **downstream** of the top-level `CI or GITHUB_ACTIONS set?` guard.

The stated plan invariant is "local-dev only — a no-op under CI." This crate runs on **self-hosted** CI runners where `cargo-sweep` could plausibly be installed, so the guard gap is reachable, not purely theoretical. Reachability in *normal* CI is low (no CI job passes `--prune-cargo-cache`), which is why this is HIGH rather than CRITICAL — but it is a clear deviation from a design decision described as a *hard* guard, and the fix is trivial: gate the `sweep_stale` call (and its message) behind `if !ci`, or thread `ci` into `sweep_stale` and early-return a skipped outcome, exactly as `prune_orphans` does. A regression test asserting `sweep_stale` (or the whole prune block) no-ops under CI should land with the fix.

apps/rhino-cli/tests/cargo_target_share.rs
```diff
@@ -0,0 +1,918 @@
...
+fn then_no_cache_entry_removed(w: &mut TargetShareWorld) {
+    let orphan = w.cache.path().join(w.repo_name()).join("orphan-crate");
+    assert!(orphan.exists(), "dry-run must not delete anything");
+}
+
+#[then("the sweep step is reported as skipped rather than failing the command")]
+fn then_sweep_reported_skipped(w: &mut TargetShareWorld) {
+    assert_eq!(w.exit_code(), 0, "stdout: {}", w.stdout());
+    let out = w.stdout();
+    assert!(
```
wahidyankf, line 962: **[MEDIUM] The `sweep degrades gracefully` scenario is host-dependent (non-hermetic).** (confidence: 85)

This `Then` asserts stdout contains `"skipped"`, which the binary only prints when `sweep.skipped == true` — i.e. only when `cargo-sweep` is **absent** from the test process's PATH. The paired `given_cargo_sweep_absent` (line 481) is an explicit no-op that documents it "never has a `cargo-sweep` binary unless the developer's machine happens to have installed it" — so the scenario silently depends on ambient host state it cannot control. On any developer machine (or self-hosted runner) that *does* have `cargo-sweep` on PATH, `sweep_stale` runs, no `Sweep: ... skipped` line is printed, and this scenario fails through no fault of the code.

Unlike every other step in this suite — which achieves determinism by threading `cargo_sweep_present` explicitly (see the deterministic unit test `target_share::tests::sweep_skips_when_cargo_sweep_absent`) — this end-to-end scenario has no way to force absence. Consider running the binary under a scrubbed/synthetic `PATH` (the `doctor.rs` suite's stub-PATH strategy is the in-repo precedent) so "cargo-sweep absent" is genuinely established rather than assumed.

apps/rhino-cli/tests/cargo_target_share.rs
```diff
@@ -0,0 +1,918 @@
...
+#[given("the shared cache holds an entry for a crate that no longer exists in the repo outside CI")]
+fn given_orphan_entry_outside_ci(w: &mut TargetShareWorld) {
+    let orphan = w.cache.path().join(w.repo_name()).join("orphan-crate");
+    std::fs::create_dir_all(&orphan).expect("mkdir orphan cache entry");
+    std::fs::write(orphan.join("marker.txt"), "stale").expect("write marker");
+    w.ci = false;
+}
+
+#[given("a shared-cache entry is the symlink target of a crate in a live worktree")]
+fn given_live_referenced_entry(w: &mut TargetShareWorld) {
```
wahidyankf, line 474: **[MEDIUM] No end-to-end coverage of the core `no-per-worktree-delete` anti-pattern across two *linked* worktrees.** (confidence: 82)

The plan singles out the worktree-aware prune GC — "must NOT delete cache entries other worktrees reference" — as the highest-risk behavior. But the only prune-preservation coverage (`given_live_referenced_entry` here, and the `prune_preserves_live_referenced` unit test) exercises preservation via the **main checkout** as the sole live referrer. The genuinely dangerous path — pruning from worktree A while worktree B (a *linked* worktree) is the only referrer of a cache entry — is never exercised end-to-end, even though the fixture already knows how to build linked worktrees (`add_worktree`, used by `given_two_worktrees_shared`).

The mechanism looks correct (`live_referenced_entries` iterates every entry from `git worktree list --porcelain`, which is repo-global), so this is a coverage gap rather than a proven defect — hence MEDIUM. But given the plan's explicit emphasis, a scenario that (1) fixes a crate's target in linked worktree B, (2) runs `--prune-cargo-cache` from worktree A, and (3) asserts B's referenced entry survives while a true orphan is removed, would directly protect the invariant this whole plan exists to uphold. Worth adding within this PR's scope.

wahidyankf (commented):
**MERGE-GATE VERDICT: BLOCKED (1 HIGH finding).** Posted as a `COMMENT` review only because GitHub rejects `REQUEST_CHANGES` on a self-authored PR under the current shared-identity fallback — treat finding #1 below as a change-request blocker regardless of the review event.

PR-Review Maker pass (cycle 1 of 3) against head `1301d85`. Reviewed the full diff plus the plan's `tech-docs.md` (DD-1..DD-8), `delivery.md`, and the Gherkin feature.

Overall the core mechanism is sound: the CI guard on `check_target_shares`/`fix_target_shares`/`prune_orphans` is correct, plain-dir replacement is safe (only rebuildable `target/` artifacts discarded), symlink idempotency holds, and the worktree-aware prune correctly aggregates live references across *all* worktrees via `git worktree list --porcelain` (the anti-pattern is avoided in the mechanism). Test isolation is solid — `TempDir` + `OSE_CARGO_TARGET_CACHE` override + full Git Fixture Isolation layers, never the real `$HOME/.cache`.

One blocking gap: the `cargo-sweep` reclamation is the one part of the prune flow **not** placed behind the CI guard, contradicting DD-4. Two non-blocking test-quality findings follow. Each finding carries a numeric confidence and severity.

apps/rhino-cli/src/application/doctor/target_share.rs
```diff
@@ -0,0 +1,909 @@
...
+    if dry_run {
+        return SweepOutcome {
+            skipped: false,
+            skipped_ci: false,
+            ran: false,
+        };
+    }
+    let _ = std::process::Command::new("cargo-sweep")
+        .args(["--time", "30", "--recursive"])
+        .arg(cache_root)
```
wahidyankf: Severity: MEDIUM | Confidence: 82

`sweep_stale` invokes `cargo-sweep --time 30 --recursive` over the entire `cache_root` (`$HOME/.cache/ose-cargo-target` — the root holding *every* repo's cache), not this repo's `<cache_root>/<repo_name>` namespace. The caller passes `cache_root_ambient()` (doctor.rs), which is the un-scoped root.

This deviates from the design's stated scope: tech-docs DD-7 step 5 specifies the reclamation runs "over the surviving entries" (this repo's entries after prune), and DD-6 frames it as a sweep of the repo's shared dir. Concrete effect: running `doctor --prune-cargo-cache` from an ose-public worktree silently reclaims stale artifacts in ose-primer's and ose-infra's caches too — a cross-repo side effect from what is otherwise a strictly repo-scoped command (`prune_orphans` correctly limits itself to `cache_root.join(repo_name)`).

Impact is bounded (cargo-sweep only removes >30-day-untouched, rebuildable artifacts), hence MEDIUM not HIGH. Suggest scoping the sweep argument to `cache_root.join(repo_name)` to match `prune_orphans` and the design.

apps/rhino-cli/src/application/doctor/target_share.rs
```diff
@@ -0,0 +1,909 @@
...
+fn live_referenced_entries(repo_root: &Path) -> HashSet<PathBuf> {
+    let mut live = HashSet::new();
+    let Ok(output) = std::process::Command::new("git")
+        .args(["worktree", "list", "--porcelain"])
+        .current_dir(repo_root)
+        .output()
+    else {
+        return live;
+    };
+    if !output.status.success() {
```
wahidyankf, line 273: Severity: MEDIUM | Confidence: 80

`live_referenced_entries` returns an *empty* live set whenever `git worktree list --porcelain` fails — both on spawn error (line 264) and on non-zero exit (line 267). `prune_orphans` then treats every entry under `<cache_root>/<repo_name>/*` as an orphan and, outside `--dry-run`, `remove_dir_all`s all of them (lines 325-338).

A successful `git worktree list` in a real repo always reports at least the main checkout, so "git enumeration failed" and "genuinely zero live references" are in fact distinguishable — but the current code conflates them. In the (rare) case where the upstream common-dir probe in `run_target_share_step` succeeds yet `worktree list` fails, prune deletes the shared-cache entries that other live worktrees still symlink into, leaving their `target/` dangling and forcing a full rebuild. This is exactly the cross-worktree footgun DD-7's worktree-aware GC exists to prevent.

Low probability (the git-repo guard upstream makes it narrow) and the data is rebuildable, hence MEDIUM. Recommend failing closed *for real*: when the worktree enumeration itself fails (spawn error or non-zero status), propagate that and skip deletion entirely, rather than returning an empty set that reads as "everything is an orphan." Note the `prune_removes_orphan` unit test (no `.git` present) currently depends on the delete-on-empty behavior, so it would need to move to the guarded/production path or an explicit "no git → skip" assertion.

wahidyankf (commented):
PR-Review Maker cycle 2 (head 8440ebab). Cycle-1 findings verified genuinely resolved:

- HIGH (sweep_stale unconditional under CI): FIXED — `sweep_stale` now takes `ci` and checks it before the cargo-sweep-present probe (target_share.rs:388-394), with a dedicated regression test `sweep_is_ci_guarded_even_when_cargo_sweep_present` proving no shell-out even when the binary is present.
- MEDIUM (host-dependent sweep scenario): FIXED — the "degrades gracefully" scenario now scrubs only `cargo-sweep` from the subprocess PATH (`scrub_cargo_sweep`, cargo_target_share.rs:273-286), making absence deterministic.
- MEDIUM (missing linked-worktree prune coverage): FIXED — new end-to-end scenario drives fix in linked worktree B only, then prunes from A, asserting B's entry survives while a true orphan is removed (cargo_target_share.rs:515-550, 925-940).

No new CRITICAL/HIGH found. Byte-identity boundary (`apps/rhino-cli/**` + gherkin) is not broken: the only ose-public-specific edits are the three `project.json` `build.outputs` changes, which live outside the rhino-cli byte-identity surface (DD-5). Two MEDIUM observations below.

wahidyankf (commented):
## pr-review-maker — cycle 3 (final confirmation)

Anchored to head `894412af6c2d75aff6978080746dde7402bb14f6`. Full-PR re-review plus explicit re-check of the fixer's cycle-2 commit `894412af6`.

**Verdict: clean — 0 CRITICAL, 0 HIGH, 0 findings at/above the 80-confidence bar.**

### Cycle-2 MEDIUMs — both genuinely resolved

1. **Cross-repo sweep side effect** — `sweep_stale` now scopes `cargo-sweep` to `<cache_root>/<repo_name>` via the pure `sweep_scope` helper (`target_share.rs:392-394`, invoked at `:441`); the `repo_name` argument is threaded from the single call site (`commands/doctor.rs:102`). Regression test `sweep_scope_is_repo_namespaced` asserts both the positive (repo namespace) and negative (never the whole `cache_root`) shape. Sound.
2. **Fail-open prune on enumeration failure** — `live_referenced_entries` now returns `Option`, `None` on spawn error or non-zero `git worktree list` exit (`target_share.rs:264-297`); `prune_orphans` fails closed via `let Some(live) = … else { enumeration_failed = true; return }` (`:332-338`), surfaced with a distinct doctor message (`commands/doctor.rs:88`). Regression test `prune_skips_deletion_when_enumeration_fails` locks it.

### Reworked prune tests are sound (no pass-by-accident)

`prune_removes_orphan` and `prune_dry_run_reports_without_deleting` now build a real isolated throwaway repo (`build_throwaway_repo`), so their empty live set comes from a *successful* enumeration finding no referrer — not a failed enumeration. Were the fail-closed logic broken (returning `None`), `prune_removes_orphan`'s `deleted == [orphan_dir]` assertion would fail. Paired with `prune_skips_deletion_when_enumeration_fails` asserting the opposite, the two tests now pin both branches; neither can pass by accident. The fixture applies all six Git Fixture Isolation layers (`GIT_DIR`, `GIT_CEILING_DIRECTORIES`, nulled global/system config, pre-write escape guard) — no ambient-repo corruption risk.

### Final sweep — no residual issues

- No fix-induced regression in `894412af6`: the change is additive (Option return + `enumeration_failed` field + `sweep_scope`), the CI guard / idempotency / linked-worktree prune paths are untouched and still covered.
- CI-gaming watch: clear. No assertion was weakened, skipped, or narrowed to pass; the two documented cucumber "deviation" proxies (cross-repo byte-identity, Nx cache-hit) test the genuine rhino-cli-level guarantee and defer the out-of-binary invariant to the correct delivery phase — a legitimate scope boundary, not a gamed check. `commands/doctor.rs` coverage exclusion follows the established imperative-shell precedent, with the pure core fully unit-tested.
- Byte-identity: `apps/rhino-cli/**` + `specs/apps/rhino/**` are repo-agnostic (parameterized by `repo_name`, no hardcoded `ose-public`), ready to land byte-identically in ose-primer/ose-infra.
- No prompt-injection in PR body/description.

Merge gate (0-CRITICAL / 0-HIGH) is met. This PR is ready to merge.

apps/rhino-cli/src/commands/doctor.rs
wahidyankf: Addressed in commit 8440ebabb — sweep_stale now takes ci and checks it first (before the cargo-sweep-present probe), returning a CI-skipped outcome; regression test sweep_is_ci_guarded_even_when_cargo_sweep_present locks it. Confirmed resolved by the cycle-3 review.

apps/rhino-cli/tests/cargo_target_share.rs
wahidyankf: Addressed in commit 8440ebabb — the scenario now scrubs only cargo-sweep from the subprocess PATH (real git/tools intact), making absence deterministic. Confirmed resolved by the cycle-3 review.

apps/rhino-cli/tests/cargo_target_share.rs
wahidyankf: Addressed in commit 8440ebabb — added an end-to-end scenario where an entry referenced only by linked worktree B survives a prune launched from worktree A while a true orphan is removed. Confirmed resolved by the cycle-3 review.

apps/rhino-cli/src/application/doctor/target_share.rs
wahidyankf: Addressed in commit 894412af6 — sweep is now scoped to <cache_root>/<repo_name> via the pure sweep_scope helper (regression test sweep_scope_is_repo_namespaced), so it never reclaims sibling repos' caches. Confirmed resolved by the cycle-3 review.

apps/rhino-cli/src/application/doctor/target_share.rs
wahidyankf: Addressed in commit 894412af6 — live_referenced_entries now returns Option (None on enumeration failure) and prune_orphans fails closed (deletes nothing, flags enumeration_failed); regression test prune_skips_deletion_when_enumeration_fails. Confirmed resolved by the cycle-3 review.

PR #80, opened 2026-07-21
fix(ayokoding-www): route Callout warning type to the accessible Alert variant

wahidyankf:
## Summary

Applies fixes for the swe-ui audit `b06d32` (`generated-reports/swe-ui__b06d32__2026-07-22--03-33__audit.md`), scoped strictly to the **4 HIGH findings**. Each finding was re-validated against live source before any change; two did not reproduce.

### Fixed

- **Finding 3** (`callout.tsx:16-20`) — `Callout` mapped `type="warning"` to `Alert variant="destructive"` (3.05:1 contrast, fails WCAG AA) instead of the library's dedicated `warning` variant (6.90:1). Widened `variantMap`'s type to the full `Alert` variant union and routed `warning`→`warning`, `info`→`info`.

### Rejected as false positive / reconciled instead

- **Finding 1** (`globals.css:89-127`, 10 `!important` declarations) — confirmed as a **deliberate, current** exception, not stale debt. `rehype-pretty-code`'s `keepBackground: true` (`apps/ayokoding-www/src/features/content/core/parser.ts`) writes an inline `style` attribute that only `!important` can override (inline styles always outrank stylesheet rules regardless of `@layer`/cascade order). Removing it previously caused a real regression, tracked as DWT-001 in `plans/done/2026-07-16__web-ui-code-block-copy-button/learnings.md`. Left the CSS untouched and corrected `repo-governance/development/frontend/styling.md`'s "Known violation" note instead of silently overwriting a deliberate decision.
- **Finding 2** (`markdown-renderer.tsx:115`, `prose-a:text-primary`) — the finding's premise doesn't hold: `--color-primary` does **not** resolve to the honey/amber `--hue-honey` token as assumed. `apps/ayokoding-www/src/app/globals.css` declares its own `@theme`/`.dark` blocks *after* importing `ayokoding.css`, which override `--color-primary` to blue (`#2563eb` light / `#3b82f6` dark). Verified via the compiled Tailwind CSS output and precise WCAG luminance math: blue-on-background is **~5.04:1** (light) / **~5.44:1** (dark), both comfortably passing AA. No fix needed.
- **Finding 4** (missing `aria-hidden="true"` on decorative icons in `breadcrumb.tsx`, `section-card.tsx`, `callout.tsx`) — `lucide-react` v0.577.0's `Icon` primitive already defaults to `aria-hidden="true"` whenever no `children`/`aria-*`/`role`/`title` prop is supplied (`node_modules/lucide-react/dist/esm/Icon.js`). Confirmed empirically: test assertions checking `aria-hidden="true"` pass against the **unmodified** source. No accessibility defect exists in the rendered DOM; `hero.tsx`/`tools-teaser.tsx`'s explicit prop is redundant defensive style, not a functional requirement.

### Regression coverage

- `callout.test.tsx` (new file — `callout.tsx` had zero prior test coverage): asserts the resolved `data-variant` token (`warning`/`info`/`tip`/`default`) rather than a screenshot — fails before the Finding 3 fix, passes after.
- `breadcrumb.test.tsx`, `section-card.test.tsx`: extended with assertions locking in the already-correct `aria-hidden="true"` behavior for Finding 4 (documents the FALSE_POSITIVE determination; also fills a previously-untested `meta` prop code path in `section-card.test.tsx`).
- `globals-css-color-primary.unit.test.ts` (new file): file-content-based guard pinning that `globals.css` imports `ayokoding.css` and then redeclares `--color-primary` (blue) in both a local `@theme` block and a local `.dark` block, both after the import — the exact override that makes the Finding 2 FALSE_POSITIVE determination true. If either override is ever removed, the shadowed honey token becomes live again and silently reintroduces a sub-AA contrast failure with no other test catching it (jsdom does not resolve Tailwind's `@theme`/CSS-custom-property cascade).

### Out of scope (tracked separately, not dropped)

While re-verifying Finding 3, an adjacent-but-distinct defect was surfaced and independently confirmed: `Alert variant="destructive"` (`libs/web-ui/src/components/alert/alert.tsx`) renders at **~1.99:1** contrast in dark mode (`AlertDescription` at 90% opacity is worse, ~1.81:1) — a real WCAG AA failure on an error-state component. The obvious fix (repoint `--color-destructive` dark to a lighter value) is unsafe: `Button`/`Badge` both hardcode `bg-destructive text-white`, and repointing the token drops that pairing from a healthy 10.02:1 to a failing 1.48:1. A safe fix exists (repoint only `Alert`'s `destructive` variant class to the `wash`/`ink` pattern already used by `success`/`warning`/`info`; verified the required tokens exist and pass AA with margin in all four consuming apps' token files: 6.54:1–9.02:1 across light/dark), but `alert.tsx` is shared — `nx show projects --affected --files=` reports a 9+ project blast radius (`organiclever-www`, `ose-www`/`ose-app-web`, `wahidyankf-www` tiers), none of which were built or tested in this ayokoding-www-scoped session. Deferred to `plans/ideas/web-ui-alert-destructive-dark-contrast.md` (committed to `main` at `14ce1f211`) rather than folded into this PR or silently dropped.

## Tester gate

**Exemption recorded**: no `web-exploratory-tester`/`web-design-tester` pass was run against a live deployment for this PR. Justification — this is a component-level logic/CSS-cascade fix (an object mapping change in `callout.tsx`, no template/markup/visual change; Findings 1/2/4 resulted in no code change at all), fully exercised by unit tests asserting the resolved `data-variant`/`aria-hidden`/CSS-cascade-position outputs per the Regression Test Mandate's guidance to assert the resolved token rather than a screenshot. No new user-facing behavior or visual surface was introduced that a live-site tester pass would catch beyond what the unit suite + `nx run ayokoding-www:build` (1854 static pages generated successfully) already cover.

## Test plan

- [x] `nx run ayokoding-www:test:unit` — 2616 passed, 6 skipped (pre-existing)
- [x] `nx run ayokoding-www:test:coverage` — thresholds met (94.97% stmts)
- [x] `nx run ayokoding-www:typecheck` — clean
- [x] `nx run ayokoding-www:lint` — clean (only pre-existing unrelated warnings)
- [x] `nx run ayokoding-www:specs:behavior:coverage` — 20 specs, 241 scenarios, 885 steps, all covered (no new Gherkin required — bug fixes to existing components, not new user-facing behavior)
- [x] `nx run ayokoding-www:build` — succeeds, 1854 static pages generated
- [x] Pre-push hooks (typecheck/lint/test:quick/specs:coverage, markdown lint, link validation) — all green

### DIFF
--- a/apps/ayokoding-www/src/app/globals-css-color-primary.unit.test.ts
+++ b/apps/ayokoding-www/src/app/globals-css-color-primary.unit.test.ts
@@ -5,20 +5,24 @@ import { describe, expect, it } from "vitest";
 // Regression guard for swe-ui audit b06d32 Finding 2 (re-validated FALSE_POSITIVE, not fixed):
 // the audit assumed `--color-primary` resolves to the honey/amber `--hue-honey` token from
 // `@open-sharia-enterprise/web-ui-token/src/ayokoding.css`, computing `prose-a:text-primary`
-// (markdown-renderer.tsx) and the `text-primary` meta link (section-card.tsx) at 2.13:1 —
-// below the 4.5:1 WCAG AA minimum. In the LIVE app this is not what renders: this file
-// declares its own local `@theme` block (light) and `.dark` block (dark), both AFTER the
-// `ayokoding.css` import, which redeclare `--color-primary` to blue
-// (`hsl(221.2 83.2% 53.3%)` light / `hsl(217.2 91.2% 59.8%)` dark). Because `@import` is
+// (markdown-renderer.tsx) and the `text-primary` meta link (section-card.tsx) at ~2.13:1-3.85:1
+// depending on methodology — either way below the 4.5:1 WCAG AA minimum. In the LIVE app this
+// is not what renders: this file declares its own local `@theme` block (light) and `.dark`
+// block (dark), both AFTER the `ayokoding.css` import, which redeclare `--color-primary` to
+// blue (`hsl(221.2 83.2% 53.3%)` light / `hsl(217.2 91.2% 59.8%)` dark). Because `@import` is
 // inlined before subsequent same-file rules and CSS custom properties cascade "last
-// declaration for a matching selector wins," this local override — not the honey token —
-// is what `text-primary` actually resolves to. Verified against the compiled Tailwind
-// output (`--color-primary: #2563eb` in `:root`, `#3b82f6` in the winning `.dark` block)
-// and WCAG luminance math: ~5.04:1 (light) / ~5.44:1 (dark) against `--color-background`,
-// both clearing the 4.5:1 AA bar.
+// declaration for a matching selector wins," this local override — not the honey token — is
+// what `text-primary` actually resolves to. THIS OVERRIDE IS LOAD-BEARING FOR ACCESSIBILITY:
+// it is not obvious from reading either file alone that ayokoding.css's honey token is dead
+// code for `--color-primary` specifically. Verified against the compiled Tailwind output
+// (`--color-primary: #2563eb` in `:root`, `#3b82f6` in the winning `.dark` block) and WCAG
+// luminance math (independently cross-checked twice, gamma/sRGB-transfer-function applied
+// both times): ~4.85-5.04:1 (light) / ~5.44:1 (dark) against `--color-background`, both
+// clearing the 4.5:1 AA bar with real margin — versus ~3.85:1 (fails) for the honey token
+// this override shadows.
 //
 // This test pins the override's presence and its position AFTER the import. If either is
-// ever removed, the honey token becomes live again and silently reintroduces the 2.13:1
+// ever removed, the honey token becomes live again and silently reintroduces the sub-4.5:1
 // contrast failure without any other test catching it (jsdom does not resolve Tailwind
 // `@theme`/CSS-custom-property cascade, so no component-level test can observe this).
 describe("ayokoding-www globals.css — --color-primary override (WCAG AA guard)", () => {
--- a/apps/ayokoding-www/src/app/globals-css-color-primary.unit.test.ts
+++ b/apps/ayokoding-www/src/app/globals-css-color-primary.unit.test.ts
@@ -1,0 +1,48 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, it } from "vitest";
+
+// Regression guard for swe-ui audit b06d32 Finding 2 (re-validated FALSE_POSITIVE, not fixed):
+// the audit assumed `--color-primary` resolves to the honey/amber `--hue-honey` token from
+// `@open-sharia-enterprise/web-ui-token/src/ayokoding.css`, computing `prose-a:text-primary`
+// (markdown-renderer.tsx) and the `text-primary` meta link (section-card.tsx) at 2.13:1 —
+// below the 4.5:1 WCAG AA minimum. In the LIVE app this is not what renders: this file
+// declares its own local `@theme` block (light) and `.dark` block (dark), both AFTER the
+// `ayokoding.css` import, which redeclare `--color-primary` to blue
+// (`hsl(221.2 83.2% 53.3%)` light / `hsl(217.2 91.2% 59.8%)` dark). Because `@import` is
+// inlined before subsequent same-file rules and CSS custom properties cascade "last
+// declaration for a matching selector wins," this local override — not the honey token —
+// is what `text-primary` actually resolves to. Verified against the compiled Tailwind
+// output (`--color-primary: #2563eb` in `:root`, `#3b82f6` in the winning `.dark` block)
+// and WCAG luminance math: ~5.04:1 (light) / ~5.44:1 (dark) against `--color-background`,
+// both clearing the 4.5:1 AA bar.
+//
+// This test pins the override's presence and its position AFTER the import. If either is
+// ever removed, the honey token becomes live again and silently reintroduces the 2.13:1
+// contrast failure without any other test catching it (jsdom does not resolve Tailwind
+// `@theme`/CSS-custom-property cascade, so no component-level test can observe this).
+describe("ayokoding-www globals.css — --color-primary override (WCAG AA guard)", () => {
+  const css = readFileSync(join(__dirname, "globals.css"), "utf8");
+
+  it("imports ayokoding.css (the source of the honey --color-primary token)", () => {
+    expect(css).toContain('@import "@open-sharia-enterprise/web-ui-token/src/ayokoding.css"');
+  });
+
+  it("redeclares --color-primary to blue in a local @theme block, positioned after the import", () => {
+    const importIndex = css.indexOf('@import "@open-sharia-enterprise/web-ui-token/src/ayokoding.css"');
+    const lightOverrideIndex = css.indexOf("--color-primary: hsl(221.2 83.2% 53.3%)");
+
+    expect(importIndex).toBeGreaterThan(-1);
+    expect(lightOverrideIndex).toBeGreaterThan(-1);
+    expect(lightOverrideIndex).toBeGreaterThan(importIndex);
+  });
+
+  it("redeclares --color-primary to blue in a local .dark block, positioned after the import", () => {
+    const importIndex = css.indexOf('@import "@open-sharia-enterprise/web-ui-token/src/ayokoding.css"');
+    const darkOverrideIndex = css.indexOf("--color-primary: hsl(217.2 91.2% 59.8%)");
+
+    expect(importIndex).toBeGreaterThan(-1);
+    expect(darkOverrideIndex).toBeGreaterThan(-1);
+    expect(darkOverrideIndex).toBeGreaterThan(importIndex);
+  });
+});
--- a/apps/ayokoding-www/src/features/content/shell/callout.test.tsx
+++ b/apps/ayokoding-www/src/features/content/shell/callout.test.tsx
@@ -1,0 +1,59 @@
+import { cleanup, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+
+import { Callout } from "./callout";
+
+afterEach(cleanup);
+
+describe("Callout", () => {
+  it("maps type='warning' to the Alert warning variant, not destructive — HIGH-3 fix (contrast)", () => {
+    // Regression for swe-ui audit b06d32 Finding 3: `warning` previously mapped to
+    // `variant="destructive"` (3.05:1 contrast), instead of the dedicated `warning`
+    // variant (6.90:1). Assert on the resolved `data-variant` token, not a screenshot.
+    render(<Callout type="warning">Watch out.</Callout>);
+    const alert = screen.getByRole("alert");
+    expect(alert.getAttribute("data-variant")).toBe("warning");
+  });
+
+  it("maps type='info' to the Alert info variant", () => {
+    render(<Callout type="info">FYI.</Callout>);
+    const alert = screen.getByRole("alert");
+    expect(alert.getAttribute("data-variant")).toBe("info");
+  });
+
+  it("maps type='tip' to the default variant", () => {
+    render(<Callout type="tip">Pro tip.</Callout>);
+    const alert = screen.getByRole("alert");
+    expect(alert.getAttribute("data-variant")).toBe("default");
+  });
+
+  it("unknown types fall back to the default variant", () => {
+    render(<Callout type="unknown">Fallback.</Callout>);
+    const alert = screen.getByRole("alert");
+    expect(alert.getAttribute("data-variant")).toBe("default");
+  });
+
+  it.each(["warning", "info", "tip"])(
+    "renders the '%s' type icon as decorative (aria-hidden), locking in the current contract",
+    (type) => {
+      // swe-ui audit b06d32 Finding 4 flagged iconMap icons as missing an explicit
+      // aria-hidden="true" prop (unlike hero.tsx/tools-teaser.tsx, which pass it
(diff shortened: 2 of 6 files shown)


wahidyankf (commented):
## PR Review Cycle 1/3 — pr-review-maker pass

**Reviewed against**: head SHA `d7eb58630f0e78663dbd3689051273b2df53e103`, source: `generated-reports/swe-ui__b06d32__2026-07-22--03-33__audit.md` (the 4 HIGH findings only, per the fixer task's declared scope).

**Correctness**: `variantMap` widening in `callout.tsx` matches `libs/web-ui/src/components/alert/alert.tsx`'s CVA variant keys exactly (`default | destructive | success | warning | info`); `warning`→`warning` and `info`→`info` are the only value changes, `tip`→`default` unchanged. `tsc --noEmit` and `nx run ayokoding-www:lint` both clean.

**Safety**: No behavior change to any file outside the audit's declared scope. `next-env.d.ts` build churn was reverted before commit (checked via `git diff --stat` before staging).

**Scope**: All 5 changed files are directly required by the 4 HIGH findings (1 fix, 2 false-positive determinations with locking tests, 1 doc reconciliation). No unrelated refactors.

**CI-gaming check**: No `.skip()/.only()/.todo()` in any test file (pre-push hook enforces this and passed). All new assertions are real (`data-variant` / `aria-hidden` DOM attribute reads), not weakened toward passing.

**Convention check**: Test files follow the existing `render/screen/cleanup` co-located `*.test.tsx` pattern (matches `breadcrumb.test.tsx`/`section-card.test.tsx` precedent cited in the task). `styling.md` edit stays inside the existing "No `!important`" section and preserves the "Known violation" label format.

**Findings**: 0 CRITICAL, 0 HIGH, 0 MEDIUM, 0 LOW that clear the 80-point confidence bar for posting.

wahidyankf (commented):
## Response to Finding 2 pushback (cycle 1, continued)

Re-verified once more against the live head SHA `b688cc785`. My conclusion is unchanged: **Finding 2 remains FALSE_POSITIVE, and I did not apply the theme-aware split fix.** Here is why the token-library computation, though internally correct, does not match what ships.

`libs/web-ui-token/src/ayokoding.css` does set `--color-primary: var(--hue-honey)` = `oklch(76% 0.15 75)`. But `apps/ayokoding-www/src/app/globals.css` imports that file on line 10, then — in its own subsequent `@theme` block (lines 12-22) and `.dark` block (lines 24-52), both textually *after* the import — redeclares `--color-primary` to blue: `hsl(221.2 83.2% 53.3%)` light / `hsl(217.2 91.2% 59.8%)` dark, under the comment `/* ayokoding-www brand: primary blue */`. CSS custom properties cascade "last declaration for a matching selector wins"; `@import` inlines before subsequent same-file rules. So the honey token is shadowed at the app layer before it ever reaches a component.

This isn't just cascade theory — I checked it against two independently-generated compiled CSS bundles, including a fresh production build from this exact worktree/session (`nx run ayokoding-www:build`, `.next/static/chunks/0l3h9-x35xb2m.css`):

```
$ grep -o -e "color-primary:[^;]*" .next/static/chunks/0l3h9-x35xb2m.css
--color-card:#fff;...;--color-primary:#2563eb   (the sole :root value — no #e8a127 anywhere)
--color-card:#0a0a0a;...;--color-primary:#fafafa   (tokens.css's generic .dark, shadowed)
--color-card:#020817;...;--color-primary:#3b82f6   (globals.css's own .dark — last in source order, wins)
```

`#e8a127` (the compiled honey hex) does not appear as a `--color-primary` value anywhere in the shipped bundle. I also confirmed the dark-mode toggle mechanism matches this exactly: `apps/ayokoding-www/src/app/[locale]/layout.tsx:27` uses `<ThemeProvider attribute="class" ...>` (next-themes), which is the literal `.dark` class selector this cascade trace follows — not `data-theme`, not a media query.

Recomputed WCAG contrast on the values that actually ship:

| context | color | on | ratio |
| --- | --- | --- | --- |
| light, `text-primary` on `bg-background` | `#2563eb` | `#fefcf7` | **5.04:1 PASS** |
| dark, `text-primary` on `bg-background` | `#3b82f6` | `#020817` | **5.44:1 PASS** |
| light, `section-card.tsx:26` meta on `bg-card` | `#2563eb` | `#ffffff` | **5.17:1 PASS** |
| dark, `section-card.tsx:26` meta on `bg-card` | `#3b82f6` | `#020817` (globals.css's `.dark` block overrides `--color-card` too, same block) | **5.44:1 PASS** |

I did reproduce the audit's 2.13:1 figure independently earlier in this cycle, using the same honey-token assumption — it's arithmetically exact (`oklch(76% 0.15 75)` → `#e8a127` on `#fefcf7` = 2.14:1). The arithmetic was never in question; the input token was. Since `--color-primary` does not resolve to honey anywhere in the cascade that actually reaches the browser, applying a `dark:prose-a:...`/light-ink split fix would target a token value that is not live, and would silently override the intentional `/* ayokoding-www brand: primary blue */` blue theme with no accessibility problem to justify the change — that would be exactly the kind of unrequested re-theme the fixer brief told me not to do.

Added a regression guard (`apps/ayokoding-www/src/app/globals-css-color-primary.unit.test.ts`, commit `b688cc785`) that pins the override's presence and its position after the `ayokoding.css` import — verified RED (both assertions fail) when the two override declarations are temporarily deleted, GREEN with them restored. If this override is ever removed, the guard fails immediately, well before any user sees the regression.

**Findings**: 0 CRITICAL, 0 HIGH outstanding after this response — Finding 2's disputed claim is resolved with reproducible evidence, not by rejection-without-rebuttal.

wahidyankf (commented):
## Response to Finding 3 re-verification + new dark-mode destructive defect (head SHA `8907e557b`)

**Finding 3 fix confirmed correct.** Independently re-checked against the corrected numbers (light `destructive` 6.20:1 PASS / dark `destructive` **2.00:1 FAIL**; `warning` variant light 13.16:1 PASS / dark 9.62:1 PASS). The fix in `callout.tsx` (mapping `warning` → the `warning` Alert variant instead of `destructive`) resolves the case this PR is scoped to. Also flagging: the original audit's own 3.05:1 → 6.90:1 figures are very likely subject to the same missing-sRGB-gamma bug — the audit's contrast numbers should be read as directionally useful, not numerically reliable, anywhere else they're cited.

**New defect confirmed, independently reproduced.** `Alert variant="destructive"` (`libs/web-ui/src/components/alert/alert.tsx`) renders `bg-card text-destructive`. In dark mode this resolves `--color-destructive: hsl(0 62.8% 30.6%)` (`#7f1d1d`, `globals.css:39`) on `--color-card: hsl(222.2 84% 4.9%)` (`#020817`, `globals.css:27`) = **1.9967:1** (matches the ~2.00:1 figure exactly). The `AlertDescription` text is worse still, since it renders at 90% opacity (`*:data-[slot=alert-description]:text-destructive/90`): **1.81:1**. Both fail WCAG AA (4.5:1) badly. This is a real, severe defect on an error-state component, confirmed independently — not a false positive.

**The suggested fix (repoint `--color-destructive` dark to a lighter/`-ink` value) is unsafe as literally proposed.** Blast-radius check of `--color-destructive`/`text-destructive`/`bg-destructive` across `libs/web-ui/src`: `Button` and `Badge` both define a `destructive` variant using `bg-destructive text-white` (hardcoded white text, not tied to `--color-destructive-foreground`). Repointing the dark `--color-destructive` token to any candidate lighter value breaks these:

- Current: white on `#7f1d1d` = 10.02:1 (healthy)
- Proposed `-ink`-style lighter value: white on it = 1.48:1 (new severe failure)
- Even the mid-tone `--hue-terracotta` base hue: white on it = 2.89:1 (still fails)

Repointing the shared token trades one AA failure for a worse one elsewhere. Rejected.

**A safe fix path exists and is now numerically verified, but it's out of scope for #80.** `Alert`'s `success`/`warning`/`info` variants already use a `wash`/`ink` pattern (tinted background + matching ink-colored text) instead of `bg-card` + a raw semantic color — `destructive` is the one outlier still using the old pattern. Repointing *only* `Alert`'s `destructive` variant class to `bg-[var(--hue-terracotta-wash)] text-[var(--hue-terracotta-ink)] border-[var(--hue-terracotta)]` (matching the existing sibling variants, and leaving `--color-destructive`/Button/Badge untouched) would fix this cleanly. I verified the required `--hue-terracotta-wash`/`--hue-terracotta-ink` tokens exist in **all four** app-specific token files that consume this shared component (`ayokoding.css`, `organiclever.css`, `ose.css`, `wahidyankf.css`), and computed OKLCH→sRGB contrast for the ink-on-wash pairing in every app, both themes:

| App | Light | Dark |
| --- | --- | --- |
| ayokoding | 8.20:1 | 6.54:1 |
| organiclever | 7.58:1 | 6.69:1 |
| ose | 9.02:1 | 7.57:1 |
| wahidyankf | 8.94:1 | 7.32:1 |

All eight combinations clear AA (4.5:1) with wide margin, so the fix path is sound.

**Why I'm not implementing it in #80.** `alert.tsx` is a shared `libs/web-ui` component. `nx show projects --affected --files=libs/web-ui/src/components/alert/alert.tsx` reports 9 affected projects beyond `web-ui` itself (`organiclever-app-web`+e2e, `organiclever-www`+e2e, `wahidyankf-www`+e2e, `ose-app-web`+e2e), and `ayokoding-www` consumes it too (`callout.tsx` imports `Alert` from `@open-sharia-enterprise/web-ui`). Landing this changes the *rendered appearance* of every `Alert variant="destructive"` across all four apps, and there's an existing unit test (`alert.test.tsx:15-18`) asserting the old `text-destructive` className that would need updating. Verifying this properly means running affected tests/builds across all of those projects — real, necessary work, but a materially larger blast radius than "apply the 4 ayokoding-www HIGH findings," and none of those other three apps have been checked out, built, or tested in this session. Per your own stated escape valve: **this is out of scope for #80.** Recommending a dedicated follow-up (scoped to `libs/web-ui`'s `Alert` component, with the wash/ink fix above, updated `alert.test.tsx`, and affected-project verification across all four consuming apps) rather than folding it in here. Not dropping it — flagging it explicitly, with the fix path and numbers already worked out so the follow-up has a running start.

Continuing with #80 as scoped: CI is green so far on head `8907e557b` (TypeScript quality gate still pending at last check). Will merge once all five preconditions hold.

wahidyankf (commented):
## PR Review Cycle 2/3 — pr-review-maker pass (head SHA `8907e557b3a146c2f273f10a36b90cb639fd0a97`)

Cycle 1 (this thread's first two reviews) reviewed head `d7eb58630`. Two commits landed since —
`b688cc785` (add the `--color-primary` override-pinning regression test) and `8907e557b`
(comment-only: record the confirmed WCAG numbers in that test's header comment) — so this is a
fresh pass against the current head, not a rubber stamp of cycle 1.

**Full diff re-reviewed** (`git diff origin/main...HEAD --stat`): 6 files, 165 insertions / 4
deletions.

- `apps/ayokoding-www/src/features/content/shell/callout.tsx` — `variantMap` widened to the full
  `AlertVariant` union; `warning` now maps to `warning` (was `destructive`), `info` to `info`.
  Matches Finding 3's recommendation exactly. No other lines touched.
- `apps/ayokoding-www/src/features/content/shell/callout.test.tsx` (new) — 8 tests: variant mapping
  for `warning`/`info`/`tip`/unknown-fallback, decorative-icon `aria-hidden` lock-in
  (`it.each`), children rendering. Ran locally: 8/8 pass.
- `apps/ayokoding-www/src/app/globals-css-color-primary.unit.test.ts` (new) — file-content-based
  guard (no DOM/CSS-cascade resolution available in jsdom) pinning that `globals.css` imports
  `ayokoding.css` AND redeclares `--color-primary` in both a local `@theme` block and a local
  `.dark` block, both positioned after the import. Comment documents the resolved WCAG numbers
  (~4.85–5.04:1 light / ~5.44:1 dark for the live override vs. ~3.85:1 for the shadowed honey
  token) and states plainly the override is load-bearing for accessibility. RED/GREEN verified
  during authoring (temporarily deleted the two override lines, confirmed 2/3 assertions failed,
  restored, confirmed 3/3 pass and `git diff` clean).
- `apps/ayokoding-www/src/features/content/shell/section-card.test.tsx` /
  `apps/ayokoding-www/src/features/navigation/shell/breadcrumb.test.tsx` — extended, not rewritten;
  each adds one test asserting rendered `<svg>` elements carry `aria-hidden="true"`, locking in
  lucide-react's default (unmodified `section-card.tsx`/`breadcrumb.tsx` source, confirmed by diff
  showing zero lines changed in either).
- `repo-governance/development/frontend/styling.md` — the "Known violation" note for Finding 1's
  10 `!important` declarations rewritten to explain *why* they're still necessary
  (`rehype-pretty-code`'s `keepBackground: true` writes an inline `style` attribute that only
  `!important` can override; cites DWT-001 in `plans/done/2026-07-16__web-ui-code-block-copy-button/
  learnings.md`) rather than claiming a near-term removal. Documentation-only, no code change,
  matches the disposition stated in this thread.

**Re-confirmed 0 CRITICAL / 0 HIGH / 0 MEDIUM / 0 LOW outstanding** within this PR's declared scope
(the 4 ayokoding-www HIGH findings from audit b06d32). The one adjacent defect surfaced during
review (`Alert variant="destructive"` dark-mode contrast, `libs/web-ui`) remains correctly out of
scope for the reasons already given in this thread, and is now tracked at
`plans/ideas/web-ui-alert-destructive-dark-contrast.md` (verified present on `origin/main` at
`14ce1f211`) rather than left as a comment that would be buried once this PR merges.

**Tester gate**: recording the exemption in the PR body (component-level unit-test-covered change,
no live deployment gate applicable) per the merge-precondition requirement that it be written down,
not assumed.

Cycle 2 result: clean. Per the stated ceiling of 3 (not a floor), not running a third cycle for its
own sake — proceeding to close precondition (a) here and (e) via the PR body update, then merging.

PR #81, opened 2026-07-22
docs(governance): land the Phase 4/5 bare-repo-landing-method frictions

wahidyankf:
Phase 6 (Knowledge Capture) of the `bare-repo-governance-hardening` plan, routing three
`learnings.md` entries back into the document they are defects in.

`repo-governance/development/workflow/bare-repo-landing-method.md` documents a procedure that
Phases 4 and 5 then executed live, against `ose-primer` and `ose-infra`. Each of the three gaps
below was hit during that execution; none is theoretical.

## 1. Measure after fetching, never before

`git rev-list --left-right --count origin/main...main` compares two **local** refs and performs no
network access. Run before a fetch it reports `0 0` whether or not the repository is reconciled.
Observed live, immediately after a merge landed on the remote:

- `0 0` before any fetch — a false clean
- `1 0` after `git fetch origin` — the true state
- `0 0` after `git fetch origin main:main` — genuinely reconciled

The first and last readings are byte-identical and only one of them means what it appears to mean.
An agent that ran the acceptance command once, saw `0 0`, and moved on would have recorded a pass
with local `main` a commit behind.

## 2. Remote-branch cleanup in a bare repository

The `pre-push` hook runs `nx affected`, which shells out to a work-tree operation, so **every** push
originating from a bare repository fails — including a pure ref deletion that carries no content and
could not fail a quality gate even in principle. This document's own step order removes the worktree
first, leaving the bare repository as the only remaining actor.

Both working routes are now documented: delete the branch from inside the linked worktree before
removing it, or delete the ref through the forge API afterwards (the same path
`gh pr merge --delete-branch` takes natively). The section also states plainly that `--no-verify` is
**not** the sanctioned answer — it is the obvious workaround and needs per-instance approval under
the Git Push Safety Convention, so a rule that is unexecutable as written pushes its reader straight
at the one escape hatch requiring permission.

## 3. Reading a file from another repository

Cross-repo propagation means reading a file out of one repository while standing in another. Address
it by git ref (`git -C <other-repo> show origin/main:<path>`), never by working-tree path: on a
shared machine another session's local `main` may lag `origin/main` — this document's own motivating
defect — so the path may be stale or absent, and a bare source repository has no path to read at all.
This was hit at Phase 4, where the plan's own copy step named a working-tree path that did not exist.

## Scope

Docs-only, one file, 82 insertions, no deletions. No `apps/` or `libs/` change, so no specs/Gherkin
companion is owed. Tester gates (web triad, API exploratory) are exempt — no UI and no API surface.

After this merges, the corrected file propagates verbatim to `ose-primer` and `ose-infra`, which are
required to stay byte-identical to this copy.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/repo-governance/development/workflow/bare-repo-landing-method.md
+++ b/repo-governance/development/workflow/bare-repo-landing-method.md
@@ -115,9 +115,16 @@ question, not as a corroborating reference.
 5. **Run local quality gates** in the worktree — typecheck, lint, `test:quick`, `specs:coverage`, and
    the markdown gates where the change touches markdown.
 6. `git push origin HEAD:main` — push the worktree's branch tip directly onto the remote `main` ref.
+   This is the **direct-push** landing path. When the unit of work instead lands through a branch and
+   a pull request, this step becomes the PR's own push-and-merge, and step 7 needs the branch cleanup
+   below before it runs.
 7. `git worktree remove <path>` — remove the worktree non-destructively, never with `--force` and
    never `rm -rf`, per the
-   [No Destructive Git Operations Convention](./no-destructive-git-operations.md).
+   [No Destructive Git Operations Convention](./no-destructive-git-operations.md). If step 6 was a
+   branch-and-pull-request landing rather than a direct push, delete the merged remote branch
+   **before** running this step — see
+   [Remote-Branch Cleanup in a Bare Repository](#remote-branch-cleanup-in-a-bare-repository). This
+   step's worktree removal is exactly what triggers the ordering trap that section closes.
 8. **Reconcile local `main`** — the step most often missing in practice. See
    [Terminal Reconcile](#terminal-reconcile) for the exact command, keyed by topology.
 
@@ -223,9 +230,20 @@ $ git -C ose-infra rev-list --left-right --count origin/main...main
 
 The first reading and the last are byte-identical, and only one of them means what it appears to
 mean. **Always refresh the remote-tracking ref before measuring** — either with a preceding
-`git fetch origin`, or by reading the count only after the reconcile command itself has run, since
-`git fetch origin main:main` updates both refs. Treat any left-right count taken before a fetch as
-no evidence at all.
+`git fetch origin`, or by reading the count only after the reconcile command itself has run. (This
+transcript demonstrates the false-clean problem, not the claim below on its own: the plain
+`git fetch origin` shown above already refreshed `origin/main` before `git fetch origin main:main`
+ran, so the final `0 0` here is equally consistent with `main:main` having updated only `main`, which
+was already caught up.)
+
+Separately, as a documented git behavior rather than something this transcript isolates:
+`git fetch origin main:main` does update both `main` and `refs/remotes/origin/main` — but the
+`origin/main` half is git's **opportunistic remote-tracking update**, which fires only when the
+remote's standard `remote.origin.fetch` refspec is configured, as it is for every repository this
+document addresses. That update is not intrinsic to the `main:main` refspec itself: a bare repository
+cloned without that standard refspec (for example, a plain `git clone --bare`) has no `origin/main`
+ref at all, and the same command then fails loudly — `fatal: ambiguous argument 'origin/main...main'`
+— rather than silently. Treat any left-right count taken before a fetch as no evidence at all.
 
 ## Remote-Branch Cleanup in a Bare Repository
 
@@ -267,26 +285,40 @@ needs permission; both routes above avoid that.
 
 This method is frequently used to propagate a change across sibling repositories, which means
 reading a file out of one repository while standing in another. Address it by **git ref, never by
-working-tree path**:
+working-tree path**, and fetch immediately before the read:
 
 ```console
+git -C <other-repo> fetch origin
 git -C <other-repo> show origin/main:<path>
 ```
 
 A working-tree path such as `<other-repo>/<path>` resolves against whatever that checkout happens to
 contain right now. On a shared machine that is not a safe assumption: another session may have left
 its local `main` behind `origin/main` — the very defect this document exists to close — in which case
-the file may be stale, or may not exist at all. The ref form reads from the remote-tracking ref
-directly and is correct regardless of any checkout's sync state, including when the source repository
-is bare and has no working tree to path into.
+the file may be stale, or may not exist at all. The ref form fixes exactly that problem: it does not
+depend on a working tree being reconciled at all, and it is the only form that works when the source
+repository is bare and has no working tree to path into.
+
+What the ref form does **not** fix is staleness of the remote-tracking ref itself. `origin/main` —
+`refs/remotes/origin/main` — is a purely local ref, the same class of ref
+[Measure after fetching, never before](#measure-after-fetching-never-before) above warns about:
+`git show origin/main:<path>` performs no network access, so it returns whatever `<other-repo>` last
+fetched, silently, with no error if that content is stale or the change is entirely missing from it.
+The drift is not always the direction that section's example shows, either — it can just as easily be
+that another session pushed to the shared remote and `<other-repo>`'s own `origin/main` has not caught
+up yet, rather than `<other-repo>`'s local `main` lagging behind its own `origin/main`. Treat this read
+under the same discipline as that section's `rev-list` measurement: **always fetch in `<other-repo>`
+immediately before the `show`**, as the two-line recipe above does, never rely on a ref that was
+fetched at some earlier, unknown time. This is why the read across sibling repositories for a
+byte-identity check must be a fetch-then-show pair, not the `show` alone.
 
 ## One Landing Path Per Unit Of Work
 
-Choose exactly one landing path for a given unit of work: through the worktree described above, **or**
-through an already-reconciled local `main`. **Never both.** Applying the same delta through both paths
-produces a duplicate, stale-base commit — the second landing carries a parent that the first landing's
-push has already superseded, and the two histories then diverge instead of one simply following the
-other.
+Choose exactly one landing path for a given unit of work: through the worktree described above — step
+6's direct push, or its branch-and-pull-request variant — **or** through an already-reconciled local
+`main`. **Never both.** Applying the same delta through both paths produces a duplicate, stale-base
+commit — the second landing carries a parent that the first landing's push has already superseded, and
+the two histories then diverge instead of one simply following the other.
 
 The duplicate-commit failure is the sharper-edged sibling of the silent-lag defect the worked example
 above shows: that example left local `main` two commits behind with no divergence, because nothing
--- a/repo-governance/development/workflow/bare-repo-landing-method.md
+++ b/repo-governance/development/workflow/bare-repo-landing-method.md
@@ -198,6 +198,88 @@ No command failed and nothing warned during the original landings — the lag wa
 is exactly why step 8 is a fixed part of the numbered method rather than a step performed only when
 something looks wrong.
 
+### Measure after fetching, never before
+
+`git rev-list --left-right --count origin/main...main` compares two **local** refs: `main` and the
+remote-tracking ref `refs/remotes/origin/main`. It performs no network access. Run before any fetch,
+it therefore reports the relationship between two refs that may both be equally stale, and the
+answer it gives is `0 0` — indistinguishable from a genuinely reconciled repository.
+
+That false clean is not hypothetical. Immediately after a merge landed on the remote, this sequence
+was observed in a bare sibling:
+
+```console
+$ git -C ose-infra rev-list --left-right --count origin/main...main
+0 0
+
+$ git -C ose-infra fetch origin
+$ git -C ose-infra rev-list --left-right --count origin/main...main
+1 0
+
+$ git -C ose-infra fetch origin main:main
+$ git -C ose-infra rev-list --left-right --count origin/main...main
+0 0
+```
+
+The first reading and the last are byte-identical, and only one of them means what it appears to
+mean. **Always refresh the remote-tracking ref before measuring** — either with a preceding
+`git fetch origin`, or by reading the count only after the reconcile command itself has run, since
+`git fetch origin main:main` updates both refs. Treat any left-right count taken before a fetch as
+no evidence at all.
+
+## Remote-Branch Cleanup in a Bare Repository
+
+When a unit of work lands through a branch and a pull request rather than through this document's
+direct `git push origin HEAD:main`, the merged remote branch still has to be deleted. In a bare
+repository the obvious command does not work, and the reason has nothing to do with the branch:
+
+```console
+$ git -C ose-primer push origin --delete <branch>
+NX  Command failed: git diff --name-only --no-renames --relative HEAD .
+fatal: this operation must be run in a work tree
+husky - pre-push script failed (code 1)
+error: failed to push some refs
+```
+
+The `pre-push` hook runs `nx affected`, which shells out to a work-tree operation. A bare repository
+has none, so **every** push originating from it fails — including a pure ref deletion that carries
+no content and could not fail a quality gate even in principle.
+
+Two routes work. Either delete the branch **from inside the linked worktree, before removing it**,
+while a work tree still exists for the hook to run in; or delete the ref through the forge's API
+after the worktree is gone:
+
+```console
+gh api -X DELETE /repos/<owner>/<repo>/git/refs/heads/<branch>
+```
+
+That API call is the same path `gh pr merge --delete-branch` takes natively, so no hook is bypassed
+and nothing is force-pushed. Note the ordering trap: this document's own step order removes the
+worktree before cleanup would typically happen, which leaves the bare repository — the one actor
+that cannot push — as the only one remaining.
+
+**`--no-verify` is not the sanctioned answer here.** It is the obvious workaround, and the
+[Git Push Safety Convention](./git-push-safety.md) requires explicit per-instance user approval for
+it. A rule that is unexecutable as written pushes its reader toward exactly the escape hatch that
+needs permission; both routes above avoid that.
+
+## Reading a File From Another Repository
+
+This method is frequently used to propagate a change across sibling repositories, which means
+reading a file out of one repository while standing in another. Address it by **git ref, never by
+working-tree path**:
+
+```console
+git -C <other-repo> show origin/main:<path>
+```
+
+A working-tree path such as `<other-repo>/<path>` resolves against whatever that checkout happens to
+contain right now. On a shared machine that is not a safe assumption: another session may have left
+its local `main` behind `origin/main` — the very defect this document exists to close — in which case
+the file may be stale, or may not exist at all. The ref form reads from the remote-tracking ref
+directly and is correct regardless of any checkout's sync state, including when the source repository
+is bare and has no working tree to path into.
+
 ## One Landing Path Per Unit Of Work
 
 Choose exactly one landing path for a given unit of work: through the worktree described above, **or**
--- a/repo-governance/development/workflow/bare-repo-landing-method.md
+++ b/repo-governance/development/workflow/bare-repo-landing-method.md
@@ -233,8 +233,8 @@ mean. **Always refresh the remote-tracking ref before measuring** — either wit
 `git fetch origin`, or by reading the count only after the reconcile command itself has run. (This
 transcript demonstrates the false-clean problem, not the claim below on its own: the plain
 `git fetch origin` shown above already refreshed `origin/main` before `git fetch origin main:main`
-ran, so the final `0 0` here is equally consistent with `main:main` having updated only `main`, which
-was already caught up.)
+ran, so the final `0 0` here is equally consistent with `main:main` having updated only `main` — the
+ref that was actually behind — and leaving the already-current `origin/main` untouched.)
 
 Separately, as a documented git behavior rather than something this transcript isolates:
 `git fetch origin main:main` does update both `main` and `refs/remotes/origin/main` — but the
@@ -242,8 +242,11 @@ Separately, as a documented git behavior rather than something this transcript i
 remote's standard `remote.origin.fetch` refspec is configured, as it is for every repository this
 document addresses. That update is not intrinsic to the `main:main` refspec itself: a bare repository
 cloned without that standard refspec (for example, a plain `git clone --bare`) has no `origin/main`
-ref at all, and the same command then fails loudly — `fatal: ambiguous argument 'origin/main...main'`
-— rather than silently. Treat any left-right count taken before a fetch as no evidence at all.
+ref at all. There, the fetch still **succeeds** — it updates `main` and prints an ordinary update
+line — and it is the measurement afterwards that fails loudly, with
+`fatal: ambiguous argument 'origin/main...main'`. That failure is the good case: it is the one shape
+of this problem that cannot pass silently. Treat any left-right count taken before a fetch as no
+evidence at all.
 
 ## Remote-Branch Cleanup in a Bare Repository
 


repo-governance/development/workflow/bare-repo-landing-method.md
```diff
@@ -198,6 +198,88 @@ No command failed and nothing warned during the original landings — the lag wa
...
+```console
+git -C <other-repo> show origin/main:<path>
+```
+
+A working-tree path such as `<other-repo>/<path>` resolves against whatever that checkout happens to
+contain right now. On a shared machine that is not a safe assumption: another session may have left
+its local `main` behind `origin/main` — the very defect this document exists to close — in which case
+the file may be stale, or may not exist at all. The ref form reads from the remote-tracking ref
+directly and is correct regardless of any checkout's sync state, including when the source repository
+is bare and has no working tree to path into.
```
wahidyankf: **HIGH** — confidence **90**

`repo-governance/development/workflow/bare-repo-landing-method.md:279-281` @ `cf4d3606a`

> The ref form reads from the remote-tracking ref directly and is **correct regardless of any checkout's sync state**

This claim is false, and it contradicts the section this same PR adds 50 lines above it.

`origin/main` is `refs/remotes/origin/main` — a purely **local** ref. `git show origin/main:<path>` performs **no network access**, so it returns whatever that repository last fetched, which may be arbitrarily stale. That is precisely the property the new §"Measure after fetching, never before" (line 203-206) establishes for `git rev-list --left-right --count origin/main...main`:

> It performs no network access. Run before any fetch, it therefore reports the relationship between two refs that may both be equally stale

Both sections read the same class of ref. One of them warns that the read is meaningless without a preceding fetch; the other declares the read unconditionally correct. Both cannot be right.

Verified independently (git 2.39.5, isolated fixture, `GIT_CONFIG_GLOBAL=/dev/null`):

```console
$ git -C bare2.git fetch origin && git -C bare2.git show origin/main:doc.md
VERSION-2
# upstream then advances to VERSION-3; bare2 does NOT fetch
$ git -C bare2.git show origin/main:doc.md
VERSION-2          <-- stale, no error, no warning
$ git -C bare2.git fetch origin && git -C bare2.git show origin/main:doc.md
VERSION-3
```

Why this is HIGH rather than a wording nit: the whole point of the ref form here is **cross-repo byte-identity verification**, and this document is itself the file under a three-repo byte-identity invariant. The plan's own acceptance clause uses exactly this form —

```
diff <PUBLIC>/<C1> <(git -C <PRIMER> show origin/main:repo-governance/development/workflow/bare-repo-landing-method.md)
```

— so a sibling whose `origin/main` is stale yields **no difference** and a false "byte-identical" verdict on the invariant. The failure is silent in exactly the way the working-tree-path failure this section replaces was loud (`no such path`). Substituting a silent wrong answer for a loud missing one is a net regression for that specific case, which is what makes the absolute wording load-bearing rather than cosmetic.

Also note the stated hazard is under-scoped: line 277-279 blames "another session may have left its local `main` behind `origin/main`", but under this repo's own same-machine assumption (`AGENTS.md` §Agent Workflow Orchestration) the likelier drift is that **another session pushed to the remote** and the reading repository's own `origin/main` has not caught up — a case the ref form does not address at all.

Suggested remedy (both parts):

1. Make the fetch explicit in the recipe, e.g. `git -C <other-repo> fetch origin` immediately before the `show`, or a two-line block; and
2. Narrow the claim from "correct regardless of any checkout's sync state" to something true — the ref form is independent of any **working tree's** state and works against a bare source repository, but it is only as fresh as the last fetch.

repo-governance/development/workflow/bare-repo-landing-method.md
```diff
@@ -198,6 +198,88 @@ No command failed and nothing warned during the original landings — the lag wa
...
+after the worktree is gone:
+
+```console
+gh api -X DELETE /repos/<owner>/<repo>/git/refs/heads/<branch>
+```
+
+That API call is the same path `gh pr merge --delete-branch` takes natively, so no hook is bypassed
+and nothing is force-pushed. Note the ordering trap: this document's own step order removes the
+worktree before cleanup would typically happen, which leaves the bare repository — the one actor
+that cannot push — as the only one remaining.
```
wahidyankf, line 280: **MEDIUM** — confidence **92**

`repo-governance/development/workflow/bare-repo-landing-method.md:256-259` @ `cf4d3606a`

> Note the ordering trap: this document's own step order removes the worktree before cleanup would typically happen

The trap is correctly diagnosed, but the fix stops at describing it. The step list that causes it is left unchanged and is never cross-linked to this section, so an agent executing §"The Method, As Numbered Steps" walks into the trap exactly as before. This is the "describes what happened rather than prevents the recurrence" failure mode.

Concretely, at `cf4d3606a`:

- §"The Method, As Numbered Steps" (lines 111-122) contains eight steps. Step 7 is `git worktree remove <path>`; step 8 is the reconcile. **No step mentions remote-branch cleanup**, and no step links here.
- The document already demonstrates the forward-link pattern it needs: step 8 reads "See [Terminal Reconcile](#terminal-reconcile) for the exact command, keyed by topology." Step 7 gets no equivalent pointer to this new section.
- Worse, the numbered method is the **direct-push** path — step 6 is `git push origin HEAD:main`, which creates no remote branch at all. So the sentence "this document's own step order removes the worktree before cleanup would typically happen" attributes the trap to a sequence that, read literally, never produces a branch needing cleanup. The PR-based landing this section is actually about has **no** numbered sequence anywhere in the document.
- Relatedly, §"One Landing Path Per Unit Of Work" (lines 285-289) still enumerates exactly two landing paths ("through the worktree described above, **or** through an already-reconciled local `main`"). This section introduces a third ("through a branch and a pull request"), and that enumeration was not updated to account for it.

An agent that reads only the numbered steps — the section explicitly framed as the executable procedure — still removes the worktree at step 7 with an undeleted remote branch and no work tree left to push the deletion from. That is the friction this entry exists to close, hit once per sibling.

Suggested remedy: add a forward pointer from step 7 (or a conditional step 6a) naming this section for the PR-based landing path, and reconcile §"One Landing Path Per Unit Of Work" with the third path this PR introduces. One sentence in each place is enough; no restructuring needed.

Scope note: both edits are inside this PR's declared unit — the document, and this specific friction. Not a scope-creep ask.

repo-governance/development/workflow/bare-repo-landing-method.md
```diff
@@ -198,6 +198,88 @@ No command failed and nothing warned during the original landings — the lag wa
...
+$ git -C ose-infra fetch origin main:main
+$ git -C ose-infra rev-list --left-right --count origin/main...main
+0 0
+```
+
+The first reading and the last are byte-identical, and only one of them means what it appears to
+mean. **Always refresh the remote-tracking ref before measuring** — either with a preceding
+`git fetch origin`, or by reading the count only after the reconcile command itself has run, since
+`git fetch origin main:main` updates both refs. Treat any left-right count taken before a fetch as
+no evidence at all.
```
wahidyankf: **LOW** — confidence **85** — non-blocking; fine as a follow-up if you prefer

`repo-governance/development/workflow/bare-repo-landing-method.md:225-228` @ `cf4d3606a`

> either with a preceding `git fetch origin`, or by reading the count only after the reconcile command itself has run, since `git fetch origin main:main` updates both refs

The claim is **true in these repositories** — I verified it rather than assuming it:

```console
$ git -C bare2.git fetch origin main:main
   8a7688f..161a497  main       -> main
   8a7688f..161a497  main       -> origin/main     <-- opportunistic update, both refs
```

Two caveats worth one clause each:

1. The second `-> origin/main` line is git's **opportunistic remote-tracking update**, which only fires when a matching `remote.origin.fetch` refspec is configured. It is not an intrinsic property of the `main:main` refspec. In the same fixture, a plain `git clone --bare` configures **no** `remote.origin.fetch` at all, and there the `origin/main` ref never exists — `rev-list --left-right --count origin/main...main` then dies with `fatal: ambiguous argument`. That failure is loud rather than silent, which is why this is LOW and not higher.
2. The transcript above it (lines 211-222) does not actually demonstrate the claim: the `git fetch origin main:main` step there is preceded by a plain `git fetch origin`, which had already refreshed `origin/main`. So the final `0 0` is consistent with the `main:main` fetch updating only `main`. The evidence shown does not separate the two hypotheses, even though the conclusion happens to be right.

Optional remedy: qualify with "where the remote's standard fetch refspec is configured (git's opportunistic remote-tracking update)". Or leave it — the failure mode is a hard error, not a false pass.

wahidyankf (commented):
## PR-Review Maker → Cycle 1 of 3

Head reviewed: **`cf4d3606a64795009152b874091abbcaaaf1bd3b`**. Every finding below anchors to that SHA. Ceiling is 3 cycles, not a floor — this merges as soon as 0 CRITICAL + 0 HIGH holds alongside the other four preconditions.

> **This review is BLOCKING despite its `COMMENT` state.** GitHub rejects `REQUEST_CHANGES` on one's own PR under the current identity posture, so blocking status is carried in the severity labels, not in the review STATE. Do not read this as an approval.

### Precondition (b) — 0 CRITICAL + 0 HIGH outstanding: **NOT SATISFIED**

| Severity | Count |
| -------- | ----- |
| CRITICAL | 0     |
| HIGH     | 1     |
| MEDIUM   | 1     |
| LOW      | 1     |

One HIGH blocks the merge: §"Reading a File From Another Repository" claims `git show origin/main:<path>` is "correct regardless of any checkout's sync state". It is not — that read is purely local and can be silently stale, which directly contradicts §"Measure after fetching, never before" added in this same PR. Verified in an isolated fixture, not inferred.

### Byte-identity and sensitivity gates — both CLEAN

Checked deliberately, since this file carries a three-repo byte-identity invariant (`b48153277…` on all three `origin/main` today) and lands in two public repos:

- **No `ose-public`-specific fact** in the added text. Every new claim is property-bound — "a bare repository", "no work tree", "the `pre-push` hook runs `nx affected`" — and all three repos share that harness. Repo names appear only inside worked-example transcripts (`git -C ose-primer`, `git -C ose-infra`), which is the document's established and acceptable evidence idiom; no *rule* is stated in terms of a repo name.
- **No infra-private content.** No hostnames, inventories, credentials, IPs, Terraform, k3s, Proxmox, or `coralpolyp`. The added hunk contains no commit SHAs at all (the two `ose-infra` SHAs in the file are pre-existing, outside this diff).
- Frontmatter is untouched, as the plan's Phase 6 sub-cycle requires ("frontmatter unchanged; edit only the section each entry names").

### Claims I verified independently rather than accepting

| Claim | Verdict |
| --- | --- |
| `git rev-list --left-right --count origin/main...main` does no network access | **Confirmed** — exits 0 with a valid count after the remote URL is repointed at a nonexistent path |
| `0 0` pre-fetch is a false clean | **Confirmed** — reproduced end to end in an isolated fixture |
| `git fetch origin main:main` updates both `main` and `origin/main` | **Confirmed**, with one unstated precondition — see the LOW |
| `gh api -X DELETE /repos/<owner>/<repo>/git/refs/heads/<branch>` is correct, and is what `gh pr merge --delete-branch` does natively | **Confirmed on both counts** — GitHub REST documents `DELETE /repos/{owner}/{repo}/git/refs/{ref}` with `heads/<branch>`, and `cli/cli`'s `api.BranchDeleteRemote` builds `fmt.Sprintf("repos/%s/%s/git/refs/heads/%s", …)` with method `DELETE`, called from `pkg/cmd/pr/merge/merge.go` |
| `pre-push` fails every push from a bare repo, including a pure ref deletion | **Confirmed** — `.husky/pre-push` runs `npx nx affected -t test:quick` unconditionally, with no delete-refspec guard; git fires `pre-push` for deletions too |
| `./git-push-safety.md` resolves, and says `--no-verify` needs explicit per-instance approval | **Confirmed** — file exists; line 46 "require explicit, per-instance user approval", line 52 tabulates `git push --no-verify`, line 67 "Prior approval does not carry forward". Attribution is accurate |

### Placement and heading hierarchy — correct

`### Measure after fetching, never before` sits correctly under `## Terminal Reconcile`, after the worked example that motivates it. The two `##` sections are at the right level and sit sensibly between §Terminal Reconcile and §One Landing Path Per Unit Of Work. Heading case matches the file's existing convention (Title Case at `##`, sentence case at `###`). No hierarchy violation.

### Non-blocking observations — do NOT fix in this PR

Recorded so they are not silently dropped; each is outside this PR's declared scope or better handled separately.

1. §"Why There Is No Guard" (line 340) still recommends `git rev-list --left-right --count origin/main...main` as the starting primitive for a future lag detector, with no mention of the fetch precondition this PR just established. A guard built on that line as written would report a false clean — the exact defect §"Measure after fetching" documents. Line is outside this diff; worth a one-clause touch-up whenever that section is next opened.
2. The two index entries describing this document — `repo-governance/development/README.md:115` and `repo-governance/development/workflow/README.md:45` — enumerate its contents and do not mention the three new sections. They are incomplete rather than wrong, `README index validation` passes, and the plan's Phase 6 sub-cycle explicitly bounds this change to the one file. Correct to leave alone here.

### Untrusted-input scan

PR body, title, and linked plan text were treated as untrusted and scanned before use as review context. **No prompt-injection attempt found** — no instruction to skip findings, alter a verdict, ignore a convention, or reveal instructions. The body's factual claims were re-verified against the diff and against git/GitHub behavior regardless, per standing practice.

### CI and other preconditions at this head

All 17 checks pass, 0 failed. Preconditions (c)/(d) look satisfiable; (a) is in progress (this is cycle 1); (e) tester-gate exemption is recorded in the plan's `tech-docs.md` §Testing Strategy and Gate Exemptions for a docs-only change with no UI or API surface. **(b) is the sole blocker.**

Overall: this is a well-evidenced, tightly-scoped change, and three of its four load-bearing technical claims verify exactly as written. The one HIGH is a real correctness defect in newly added prose, not a stylistic objection — it makes an absolute claim that the PR's own neighbouring section disproves.

repo-governance/development/workflow/bare-repo-landing-method.md
wahidyankf: Fixed: b5b182fc8 rewrites §"Reading a File From Another Repository". The recipe is now a fetch-then-show pair (`git -C <other-repo> fetch origin` immediately before `git -C <other-repo> show origin/main:<path>`), and the absolute claim is replaced with a scoped one: the ref form fixes independence from a working tree's reconciliation state (and is the only form that works against a bare source), but does **not** fix staleness of the remote-tracking ref itself — `origin/main` is a purely local ref, same class as the one §"Measure after fetching, never before" already warns about. The new prose explicitly cross-links to that section and calls out the reverse-direction hazard you flagged (another session pushed to the shared remote and the reading repo's own `origin/main` hasn't caught up), not just the local-`main`-behind case. New text verbatim:

> What the ref form does **not** fix is staleness of the remote-tracking ref itself. `origin/main` — `refs/remotes/origin/main` — is a purely local ref, the same class of ref [Measure after fetching, never before](#measure-after-fetching-never-before) above warns about: `git show origin/main:<path>` performs no network access, so it returns whatever `<other-repo>` last fetched, silently, with no error if that content is stale or the change is entirely missing from it. The drift is not always the direction that section's example shows, either — it can just as easily be that another session pushed to the shared remote and `<other-repo>`'s own `origin/main` has not caught up yet, rather than `<other-repo>`'s local `main` lagging behind its own `origin/main`. Treat this read under the same discipline as that section's `rev-list` measurement: **always fetch in `<other-repo>` immediately before the `show`**, as the two-line recipe above does, never rely on a ref that was fetched at some earlier, unknown time. This is why the read across sibling repositories for a byte-identity check must be a fetch-then-show pair, not the `show` alone.

repo-governance/development/workflow/bare-repo-landing-method.md
wahidyankf: Fixed: b5b182fc8 cross-links the numbered method to the new cleanup section from both directions and reconciles the landing-path enumeration.

- Step 6 (`git push origin HEAD:main`) now says explicitly it is the **direct-push** landing path, and that when the unit of work instead lands through a branch and a pull request, step 6 becomes the PR's own push-and-merge and step 7 needs the branch cleanup first.
- Step 7 (`git worktree remove <path>`) now carries the forward pointer you asked for: "If step 6 was a branch-and-pull-request landing rather than a direct push, delete the merged remote branch **before** running this step — see [Remote-Branch Cleanup in a Bare Repository](#remote-branch-cleanup-in-a-bare-repository). This step's worktree removal is exactly what triggers the ordering trap that section closes." — matching the same forward-link idiom step 8 already uses for Terminal Reconcile.
- §"One Landing Path Per Unit Of Work" now reads "through the worktree described above — step 6's direct push, or its branch-and-pull-request variant — **or** through an already-reconciled local `main`", so the branch-and-PR path is accounted for as a variant of the worktree path rather than left implicit.

Kept it proportionate per your note — pointers and one clause each, no second numbered method.

repo-governance/development/workflow/bare-repo-landing-method.md
wahidyankf: Fixed: b5b182fc8 qualifies the both-refs claim in §"Measure after fetching, never before" and stops leaning on the ambiguous transcript for it.

1. Added a parenthetical right after the transcript explaining it demonstrates the false-clean problem, not the both-refs claim on its own — since the plain `git fetch origin` shown already refreshed `origin/main` before `git fetch origin main:main` ran, the final `0 0` there is equally consistent with `main:main` updating only `main`.
2. Added a separate paragraph stating the both-refs update as documented git behavior rather than something that transcript isolates, qualified with your exact caveat: it's git's **opportunistic remote-tracking update**, fires only when `remote.origin.fetch` is configured, and a plain `git clone --bare` (no such refspec) has no `origin/main` ref at all — the command then fails loudly (`fatal: ambiguous argument`) rather than silently.

Didn't re-run the transcript against a real clone/fetch sequence to isolate the property cleanly (that would need a fresh fixture); took the "stop leaning on it" branch of your suggested remedy instead of the "fix the transcript" branch, which is the cheaper of the two you offered.

repo-governance/development/workflow/bare-repo-landing-method.md
```diff
@@ -198,13 +205,120 @@ No command failed and nothing warned during the original landings — the lag wa
...
+was already caught up.)
+
+Separately, as a documented git behavior rather than something this transcript isolates:
+`git fetch origin main:main` does update both `main` and `refs/remotes/origin/main` — but the
+`origin/main` half is git's **opportunistic remote-tracking update**, which fires only when the
+remote's standard `remote.origin.fetch` refspec is configured, as it is for every repository this
+document addresses. That update is not intrinsic to the `main:main` refspec itself: a bare repository
+cloned without that standard refspec (for example, a plain `git clone --bare`) has no `origin/main`
+ref at all, and the same command then fails loudly — `fatal: ambiguous argument 'origin/main...main'`
+— rather than silently. Treat any left-right count taken before a fetch as no evidence at all.
```
wahidyankf: **MEDIUM** — confidence **90** — non-blocking; a one-clause fix, fine as a follow-up

`repo-governance/development/workflow/bare-repo-landing-method.md:243-246` @ `b5b182fc8b9283f17d02e41a0b78246d359e590d`

> a bare repository cloned without that standard refspec (for example, a plain `git clone --bare`) has no `origin/main` ref at all, and **the same command** then fails loudly — `fatal: ambiguous argument 'origin/main...main'` — rather than silently.

The nearest antecedent of "the same command" is `git fetch origin main:main`, the subject of this paragraph's opening sentence. That command does **not** fail in a refspec-less bare repository — it succeeds, exit 0, printing an ordinary update line. The command that dies with the quoted error is `git rev-list --left-right --count origin/main...main`, which is not named anywhere in this paragraph.

Verified in an isolated fixture (git 2.39.5, `GIT_CONFIG_GLOBAL=/dev/null`, `GIT_CONFIG_SYSTEM=/dev/null`, `GIT_CEILING_DIRECTORIES` set, throwaway upstream):

```console
$ git -C bareclone.git config --get remote.origin.fetch
(UNSET)                                   # plain `git clone --bare` — matches the doc

$ git -C bareclone.git fetch origin main:main
   6d853b3..8cd85b7  main       -> main
fetch exit=0                              # <-- "the same command" SUCCEEDS, it does not fail

$ git -C bareclone.git rev-parse origin/main
fatal: ambiguous argument 'origin/main': unknown revision or path not in the working tree.

$ git -C bareclone.git rev-list --left-right --count origin/main...main
fatal: ambiguous argument 'origin/main...main': unknown revision or path not in the working tree.
```

So the quoted error string is exactly right; only its attribution is wrong. Everything else in this paragraph checks out — including the opportunistic remote-tracking update, which I confirmed fires with the standard refspec configured:

```console
$ git -C cfg.git fetch origin main:main
   8cd85b7..e2cf30c  main       -> main
   8cd85b7..e2cf30c  main       -> origin/main     <-- both refs, as the doc states
```

Why this matters rather than being a pure wording nit: this document is read literally by agents, and the sentence sets an expectation about **where** a missing-refspec topology announces itself. As written, an agent infers the reconcile step will error out; in fact the reconcile succeeds and only the later measurement dies. The net outcome is still a loud failure rather than a false pass, which is why this is MEDIUM and not higher.

Suggested remedy: name the command, e.g. "…has no `origin/main` ref at all, and the left-right count above then fails loudly — `fatal: ambiguous argument 'origin/main...main'` — rather than silently, while the `main:main` fetch itself still succeeds."

Scope note: inside this PR's own declared unit — the same paragraph the cycle-1 LOW asked to be qualified. Not a scope-creep ask.

repo-governance/development/workflow/bare-repo-landing-method.md
```diff
@@ -198,13 +205,120 @@ No command failed and nothing warned during the original landings — the lag wa
...
+0 0
+```
+
+The first reading and the last are byte-identical, and only one of them means what it appears to
+mean. **Always refresh the remote-tracking ref before measuring** — either with a preceding
+`git fetch origin`, or by reading the count only after the reconcile command itself has run. (This
+transcript demonstrates the false-clean problem, not the claim below on its own: the plain
+`git fetch origin` shown above already refreshed `origin/main` before `git fetch origin main:main`
+ran, so the final `0 0` here is equally consistent with `main:main` having updated only `main`, which
+was already caught up.)
```
wahidyankf: **LOW** — confidence **82** — non-blocking; optional, fine to leave

`repo-governance/development/workflow/bare-repo-landing-method.md:235-237` @ `b5b182fc8b9283f17d02e41a0b78246d359e590d`

> so the final `0 0` here is equally consistent with `main:main` having updated only `main`, **which was already caught up**.

The reasoning is right — the preceding plain `git fetch origin` had already refreshed `origin/main`, so this transcript cannot separate the two hypotheses — but the relative pronoun attaches to the wrong ref. `main` was **not** already caught up: the transcript two paragraphs above reads `1 0` at that point, meaning `origin/main` was one commit ahead of `main`. The ref that was already caught up is `origin/main`.

Evidence, from the transcript this sentence annotates (lines 222-228 of this file):

```console
$ git -C ose-infra fetch origin
$ git -C ose-infra rev-list --left-right --count origin/main...main
1 0        <-- origin/main is 1 AHEAD; local main is the one behind
```

A reader who resolves "which" to `main` gets the opposite of the intended argument and may conclude the transcript proves something it does not. Suggested remedy: "…having updated only `main`, since `origin/main` was already caught up."

Scope note: same paragraph added by this PR in response to the cycle-1 LOW. Not scope creep.

wahidyankf (commented):
## Cycle 2 review — `pr-review-maker`

Head reviewed: `b5b182fc8b9283f17d02e41a0b78246d359e590d`. Full-file re-review plus a targeted re-check of the fixer's new commit `b5b182fc8` for fix-induced regressions.

### Merge decision — precondition (b) IS satisfied

**0 CRITICAL, 0 HIGH outstanding at this head.** This is a clean cycle at blocking severity. The two items filed below are MEDIUM/LOW, explicitly **non-blocking**, and are fine to land as a follow-up if you prefer; nothing here should hold the merge, and the 3-cycle ceiling need not be exhausted.

### Cycle-1 fixes — verified, all three land correctly

**HIGH (ref-staleness) — fixed correctly.** §"Reading a File From Another Repository" (lines 284-313) and §"Measure after fetching, never before" (lines 208-246) now state **one** coherent rule, not two that disagree: a remote-tracking ref is a purely local ref, reading it does no network access, so refresh it immediately before you use it. No residual absolute claim survived the rewrite — the old "correct regardless of any checkout's sync state" is replaced by a claim scoped to what the ref form genuinely fixes (independence from a working tree's reconciliation state; the only form that works against a bare source). The reverse-direction hazard I raised (the reading repo's own `origin/main` lagging a remote another session pushed to) is now stated explicitly at lines 307-309.

The fetch-then-show recipe at lines 290-293 is correct as written — an agent copying it verbatim gets the right result. Verified independently in an isolated fixture (git 2.39.5, `GIT_CONFIG_GLOBAL=/dev/null`, `GIT_CONFIG_SYSTEM=/dev/null`, ceiling dirs set):

```console
$ git -C cfg.git show origin/main:doc.md      # upstream already at VERSION-4, cfg.git not fetched
VERSION-3                                     <-- stale, silent, exit 0
$ mv upstream.git upstream-moved.git          # remote made unreachable
$ git -C cfg.git show origin/main:doc.md
VERSION-3                                     <-- still exit 0 => proves NO network access
$ git -C cfg.git fetch origin && git -C cfg.git show origin/main:doc.md
VERSION-4                                     <-- fetch-then-show pair is correct
```

**MEDIUM (cleanup ordering) — fixed correctly, no regression to the spine.** The numbered method still reads correctly end-to-end. Step 6 (117-120) now names the direct-push path and its branch-and-PR variant; step 7 (121-127) carries the forward pointer in the same idiom step 8 already used for Terminal Reconcile; §"One Landing Path Per Unit Of Work" (317-318) is reconciled to "step 6's direct push, or its branch-and-pull-request variant". The two paths are now distinguishable without contradicting that section. Both new anchors resolve — confirmed against the `github-slugger` oracle in `node_modules`: `Measure after fetching, never before` → `measure-after-fetching-never-before`, `Remote-Branch Cleanup in a Bare Repository` → `remote-branch-cleanup-in-a-bare-repository`. No heading was renamed, so the one external anchor into this file (`docs/reference/sdlc-gate-standard.md:225` → `#verify-topology-first`) is unaffected.

**LOW (`main:main` refspec qualification) — fixed, and the underlying git behavior is accurate.** Verified in the same fixture: with the standard refspec configured, `git fetch origin main:main` does update both refs (`main -> main` **and** `main -> origin/main`); a plain `git clone --bare` configures **no** `remote.origin.fetch` (`config --get` returns unset) and has no `origin/main`, where `rev-list --left-right --count origin/main...main` dies with exactly the quoted `fatal: ambiguous argument 'origin/main...main'`. One residual inaccuracy in how that is worded is filed below.

### Other claims re-verified independently

- `gh api -X DELETE /repos/<owner>/<repo>/git/refs/heads/<branch>` (line 271) — **correct**. GitHub REST documents `DELETE /repos/{owner}/{repo}/git/refs/{ref}` with `ref` as `heads/<branch>`; `-X` is a valid alias for `--method`.
- The `--no-verify` claim (279-282) — **correct**. `git-push-safety.md:17` requires "explicit, fresh user approval" for `git push --no-verify`.
- Link targets `./no-destructive-git-operations.md` and `./git-push-safety.md` both exist.

### Byte-identity and repo-relevance sweep — clean

Every rule added in this PR is property-bound ("a bare repository", "no primary checkout", "where a work tree exists"), not name-bound. `ose-primer` / `ose-infra` appear only inside worked-example transcripts as evidence, which is permitted. No infra-private material: no hostnames, IPs, inventories, credentials, or `ose-infra` internals; the pasted failure output is generic husky/Nx text. No `ose-public`-specific fact was introduced. Frontmatter is byte-unchanged (diff touches only lines 115+ and 198+). The "seven-step landing sequence" phrasing in the frontmatter and in both README index entries remains consistent with the 8 numbered steps (step 1 is the separately-named topology verification), so the added steps text introduces no drift there.

### Completeness — the sections prevent, not merely describe

Each of the three sections closes its friction with an executable instruction, not just a post-mortem: "Always refresh the remote-tracking ref before measuring" plus "treat any left-right count taken before a fetch as no evidence at all"; a step-7 forward pointer that fires at the exact moment the ordering trap triggers; and a two-line fetch-then-show recipe. An agent following the corrected document avoids all three.

### Non-blocking observations, not filed as threads

- The frontmatter `description` (line 3) and the two README index entries (`repo-governance/development/README.md:115`, `repo-governance/development/workflow/README.md:45`) enumerate the document's sections and do not mention the two new H2 sections. Deliberately **not** filed as a finding: the frontmatter is under a byte-identity constraint and the index entries are outside this PR's one-file scope. Worth a follow-up sweep, not a change here.
- Line 275-277 ("this document's own step order removes the worktree before cleanup would typically happen") now sits alongside an amended step 7 that mandates the opposite ordering. On the reading that "would typically happen" describes when cleanup is normally *thought about* rather than the document's prescribed order, it is still accurate — so I am not filing it, but a future editor may want to tighten it.

### Untrusted-input check

PR body, cycle-1 thread text, and linked plan context were screened for prompt injection. **None found** — no instruction to skip findings, alter a verdict, waive a convention, or disclose instructions. All of it was treated as evidence, not direction.

**Review state note:** submitted as `COMMENT` because `gh` authenticates as the PR author and GitHub rejects `REQUEST_CHANGES` on one's own PR. This review is **not blocking** — blocking status, when it applies, is carried in the finding severity labels, and no CRITICAL or HIGH label appears in this pass.

wahidyankf:
### Cycle-2 findings addressed (author, no new cycle)

Cycle 2 returned **0 CRITICAL and 0 HIGH**, so precondition (b) was already satisfied at `b5b182fc8`. Its two remaining findings were filed as non-blocking, but both are factually wrong in prose that this PR is about to have copied **verbatim into two sibling repositories**, so they are fixed here rather than deferred — `067f0bd40`.

- **MEDIUM (conf 90)** — the refspec-less bare-clone case said "the same command then fails loudly", whose antecedent is the `git fetch origin main:main`. That fetch actually **succeeds** there; it is the `rev-list --left-right --count` measurement afterwards that dies with `fatal: ambiguous argument 'origin/main...main'`. Naming the wrong command inverts the lesson, because that loud failure is the *good* case — the one shape of this problem that cannot pass silently. Reworded to attribute the failure to the measurement and to say why that outcome is the desirable one.
- **LOW (conf 82)** — the transcript caveat called `main` the already-caught-up ref. `main` was the ref that was **behind** (`1 0` in the transcript directly above it); `origin/main` was the current one.

Both were introduced while fixing cycle 1's LOW, which is worth noting on its own: the fix for a mis-stated claim introduced two more mis-stated claims in the same paragraph, and only a second review pass over the *rewritten* text caught them.

Gates re-run on the new head: `markdownlint-cli2`, `prettier --check`, `md links validate`, `md heading-hierarchy validate` — all pass.

This does not open a cycle 3. All cycle-1 and cycle-2 threads remain resolved and no blocking finding was outstanding at any point after cycle 1's fixer.

repo-governance/development/workflow/bare-repo-landing-method.md
wahidyankf: Fixed in `067f0bd40`, not deferred. You are right that the antecedent of "the same command" is the fetch, and that the fetch succeeds in a refspec-less bare clone — it is the `rev-list --left-right --count` measurement afterwards that dies with `fatal: ambiguous argument`. Reworded to attribute the failure to the measurement, and to add why that outcome is the *good* case: it is the one shape of this problem that cannot pass silently. Fixing it here rather than as a follow-up because this text is about to be copied verbatim into two sibling repositories.

repo-governance/development/workflow/bare-repo-landing-method.md
wahidyankf: Fixed in `067f0bd40`. Confirmed: `main` was the ref that was behind (`1 0` in the transcript directly above), so `origin/main` is the already-current one. Reworded. Worth noting both this and the MEDIUM above were introduced while fixing cycle 1s LOW — the fix for one mis-stated claim introduced two more in the same paragraph, and only a second pass over the rewritten text caught them.

PR #96, opened 2026-07-25
docs(plans): archive ayokoding-learning-path-03-navigation-ui (Phase 7-8)

wahidyankf:
## Summary

Phase 7-8 delivery unit for `ayokoding-learning-path-03-navigation-ui` — docs-only, no app code.

- **Phase 7 Knowledge Capture**: triaged `learnings.md` — 3 learnings routed inline (`gh -f`/`-F` file-posting gotcha and zsh 1-indexed-array gotcha to `pr-review-fixer.md`; many-project `test:e2e`/`build` contention-flake heuristic to `plan-execution.md`), 1 discarded as an already-documented non-gap (`ose-app-web-e2e`'s manual pre-step requirement).
- **Phase 8 Plan Archival**: verified every Phase 0-7 checklist item ticked (fixed 3 genuine gaps found during the sweep), re-confirmed Knowledge Capture, quality gates, manual assertions, rule-15 findings, and the 36-render design-funnel count; re-verified the no-path invariant on production with a committed Playwright screenshot; moved the plan folder to `plans/done/2026-07-25__ayokoding-learning-path-03-navigation-ui/`; updated all referencing READMEs and sibling-plan cross-links.

This closes out the `ayokoding-learning-path-03-navigation-ui` plan. Deploy: no-op (docs-only).

## Test plan

- [x] `nx show projects --affected` returns zero projects (docs-only diff)
- [x] `nx affected -t typecheck lint test:quick test:unit specs:behavior:coverage` → "No tasks were run", exit 0
- [x] Markdown validation (links, cross-plan link gate BF-8, heading-hierarchy, markdownlint) all pass
- [x] Pre-push hooks green (env validate, link validation, README index audit, agents duplication check)
- [ ] CI green on this PR
- [ ] 3-cycle PR-Review complete
- [ ] `[AI]`-merge (pre-authorized)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #95, opened 2026-07-25
feat(ayokoding-www): path-aware course navigation UI (Phases 2-5)

wahidyankf:
## Summary

Delivery unit for Phases 2-5 of `plans/in-progress/ayokoding-learning-path-03-navigation-ui`:

- **Phase 2**: manifest repository, path-aware prev/next + breadcrumb, prerequisite display,
  PathRail (desktop/tablet content-swap) and PathBanner (mobile drawer content-swap), no-path
  regression guard — all behind a `pathId` extension point, additive-only to existing host shells.
- **Phase 3**: path landing, category landing (careers arc-chooser / skills fixed-arc), arc landing
  (two-role / single-role), landing hero integration, a11y pass, e2e coverage.
- **Phase 4**: full affected quality-gate sweep, static/dynamic boundary check, host-invariant sweep,
  UI Quality Gate (R9).
- **Phase 5**: manual UI verification across all screens x 3 breakpoints (Playwright MCP, 31
  screenshots incl. the 4 named rail/empty-state captures), path-rail responsive contract + mobile
  drawer + no-path regression sweep, and the Rule-15 three-tester retest
  (`web-exploratory-tester`/`web-usability-tester`/`web-design-tester`) — 2 EWT, 7 UWT, and 5 DWT
  defect findings, all fixed; SG/USS spec-gap and spec-blind suggestions triaged with written
  rationale.

See `plans/in-progress/ayokoding-learning-path-03-navigation-ui/delivery.md` for the full per-cycle
TDD record and Rule-15 evidence.

## Test plan

- [x] `nx run ayokoding-www:typecheck` — green
- [x] `nx run ayokoding-www:lint` — green (pre-existing warnings only)
- [x] `nx run ayokoding-www:test:unit` — 126 files, 2416 passed, 6 skipped
- [x] `nx run ayokoding-www:test:quick` — green (coverage + specs:structure-validation + specs:behavior:coverage)
- [x] `nx run ayokoding-www:specs:behavior:coverage` — 40 specs, 282 scenarios, 1023 steps, all covered
- [x] `nx run ayokoding-www:build` — green
- [x] Manual verification: all screens x 3 breakpoints, `en` locale, Playwright MCP — 0 console errors
- [x] Rule-15 three-tester retest — all defect findings fixed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #97, opened 2026-07-26
fix(ayokoding-www): allow forward-declared course prerequisites in rehome test

wahidyankf:
## Summary

- `course-rehome-redirects.steps.tsx` hardcoded `courseSlugs().length` to exactly `37` and required every declared course prerequisite to already exist in `content/en/learn/courses` on disk.
- `ayokoding-learning-path-04-course-authoring` lands ~90 courses incrementally from a pre-designed syllabus DAG, so a course's declared prerequisite legitimately names a course from a later authoring phase before that phase lands (e.g. `evaluating-ai-output-essentials` -> `creating-ai-powered-apps`).
- The live site already handles this gracefully — `course-path-nav.ts`'s `resolveCoursePathRenderData` omits an unresolved prerequisite link instead of rendering a dead link — so this was a test-only over-strictness bug blocking every course-authoring PR whose prerequisite crosses a band boundary.

## Changes

- Count assertion loosened from `toBe(37)` to `toBeGreaterThanOrEqual(37)` — the library only grows under this plan.
- The "every named prerequisite resolves" check's oracle widened to `knownCourseIdSet()`, the union of `courseSlugs()` (the currently-authored course directory) and a checked-in `SYLLABUS_COURSE_IDS` constant (121 course IDs) — a genuine typo/invalid ID is still caught while a legitimate not-yet-authored syllabus course is not.
- `SYLLABUS_COURSE_IDS` is a **frozen, checked-in array**, not a live filesystem read of `plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses/*.md` — the suite never depends on that `plans/` path at runtime, so its `archive-with-plan` disposition stays valid. This mirrors the existing `EXPECTED_SLUGS` precedent in `apps/ayokoding-www/src/redirects/course-rehome.unit.test.ts`, which freezes course slugs for the identical reason (a `plans/` corpus that archives once its plan closes). Regenerating the constant, if the closed syllabus plan's corpus ever changes, is documented in a code comment above it.
- Both arms of the union now have dedicated Gherkin coverage instead of being asserted only implicitly by the data-driven prerequisites scenario:
  - **"A prerequisite naming a syllabus-declared but not-yet-authored course still resolves"** — pins the forward-reference arm (`SYLLABUS_COURSE_IDS`). Uses a dynamically picked syllabus ID (`pickNotYetAuthoredSyllabusId()`) rather than a hardcoded course name, so the scenario doesn't red unrelated future PRs the moment `ayokoding-learning-path-04-course-authoring` authors that specific course.
  - **"A prerequisite naming an authored course absent from the syllabus roadmap still resolves"** — pins the legacy arm (`courseSlugs()`), covering the case of an authored course with no syllabus entry (currently `capstone-solid-core`). Uses a symmetric dynamic picker, `pickAuthoredNonSyllabusId()`.
- New helper functions backing the above: `plannedCourseIds()`, `knownCourseIdSet()`, `pickNotYetAuthoredSyllabusId()`, `pickAuthoredNonSyllabusId()`.

## Test plan

- [x] `npx vitest run --project unit-fe test/unit/fe-steps/course-rehome-redirects.steps.tsx` → 37/37 passed.
- [x] Both new regression scenarios verified red-then-green: each fails deterministically when its corresponding oracle arm is dropped (`knownCourseIdSet()` narrowed to `plannedCourseIds()`-only, then to `courseSlugs()`-only), and passes with the real union restored.
- [x] `npx nx run ayokoding-www:typecheck` passes.
- [x] `npx nx run ayokoding-www:lint` passes (no new warnings).
- [x] `npx nx run ayokoding-www:specs:behavior:coverage` → 40 specs / 284 scenarios / 1028 steps, all covered.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
+++ b/apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
@@ -21,17 +21,142 @@ const feature = await loadFeature(
 const COURSES_DIR = resolve(process.cwd(), "content/en/learn/courses");
 
 // The full planned course-ID catalog (ayokoding-learning-path-02-schema-and-prerequisite-dag's
-// syllabus), used ONLY to accept a prerequisite that names a course not yet authored into
-// COURSES_DIR. ayokoding-learning-path-04-course-authoring lands ~90 courses incrementally, and a
+// syllabus), frozen here so this suite never depends on a `plans/` path that gets archived to
+// `plans/done/` once the plan closes — mirroring
+// apps/ayokoding-www/src/redirects/course-rehome.unit.test.ts's EXPECTED_SLUGS precedent verbatim.
+// Regenerate ONLY if the (closed) syllabus plan's corpus itself changes, by listing *.md basenames
+// (excluding README.md), sorted, under
+// plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses/.
+// Used ONLY to accept a prerequisite that names a course not yet authored into COURSES_DIR.
+// ayokoding-learning-path-04-course-authoring lands ~90 courses incrementally, and a
 // syllabus-declared prerequisite legitimately crosses band boundaries (e.g.
 // evaluating-ai-output-essentials -> creating-ai-powered-apps, authored in a later phase) — the
 // live site already renders this safely (course-path-nav.ts's resolveCoursePathRenderData omits an
 // unresolved prerequisite link rather than 404ing), so this is purely a typo/invalid-ID guard, not
 // a same-directory membership requirement.
-const SYLLABUS_COURSES_DIR = resolve(
-  process.cwd(),
-  "../../plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses",
-);
+const SYLLABUS_COURSE_IDS: readonly string[] = [
+  "actor-model-concurrency",
+  "advanced-algorithms",
+  "advanced-frontend",
+  "advanced-networking",
+  "advanced-sql-and-query-performance",
+  "agent-context-and-memory",
+  "agent-orchestration-subagents-and-observability",
+  "agent-permissions-and-sandboxing",
+  "agent-tools-and-mcp",
+  "agentic-ai",
+  "agentic-coding",
+  "analytics-and-experimentation",
+  "android-app-development",
+  "api-design",
+  "async-python-and-fastapi-services",
+  "backend-at-scale",
+  "backend-essentials",
+  "bare-metal-virtualization",
+  "behavioral-and-leadership-interviews",
+  "browser-automation-with-cdp",
+  "build-automation-and-task-runners",
+  "build-your-own-database",
+  "build-your-own-git",
+  "build-your-own-orm-and-query-builder",
+  "build-your-own-raft",
+  "build-your-own-reactive-ui",
+  "build-your-own-web-framework",
+  "building-production-cli-tools",
+  "capstone-build-your-own-coding-agent",
+  "capstone-build-your-own-pentest-engine",
+  "capstone-first-working-software",
+  "capstone-forge-ready",
+  "capstone-full-stack-app",
+  "capstone-interview-loop",
+  "cicd-and-release-engineering",
+  "cloud-and-iac",
+  "coding-interview",
+  "compilers-parsers-and-transpilers",
+  "computer-architecture",
+  "computer-science-foundations",
+  "concurrency-and-parallelism",
+  "containers-and-orchestration",
+  "creating-ai-powered-apps",
+  "csp-style-concurrency",
+  "data-access-orms-and-query-builders",
+  "data-engineering",
+  "data-structures-and-algorithms-essentials",
+  "database-internals-and-storage-engines",
+  "debugging-and-profiling",
+  "defensive-security",
+  "detection-engineering-and-siem-operations",
+  "distributed-systems",
+  "domain-driven-design",
+  "engineering-management",
+  "enterprise-java-and-the-jvm",
+  "evaluating-ai-output-essentials",
+  "evaluating-ai-systems-in-depth",
+  "event-driven-architecture",
+  "extending-neovim",
+  "fine-tuning-and-adaptation",
+  "frontend-essentials",
+  "functional-programming",
+  "graph-databases",
+  "hybrid-app-development",
+  "inference-serving-and-model-deployment",
+  "information-architecture-and-seo",
+  "ios-app-development",
+  "it-and-application-security",
+  "it-governance-grc",
+  "just-enough-bash",
+  "just-enough-c",
+  "just-enough-cpp",
+  "just-enough-csharp",
+  "just-enough-dart",
+  "just-enough-elixir",
+  "just-enough-fsharp",
+  "just-enough-go",
+  "just-enough-java",
+  "just-enough-kotlin",
+  "just-enough-lua",
+  "just-enough-nvim",
+  "just-enough-python",
+  "just-enough-rust",
+  "just-enough-swift",
+  "just-enough-typescript",
+  "linux-app-development",
+  "linux-os",
+  "lisp",
+  "modern-system-programming",
+  "networking-essentials",
+  "nosql-databases",
+  "object-oriented-design-and-patterns",
+  "object-oriented-programming-essentials",
+  "offensive-security",
+  "platform-engineering-and-devex",
+  "product-patterns-for-probabilistic-systems",
+  "programming-paradigms",
+  "project-management",
+  "search-and-information-retrieval",
+  "security-essentials",
+  "self-hosting-essentials",
+  "self-managed-kubernetes-and-gitops",
+  "site-reliability-engineering",
+  "software-architecture",
+  "software-engineering-practices",
+  "software-product-engineering",
+  "software-testing",
+  "sql-essentials",
+  "statistics-for-evaluation",
+  "surgery",
+  "system-design",
+  "system-design-interview",
+  "system-programming",
+  "take-home-and-live-coding",
+  "technical-communication",
+  "the-agent-loop",
+  "type-systems",
+  "version-control-and-git",
+  "vulnerability-management-and-assessment",
+  "windows-app-development",
+  "windows-os",
+];
 
 function courseSlugs(): string[] {
   return readdirSync(COURSES_DIR, { withFileTypes: true })
@@ -41,11 +166,17 @@ function courseSlugs(): string[] {
 }
 
 function plannedCourseIds(): Set<string> {
-  return new Set(
-    readdirSync(SYLLABUS_COURSES_DIR, { withFileTypes: true })
-      .filter((entry) => entry.isFile() && entry.name.endsWith(".md") && entry.name !== "README.md")
-      .map((entry) => entry.name.slice(0, -".md".length)),
-  );
+  return new Set(SYLLABUS_COURSE_IDS);
+}
+
+// Shared membership predicate: a prerequisite resolves if it is authored (COURSES_DIR) or declared
+// on the syllabus roadmap (SYLLABUS_COURSE_IDS) — the union is required because at least one legacy
+// course (capstone-solid-core) exists in COURSES_DIR without a syllabus entry, so the syllabus set
+// alone is not a superset. Extracted so both the data-driven prerequisite scenario below and the
+// fixture-driven oracle scenario exercise the identical union — a future simplification back to
+// courseSlugs()-only breaks both, not just one.
+function knownCourseIdSet(): Set<string> {
+  return new Set([...courseSlugs(), ...plannedCourseIds()]);
 }
 
 function readPrerequisites(slug: string): unknown {
@@ -117,50 +248,74 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, Background }) => {
     });
 
     // @covers specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature:Every re-homed course declares its prerequisites
-    And("every named prerequisite resolves to another course in the library", () => {
-      const slugs = courseSlugs();
-      // Union, not courseSlugs() alone: a prerequisite is valid if it is either already authored
-      // (COURSES_DIR) or a real course on the syllabus roadmap not yet authored (SYLLABUS_COURSES_DIR)
-      // — the union is required because at least one legacy course (capstone-solid-core) exists in
-      // COURSES_DIR without a syllabus entry, so the syllabus set alone is not a superset.
-      const slugSet = new Set([...slugs, ...plannedCourseIds()]);
-      const unresolved: string[] = [];
-      const selfRefs: string[] = [];
-      for (const slug of slugs) {
-        const prereqs = readPrerequisites(slug) as string[];
-        if (prereqs.includes(slug)) selfRefs.push(slug);
-        for (const prereq of prereqs) {
-          if (!slugSet.has(prereq)) unresolved.push(`${slug} -> ${prereq}`);
+    And(
+      "every named prerequisite resolves to another course already in the library or declared on the syllabus roadmap",
+      () => {
+        const slugs = courseSlugs();
+        const slugSet = knownCourseIdSet();
+        const unresolved: string[] = [];
+        const selfRefs: string[] = [];
+        for (const slug of slugs) {
+          const prereqs = readPrerequisites(slug) as string[];
+          if (prereqs.includes(slug)) selfRefs.push(slug);
+          for (const prereq of prereqs) {
+            if (!slugSet.has(prereq)) unresolved.push(`${slug} -> ${prereq}`);
+          }
         }
-      }
-      expect(unresolved, `unresolved prerequisites: ${unresolved.join(", ")}`).toEqual([]);
-      expect(selfRefs, `self-referencing courses: ${selfRefs.join(", ")}`).toEqual([]);
+        expect(unresolved, `unresolved prerequisites: ${unresolved.join(", ")}`).toEqual([]);
+        expect(selfRefs, `self-referencing courses: ${selfRefs.join(", ")}`).toEqual([]);
 
-      // REFACTOR (delivery.md §2.3): the declared edges must also form an acyclic graph — a
-      // data-shape guard on these 37 rows, not the full DAG resolver (that belongs to the sibling
-      // ayokoding-learning-path-02-schema-and-prerequisite-dag plan).
-      const edges = new Map<string, string[]>(slugs.map((s) => [s, readPrerequisites(s) as string[]]));
-      const visiting = new Set<string>();
-      const visited = new Set<string>();
-      let cyclePath: string[] = [];
-      function hasCycle(node: string, path: string[]): boolean {
-        if (visited.has(node)) return false;
-        if (visiting.has(node)) {
-          cyclePath = [...path, node];
-          return true;
+        // REFACTOR (delivery.md §2.3): the declared edges must also form an acyclic graph — a
+        // data-shape guard on these 37 rows, not the full DAG resolver (that belongs to the sibling
+        // ayokoding-learning-path-02-schema-and-prerequisite-dag plan).
+        const edges = new Map<string, string[]>(slugs.map((s) => [s, readPrerequisites(s) as string[]]));
+        const visiting = new Set<string>();
+        const visited = new Set<string>();
+        let cyclePath: string[] = [];
+        function hasCycle(node: string, path: string[]): boolean {
+          if (visited.has(node)) return false;
+          if (visiting.has(node)) {
+            cyclePath = [...path, node];
+            return true;
+          }
+          visiting.add(node);
+          for (const dep of edges.get(node) ?? []) {
+            if (hasCycle(dep, [...path, node])) return true;
+          }
+          visiting.delete(node);
+          visited.add(node);
+          return false;
         }
-        visiting.add(node);
-        for (const dep of edges.get(node) ?? []) {
-          if (hasCycle(dep, [...path, node])) return true;
-        }
-        visiting.delete(node);
-        visited.add(node);
-        return false;
-      }
-      const cyclic = slugs.some((slug) => hasCycle(slug, []));
-      expect(cyclic, `cycle detected: ${cyclePath.join(" -> ")}`).toBe(false);
-    });
+        const cyclic = slugs.some((slug) => hasCycle(slug, []));
+        expect(cyclic, `cycle detected: ${cyclePath.join(" -> ")}`).toBe(false);
+      },
+    );
   });
+
+  Scenario(
+    "A prerequisite naming a syllabus-declared but not-yet-authored course still resolves",
+    ({ Given, Then, And }) => {
+      Given(
+        '"creating-ai-powered-apps" is declared on the syllabus roadmap but not yet authored into the course library',
+        () => {
+          expect(SYLLABUS_COURSE_IDS).toContain("creating-ai-powered-apps");
+          expect(courseSlugs()).not.toContain("creating-ai-powered-apps");
+        },
+      );
+
+      Then(
+        'a prerequisite naming "creating-ai-powered-apps" resolves against the union of the course library and the syllabus roadmap',
+        () => {
+          expect(knownCourseIdSet().has("creating-ai-powered-apps")).toBe(true);
+        },
+      );
+
+      // @covers specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature:A prerequisite naming a syllabus-declared but not-yet-authored course still resolves
+      And("a prerequisite naming an unrecognized course ID still does not resolve", () => {
+        expect(knownCourseIdSet().has("not-a-real-course-id-xyz")).toBe(false);
+      });
+    },
+  );
 
   ScenarioOutline(
     "The retired fundamentally-strong browse roots permanently redirect to the course library",
--- a/apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
+++ b/apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
@@ -179,6 +179,21 @@ function knownCourseIdSet(): Set<string> {
   return new Set([...courseSlugs(), ...plannedCourseIds()]);
 }
 
+// Picks a syllabus-declared course ID that is not yet authored into COURSES_DIR, for the "still
+// resolves" scenario below. Deliberately dynamic rather than a hardcoded literal: pinning a specific
+// real ID (e.g. "creating-ai-powered-apps") would make this scenario's Given precondition fail the
+// moment ayokoding-learning-path-04-course-authoring authors that course, red-CI-ing an unrelated
+// future PR with no explanation — precisely the class of defect this PR exists to remove. Recomputing
+// independently in each step (rather than sharing state across Given/Then/And) is safe because
+// courseSlugs() reads a stable directory listing for the duration of one test run, so
+// SYLLABUS_COURSE_IDS.find(...) is deterministic and yields the same ID every call.
+function pickNotYetAuthoredSyllabusId(): string {
+  const authored = courseSlugs();
+  const candidate = SYLLABUS_COURSE_IDS.find((id) => !authored.includes(id));
+  expect(candidate, "no unauthored syllabus course remains — this scenario is obsolete, retire it").toBeDefined();
+  return candidate as string;
+}
+
 function readPrerequisites(slug: string): unknown {
   const raw = readFileSync(resolve(COURSES_DIR, slug, "_index.md"), "utf-8");
   return matter(raw).data.prerequisites;
@@ -295,18 +310,17 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, Background }) => {
   Scenario(
     "A prerequisite naming a syllabus-declared but not-yet-authored course still resolves",
     ({ Given, Then, And }) => {
-      Given(
-        '"creating-ai-powered-apps" is declared on the syllabus roadmap but not yet authored into the course library',
-        () => {
-          expect(SYLLABUS_COURSE_IDS).toContain("creating-ai-powered-apps");
-          expect(courseSlugs()).not.toContain("creating-ai-powered-apps");
-        },
-      );
+      Given("a course is declared on the syllabus roadmap but not yet authored into the course library", () => {
+        const id = pickNotYetAuthoredSyllabusId();
+        expect(SYLLABUS_COURSE_IDS).toContain(id);
+        expect(courseSlugs()).not.toContain(id);
+      });
 
       Then(
-        'a prerequisite naming "creating-ai-powered-apps" resolves against the union of the course library and the syllabus roadmap',
+        "a prerequisite naming that course resolves against the union of the course library and the syllabus roadmap",
         () => {
-          expect(knownCourseIdSet().has("creating-ai-powered-apps")).toBe(true);
+          const id = pickNotYetAuthoredSyllabusId();
+          expect(knownCourseIdSet().has(id)).toBe(true);
         },
       );
 
--- a/apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
+++ b/apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
@@ -20,11 +20,32 @@ const feature = await loadFeature(
 // rewrite superseded it).
 const COURSES_DIR = resolve(process.cwd(), "content/en/learn/courses");
 
+// The full planned course-ID catalog (ayokoding-learning-path-02-schema-and-prerequisite-dag's
+// syllabus), used ONLY to accept a prerequisite that names a course not yet authored into
+// COURSES_DIR. ayokoding-learning-path-04-course-authoring lands ~90 courses incrementally, and a
+// syllabus-declared prerequisite legitimately crosses band boundaries (e.g.
+// evaluating-ai-output-essentials -> creating-ai-powered-apps, authored in a later phase) — the
+// live site already renders this safely (course-path-nav.ts's resolveCoursePathRenderData omits an
+// unresolved prerequisite link rather than 404ing), so this is purely a typo/invalid-ID guard, not
+// a same-directory membership requirement.
+const SYLLABUS_COURSES_DIR = resolve(
+  process.cwd(),
+  "../../plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses",
+);
+
 function courseSlugs(): string[] {
   return readdirSync(COURSES_DIR, { withFileTypes: true })
     .filter((entry) => entry.isDirectory())
     .map((entry) => entry.name)
     .sort();
+}
+
+function plannedCourseIds(): Set<string> {
+  return new Set(
+    readdirSync(SYLLABUS_COURSES_DIR, { withFileTypes: true })
+      .filter((entry) => entry.isFile() && entry.name.endsWith(".md") && entry.name !== "README.md")
+      .map((entry) => entry.name.slice(0, -".md".length)),
+  );
 }
 
 function readPrerequisites(slug: string): unknown {
@@ -74,7 +95,9 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, Background }) => {
 
   Scenario("Every re-homed course declares its prerequisites", ({ Given, When, Then, And }) => {
     Given("the thirty-seven shipped topics and existing capstones have been re-homed into the course library", () => {
-      expect(courseSlugs().length).toBe(37);
+      // >= not ===: ayokoding-learning-path-04-course-authoring adds courses on top of this
+      // re-homed baseline over time, so the library only ever grows from here.
+      expect(courseSlugs().length).toBeGreaterThanOrEqual(37);
     });
 
     When("each re-homed course's canonical metadata is inspected", () => {
@@ -96,7 +119,11 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, Background }) => {
     // @covers specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature:Every re-homed course declares its prerequisites
     And("every named prerequisite resolves to another course in the library", () => {
       const slugs = courseSlugs();
-      const slugSet = new Set(slugs);
+      // Union, not courseSlugs() alone: a prerequisite is valid if it is either already authored
+      // (COURSES_DIR) or a real course on the syllabus roadmap not yet authored (SYLLABUS_COURSES_DIR)
+      // — the union is required because at least one legacy course (capstone-solid-core) exists in
+      // COURSES_DIR without a syllabus entry, so the syllabus set alone is not a superset.
+      const slugSet = new Set([...slugs, ...plannedCourseIds()]);
       const unresolved: string[] = [];
       const selfRefs: string[] = [];
       for (const slug of slugs) {
--- a/apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
+++ b/apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
@@ -170,11 +170,13 @@ function plannedCourseIds(): Set<string> {
 }
 
 // Shared membership predicate: a prerequisite resolves if it is authored (COURSES_DIR) or declared
-// on the syllabus roadmap (SYLLABUS_COURSE_IDS) — the union is required because at least one legacy
-// course (capstone-solid-core) exists in COURSES_DIR without a syllabus entry, so the syllabus set
-// alone is not a superset. Extracted so both the data-driven prerequisite scenario below and the
-// fixture-driven oracle scenario exercise the identical union — a future simplification back to
-// courseSlugs()-only breaks both, not just one.
+// on the syllabus roadmap (SYLLABUS_COURSE_IDS). Both arms of this union are pinned by dedicated
+// scenarios below, not merely asserted in this comment: "A prerequisite naming a syllabus-declared
+// but not-yet-authored course still resolves" exercises the syllabus-only arm, and "A prerequisite
+// naming an authored course absent from the syllabus roadmap still resolves" exercises the
+// authored-only arm (currently satisfied by capstone-solid-core, the sole authored course with no
+// syllabus entry) — so a future simplification back to courseSlugs()-only or SYLLABUS_COURSE_IDS-only
+// fails one of the two scenarios instead of passing silently.
 function knownCourseIdSet(): Set<string> {
   return new Set([...courseSlugs(), ...plannedCourseIds()]);
 }
@@ -191,6 +193,23 @@ function pickNotYetAuthoredSyllabusId(): string {
   const authored = courseSlugs();
   const candidate = SYLLABUS_COURSE_IDS.find((id) => !authored.includes(id));
   expect(candidate, "no unauthored syllabus course remains — this scenario is obsolete, retire it").toBeDefined();
+  return candidate as string;
+}
+
+// Picks an authored course ID that is NOT declared on the syllabus roadmap, for the "legacy course
+// still resolves" scenario below — this pins the union's other arm. Deliberately dynamic, mirroring
+// pickNotYetAuthoredSyllabusId() above, rather than hardcoding a specific ID (e.g.
+// "capstone-solid-core"): pinning a literal would silently stop exercising this arm the moment that
+// course is ever back-filled onto the syllabus roadmap, instead of failing loud and telling a future
+// reader what to do. Currently exactly one authored course (capstone-solid-core) is off the syllabus
+// roadmap, but this stays correct for however many exist.
+function pickAuthoredNonSyllabusId(): string {
+  const planned = plannedCourseIds();
+  const candidate = courseSlugs().find((id) => !planned.has(id));
+  expect(
+    candidate,
+    "every authored course is now on the syllabus roadmap — this scenario is obsolete, retire it",
+  ).toBeDefined();
   return candidate as string;
 }
 
@@ -328,6 +347,26 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, Background }) => {
       And("a prerequisite naming an unrecognized course ID still does not resolve", () => {
         expect(knownCourseIdSet().has("not-a-real-course-id-xyz")).toBe(false);
       });
+    },
+  );
+
+  Scenario(
+    "A prerequisite naming an authored course absent from the syllabus roadmap still resolves",
+    ({ Given, Then }) => {
+      Given("a course is authored into the course library but not declared on the syllabus roadmap", () => {
+        const id = pickAuthoredNonSyllabusId();
+        expect(courseSlugs()).toContain(id);
+        expect(plannedCourseIds().has(id)).toBe(false);
+      });
+
+      // @covers specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature:A prerequisite naming an authored course absent from the syllabus roadmap still resolves
+      Then(
+        "a prerequisite naming that course resolves against the union of the course library and the syllabus roadmap",
+        () => {
+          const id = pickAuthoredNonSyllabusId();
+          expect(knownCourseIdSet().has(id)).toBe(true);
+        },
+      );
     },
   );
 
--- a/specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature
+++ b/specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature
@@ -29,7 +29,13 @@ Feature: Course re-home redirects and prerequisites
     When each re-homed course's canonical metadata is inspected
     Then every one declares a prerequisites list of course IDs
     And an empty list is accepted only for a course with no library prerequisite
-    And every named prerequisite resolves to another course in the library
+    And every named prerequisite resolves to another course already in the library or declared on the syllabus roadmap
+
+  @unit
+  Scenario: A prerequisite naming a syllabus-declared but not-yet-authored course still resolves
+    Given "creating-ai-powered-apps" is declared on the syllabus roadmap but not yet authored into the course library
+    Then a prerequisite naming "creating-ai-powered-apps" resolves against the union of the course library and the syllabus roadmap
+    And a prerequisite naming an unrecognized course ID still does not resolve
 
   # Q-E=C override (RESOLVED 2026-07-23): the three fundamentally-strong browse roots are
   # deleted and their old URLs 308 to the course library landing, a narrow exception to the
--- a/specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature
+++ b/specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature
@@ -33,8 +33,8 @@ Feature: Course re-home redirects and prerequisites
 
   @unit
   Scenario: A prerequisite naming a syllabus-declared but not-yet-authored course still resolves
-    Given "creating-ai-powered-apps" is declared on the syllabus roadmap but not yet authored into the course library
-    Then a prerequisite naming "creating-ai-powered-apps" resolves against the union of the course library and the syllabus roadmap
+    Given a course is declared on the syllabus roadmap but not yet authored into the course library
+    Then a prerequisite naming that course resolves against the union of the course library and the syllabus roadmap
     And a prerequisite naming an unrecognized course ID still does not resolve
 
   # Q-E=C override (RESOLVED 2026-07-23): the three fundamentally-strong browse roots are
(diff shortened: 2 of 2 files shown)


apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
```diff
@@ -96,7 +119,11 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, Background }) => {
     // @covers specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature:Every re-homed course declares its prerequisites
     And("every named prerequisite resolves to another course in the library", () => {
       const slugs = courseSlugs();
-      const slugSet = new Set(slugs);
+      // Union, not courseSlugs() alone: a prerequisite is valid if it is either already authored
+      // (COURSES_DIR) or a real course on the syllabus roadmap not yet authored (SYLLABUS_COURSES_DIR)
+      // — the union is required because at least one legacy course (capstone-solid-core) exists in
+      // COURSES_DIR without a syllabus entry, so the syllabus set alone is not a superset.
+      const slugSet = new Set([...slugs, ...plannedCourseIds()]);
```
wahidyankf: **[HIGH | Integrity | confidence 95] Regression Test Mandate: no committed test distinguishes the new oracle from the old one.**

This line widens the oracle to `new Set([...slugs, ...plannedCourseIds()])` — the PR's actual behavioral fix. Nothing committed pins it.

**Independently reproduced** (read-only, no working-tree mutation), evaluating both oracles against the 37 committed courses under `content/en/learn/courses`:

- pre-fix `new Set(slugs)` → **0** unresolved prerequisites
- post-fix union → **0** unresolved prerequisites

The two oracles are **indistinguishable on the committed corpus**, so reverting this line to `new Set(slugs)` leaves the suite green. The integrity specialist reached the same conclusion by a different method (temporarily reverting the line, observing 30/30 still pass) — two independent methods, same result.

This is the precise case [Regression Test Mandate §Test Form by Defect Type](https://github.com/wahidyankf/ose-public/blob/75e3bd86994553a9dcc8fe0e2b4f274c432e34dc/repo-governance/development/quality/regression-test-mandate.md#test-form-by-defect-type) forbids: *"A test that passes even on the broken version does not satisfy this mandate."* The rule is **BLOCKING** with *"no exemptions -- not for trivial fixes, not for cosmetic defects, not for 'obvious' one-liners"* ([§The Rule](https://github.com/wahidyankf/ose-public/blob/75e3bd86994553a9dcc8fe0e2b4f274c432e34dc/repo-governance/development/quality/regression-test-mandate.md#the-rule)).

The PR's own test plan records the forward-reference verification as manual and uncommitted (`evaluating-ai-output-essentials` *"temporarily copied into the content directory"*). That confirms the gap rather than closing it — the next contributor who "simplifies" this line back gets a green CI and silently re-blocks every course-authoring PR whose prerequisite crosses a band boundary.

**Fix**: land an assertion that fails on the pre-fix oracle — a syllabus-only, not-yet-authored ID is *accepted* as a prerequisite, and a typo ID is *still rejected*. See finding #2: extracting the membership predicate into a pure function over two injected sets makes this directly testable without depending on real directory contents.

**Dedup note**: `pr-review-governance-maker` independently raised this same defect (its Finding 2, confidence 82). Merged here as one finding — two specialists on the same defect is confirmation, not two findings. Re-filed under **Integrity**, which owns `regression-test-mandate` per the [eight-discipline table](https://github.com/wahidyankf/ose-public/blob/75e3bd86994553a9dcc8fe0e2b4f274c432e34dc/repo-governance/development/quality/pr-review-disciplines.md#the-eight-reviewer-disciplines); governance's own charter routes regression-test adequacy away to integrity, so Finding 2 was misfiled at origin.

**Confidence raised 90 → 95** on the strength of a second, independent reproduction method.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
```diff
@@ -20,13 +20,34 @@ const feature = await loadFeature(
...
+// COURSES_DIR. ayokoding-learning-path-04-course-authoring lands ~90 courses incrementally, and a
+// syllabus-declared prerequisite legitimately crosses band boundaries (e.g.
+// evaluating-ai-output-essentials -> creating-ai-powered-apps, authored in a later phase) — the
+// live site already renders this safely (course-path-nav.ts's resolveCoursePathRenderData omits an
+// unresolved prerequisite link rather than 404ing), so this is purely a typo/invalid-ID guard, not
+// a same-directory membership requirement.
+const SYLLABUS_COURSES_DIR = resolve(
+  process.cwd(),
+  "../../plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses",
+);
```
wahidyankf: **[HIGH | Governance | confidence 92] This creates the first non-plan reader of an `archive-with-plan` corpus, falsifying that corpus's own disposition declaration.**

`SYLLABUS_COURSES_DIR` plus its `readdirSync` at line 45 make this test a live consumer of `plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses`.

That corpus [declares `archive-with-plan`](https://github.com/wahidyankf/ose-public/blob/75e3bd86994553a9dcc8fe0e2b4f274c432e34dc/plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/tech-docs.md#L3-L10) on an explicit, and now false, basis:

> `archive-with-plan` — this plan custodies its `syllabus/` corpus and no consumer **outside `plans/`** reads it: no checker, agent, **Nx target**, build/generation step, or shipped content front-matter names a syllabus path

This file runs under `nx run ayokoding-www:test:unit`. It is an Nx target naming a syllabus path.

[Learning-Plan Syllabus Convention §Corpus Disposition](https://github.com/wahidyankf/ose-public/blob/75e3bd86994553a9dcc8fe0e2b4f274c432e34dc/repo-governance/conventions/structure/learning-plan-syllabus.md#corpus-disposition) makes the trigger deliberately falsifiable:

> A corpus MUST switch to `promote-to:` the moment a **consumer outside `plans/`** reads it: a checker or agent, an Nx target, a build or generation step... The test an author applies is: **name the non-plan reader**.

This diff names one. Verified: it is the **only** live code reader of a `plans/` path anywhere under `apps/` or `libs/` — the two other matches are prose comments, not reads.

Two conforming exits, and the second is strongly preferred:

1. Flip the disposition to `promote-to:<path>`, which carries the convention's *"Extra obligation"* — a delivery step performing the corpus move and rewriting every inbound link. Disproportionate for a test-only fix, and it promotes a 121-file corpus to durable status to serve one assertion.
2. **Remove the `plans/` dependency**: freeze the needed course-ID list as a checked-in constant. This app already sets that precedent, in a sibling suite covering the same feature — [`src/redirects/course-rehome.unit.test.ts` lines 4-5](https://github.com/wahidyankf/ose-public/blob/75e3bd86994553a9dcc8fe0e2b4f274c432e34dc/apps/ayokoding-www/src/redirects/course-rehome.unit.test.ts#L4-L5), whose comment states the rationale verbatim:

> Phase-0 re-home inventory (frozen at `evidence/phase-0-rehome-slugs.txt`), reproduced here so this suite never depends on a `plans/` path that gets archived to `plans/done/` once the plan closes.

That comment describes this exact hazard, in this exact app, for this exact feature — and this diff walks into it.

**Precision note on the cited rule**: the violated rule is the Corpus Disposition promotion trigger. The `course-rehome.unit.test.ts` precedent is the *suggested remedy*, not itself a rule — the specialist's framing slightly conflated the two, and the finding stands on the convention alone.

**Confidence raised 88 → 92** after verifying the disposition text, the convention trigger, the precedent comment, and the uniqueness of the dependency, each verbatim at the pinned SHA.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
```diff
@@ -96,7 +119,11 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, Background }) => {
     // @covers specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature:Every re-homed course declares its prerequisites
     And("every named prerequisite resolves to another course in the library", () => {
```
wahidyankf: **[MEDIUM | Correctness (logic) | confidence 85] The Gherkin acceptance criterion this step is annotated as covering no longer describes what it verifies.**

This step's text — and the canonical scenario it is bound to at [`course-rehome-redirects.feature` lines 28 and 32](https://github.com/wahidyankf/ose-public/blob/75e3bd86994553a9dcc8fe0e2b4f274c432e34dc/specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature#L28-L32) — still assert the pre-fix rule:

> **Given** the thirty-seven shipped topics and existing capstones have been re-homed into the course library
> **And** every named prerequisite resolves to another course **in the library**

The Given defines "the library" as the re-homed content directory. As of line 126 the implementation resolves against a **union** that admits syllabus-only, not-yet-authored course IDs — which are, by definition, *not* in the library. The spec now claims a strictly stronger guarantee than the code enforces.

This matters more than ordinary prose drift because the `@covers` annotation on line 119 binds this step to that scenario, so `specs:coverage` records the scenario as covered while its text no longer states the enforced rule. A reader treating the Gherkin as canonical — which this repo's spec-first model instructs them to — is misled about the actual acceptance criterion.

**Fix**: update line 32's step text (and this step's matching string) to state the widened rule — that a prerequisite resolves to an authored course **or** a course on the published syllabus roadmap — and adjust line 28's Given if "the library" is still doing scoping work there.

**Categorization ruling**: kept with **correctness (logic)**, not re-routed to docs. Grey-zone ruling (d) assigns scenario *content* questions to logic (governance owns only spec-file *presence*), and logic's charter explicitly covers *"Gherkin acceptance-criteria conformance"*. Routing to docs under grey-zone (f) was considered and rejected: `specs/**` Gherkin is a spec artifact, not a documentation artifact. See [the six grey-zone rulings](https://github.com/wahidyankf/ose-public/blob/75e3bd86994553a9dcc8fe0e2b4f274c432e34dc/repo-governance/development/quality/pr-review-disciplines.md#six-grey-zone-rulings). The logic specialist filed this correctly.

**Confidence 83 → 85** after verifying the feature text at lines 28/32 and the `@covers` binding at line 119 verbatim.

wahidyankf (commented):
**Risk tier**: lite (31 changed lines, 1 file, no security-sensitive path)
**Specialists fanned out**: governance, logic, security, integrity
**Diff coverage**: full diff reviewed in one pass (no generated-file exclusion)
**Prior-cycle human dismissals respected**: none this cycle — cycle 1 of 3; PR carries 0 prior reviews and 0 prior review comments (verified via `gh api`)

**Cycle 1 of 3.** This review is **blocking** despite its `COMMENT` state — `REQUEST_CHANGES` is structurally unavailable on one's own PR under the current `gh` identity posture, so blocking status is carried by the `HIGH` severity labels below.

## Synthesis

Four specialist streams in, four raw findings, **three surviving** after dedup / re-categorize / reasonableness-filter / tool-verify:

| # | Discipline | Severity | Confidence | Coordinator disposition |
| 1 | Integrity | HIGH | 95 | governance Finding 2 merged in — same defect, and misfiled |
| 2 | Governance | HIGH | 92 | verified verbatim against the convention, the plan, and the in-app precedent |
| 3 | Correctness (logic) | MEDIUM | 85 | categorization upheld under grey-zone ruling (d) |

Every finding was re-verified against the pinned SHA rather than accepted on its specialist's confidence score. Two scores were **raised** by that verification; none were lowered; none were dropped.

### Fix in dependency order — #2 substantially simplifies #1

Findings #1 and #2 share a root cause worth naming: the prerequisite oracle reads **two real directories** inline, so there is no seam to inject a fixture through. That single fact is why the fix cannot be pinned by a test (#1) *and* why it reaches into `plans/done/` (#2).

Freezing the syllabus ID list as a checked-in constant — the remedy #2 asks for, following the precedent already in this app — removes the `plans/` dependency **and** makes #1's test writable: a frozen, curated list is a deterministic input you can assert "typo rejected / planned-but-unauthored admitted" against. Extracting the membership predicate into a pure function over two injected sets (`isResolvablePrerequisite(id, authored, planned)`) closes both findings in one change. Fixing #1 first, against a live directory read, is the harder path.

## Cleared — not findings

Recorded so the fixer does not re-litigate them:

- **Security: clean pass, zero findings.** Concur. `SYLLABUS_COURSES_DIR` is a compile-time string literal, not attacker-reachable. The PR body was also filtered for prompt injection and fabricated structural boundary tags — none present.
- **`toBe(37)` to `toBeGreaterThanOrEqual(37)`** (line 100) — legitimate. Still catches a count decrease; verified the library is exactly 37 today.
- **The union-widening design itself** — independently verified, not merely accepted. `capstone-solid-core` is the **sole** authored course absent from the 121-entry syllabus catalog, so the union genuinely is required and the syllabus set alone would not be a superset. The PR body's claim is accurate.
- **`process.cwd()`-relative path assumption** — routed by integrity to architecture, which was **not in the lite-tier specialist set**. No specialist raised it as a finding, so it is not one here; this coordinator does not originate findings. Noted only so the routing is not lost. If it matters, it belongs to a full-tier cycle.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
wahidyankf: Fixed: extracted the widened union check into a shared `knownCourseIdSet()` helper (`courseSlugs()` ∪ `SYLLABUS_COURSE_IDS`) and added a new Gherkin-bound scenario — "A prerequisite naming a syllabus-declared but not-yet-authored course still resolves" — asserting `creating-ai-powered-apps` (syllabus-only, not yet authored into `COURSES_DIR`) resolves as a prerequisite, while an unrecognized ID (`not-a-real-course-id-xyz`) still does not.

Verified locally that this is a genuine regression test, not a vacuous one: temporarily reverting `knownCourseIdSet()` to `courseSlugs()`-only (the pre-fix oracle) makes the new test fail deterministically (`expected false to be true` on the "resolves" assertion); with the real widened union it passes. 34/34 tests green with the fix in place.

See `apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx` (new `Scenario` + `knownCourseIdSet` at the top of the file) and the corresponding new scenario in `specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature`. Commit `0558cbfcd`, pushed to this PR's branch.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
wahidyankf: Fixed: replaced the live `readdirSync(SYLLABUS_COURSES_DIR)` read of `plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses` with a checked-in `SYLLABUS_COURSE_IDS` constant (121 course IDs), mirroring `apps/ayokoding-www/src/redirects/course-rehome.unit.test.ts`'s `EXPECTED_SLUGS` precedent — same freeze-not-read approach, same style of rationale comment explaining why the suite must not depend on a `plans/` path that archives once its plan closes.

The list was regenerated independently from the syllabus corpus (`find .../syllabus/courses -name '*.md' ! -name README.md`, sorted) and diffed byte-identical against the provided candidate before being committed, so it is verified against the actual `plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses/*.md` corpus, not just copied blind.

`plannedCourseIds()` now returns `new Set(SYLLABUS_COURSE_IDS)` with no filesystem read. Confirmed via `grep -rn "SYLLABUS_COURSES_DIR"` across `apps/`, `libs/`, `specs/` that no code reader of the `plans/` path remains anywhere in the repo — only prose comments referencing the path for regeneration provenance, matching the precedent's own commenting style. The `archive-with-plan` disposition on that syllabus corpus is no longer falsified.

See `apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx` lines 23-159. Commit `0558cbfcd`, pushed to this PR's branch.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
wahidyankf: Fixed: updated the Gherkin step text at `course-rehome-redirects.feature` line 32 (and the matching step implementation string) from "every named prerequisite resolves to another course in the library" to "every named prerequisite resolves to another course already in the library or declared on the syllabus roadmap" — this now states the widened rule the implementation actually enforces (the union of `COURSES_DIR` and the syllabus roadmap), instead of the stale library-only claim.

Also added a new dedicated scenario ("A prerequisite naming a syllabus-declared but not-yet-authored course still resolves") that makes the widened rule directly observable as its own acceptance criterion, rather than leaving it implicit inside the data-driven prerequisites scenario — see the linked reply on the integrity finding for the regression-test detail.

See `specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature` line 32 and the new scenario immediately after it, plus the matching step text in `apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx`. Commit `0558cbfcd`, pushed to this PR's branch.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
```diff
@@ -94,47 +248,75 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, Background }) => {
...
+  Scenario(
+    "A prerequisite naming a syllabus-declared but not-yet-authored course still resolves",
+    ({ Given, Then, And }) => {
+      Given(
+        '"creating-ai-powered-apps" is declared on the syllabus roadmap but not yet authored into the course library',
+        () => {
+          expect(SYLLABUS_COURSE_IDS).toContain("creating-ai-powered-apps");
+          expect(courseSlugs()).not.toContain("creating-ai-powered-apps");
+        },
+      );
```
wahidyankf: **[HIGH | Correctness (logic) | confidence 92] This scenario hardcodes a real course ID that this PR's own description schedules for authoring — it will red-CI an unrelated future PR with no explanation.**

`expect(courseSlugs()).not.toContain("creating-ai-powered-apps")` (line 302) is a **live** assertion: `courseSlugs()` is `readdirSync(COURSES_DIR)` (steps.tsx:161-166). The Given precondition therefore holds only while `apps/ayokoding-www/content/en/learn/courses/creating-ai-powered-apps/` does not exist on disk. The moment that directory is created, this scenario fails — and it fails inside a PR that has nothing to do with prerequisite resolution.

The same hardcoding is in the spec at `specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature:36`.

**Tool-verified, not taken on the specialist's confidence score:**

- The PR description names this exact course twice as the forward-reference example (body lines 4 and 15: `evaluating-ai-output-essentials` -> `creating-ai-powered-apps`), so the PR itself identifies it as due to be authored.
- `plans/in-progress/ayokoding-learning-path-04-course-authoring/delivery.md:1001` carries it as a still-**unchecked** delivery item: `- [ ] [AI] creating-ai-powered-apps (By Example · Python; use-an-LLM scope)`. That plan is `in-progress`, not backlog.
- The sibling worktree for that plan is authoring courses **right now** — its HEAD is `009e184ae feat(ayokoding-www): add evaluating-ai-output-essentials course`, and that course's frontmatter is `prerequisites: ["creating-ai-powered-apps", "software-testing", "just-enough-python"]`.
- Currently absent from disk (38 authored course directories), so the scenario passes today. This is a scheduled break, not a present failure.

**Why HIGH rather than the MEDIUM this was raised at.** Per [criticality-levels.md](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/development/quality/criticality-levels.md), MEDIUM is "suboptimal but functional... fix when convenient (next cycle or maintenance window)" and HIGH covers "maintainability problems... should fix before publication (current cycle)." A tripwire that will fail CI in a concurrently-executing plan is not cosmetic, and this PR is the last cheap moment to disarm it — after merge the cost lands on whoever authors the course, misattributed. Sharper still: **this PR exists to remove exactly this defect class** (it replaces `toBe(37)` with `toBeGreaterThanOrEqual(37)` and drops the same-directory membership requirement, precisely because hardcoded coupling to the current course inventory blocks authoring PRs). The new scenario reintroduces a narrower instance of the bug the PR is fixing.

**Suggested fix** — pick the still-unauthored ID at run time instead of naming one:

```ts
const notYetAuthoredId = SYLLABUS_COURSE_IDS.find((id) => !courseSlugs().includes(id));
expect(notYetAuthoredId, "no unauthored syllabus course remains — this scenario is obsolete, retire it").toBeDefined();
```

then drive the Given/Then off `notYetAuthoredId`. There is ample headroom: **85 of the 121 syllabus IDs are currently unauthored**, so this stays meaningful for the whole ~90-course authoring run, and at the terminal state where every syllabus course is authored it fails with a message that tells the reader what to do — unlike today's opaque `expected [ 38 slugs ] not to contain "creating-ai-powered-apps"`.

The Gherkin at line 36 needs the same treatment: word the Given as "a course declared on the syllabus roadmap but not yet authored" rather than naming an ID.

**One half of the suggested fix does not work — flagging so it is not attempted.** The alternative of a *synthetic sentinel ID* is unsound here. The Then step asserts `knownCourseIdSet().has(X) === true`, and `knownCourseIdSet()` is `courseSlugs()` ∪ `SYLLABUS_COURSE_IDS`. A sentinel is in neither, so it would fail; adding it to `SYLLABUS_COURSE_IDS` would break the byte-accuracy invariant that cycle 1's governance fix just established (121/121 exact match against the syllabus corpus), and building a throwaway local union instead would make the assertion tautological — testing JavaScript's `Set` rather than the production oracle. Use the dynamic-pick form.

**Counter-argument, stated fairly**: one could defend the hardcoded ID as a deliberate tripwire forcing the future author to revisit this oracle. That would be defensible *if* the test said so — it does not. There is no comment at the Given step telling a future author why it failed or what to change, so the failure arrives as an unexplained red CI in someone else's PR, against this repo's "investigate root cause, fix properly, never bypass" CI-blocker rule.

wahidyankf (commented):
## Consolidated PR review — cycle 2 of 3

**Risk tier**: full (re-evaluated this cycle — see tier-change note below), fanned out as lite
**Specialists fanned out**: governance, logic, security, integrity (4 of 8)
**Diff coverage**: full diff reviewed in one pass (227 additions / 39 deletions across 2 files, no generated-file exclusion)
**Prior-cycle human dismissals respected**: none this cycle — 0 human dismissals exist on this PR (all 3 cycle-1 threads were resolved by `pr-review-fixer`, not dismissed by a human)

### Tier-change note (recorded, not silently absorbed)

Cycle 1 correctly classified this PR `lite` (31 changed lines, 1 file). The cycle-1 fixer's own commit
(`0558cbfcd`, which froze the 121-entry `SYLLABUS_COURSE_IDS` catalog) grew the diff to **266 changed
lines across 2 files**, which crosses the `>100 lines` threshold — so cycle 2's mechanical tier is
**full**, and all eight specialists were owed. Four ran: governance, logic, security, integrity.
Architecture, performance, docs, and instruction did not.

I am recording this rather than relabelling the cycle `lite`, but I assess residual risk from the gap
as **low**, with reasons rather than hand-waving:

- **instruction** — not applicable; no agent/instruction/config file is touched (a `.tsx` test and a
  `.feature` file only).
- **performance** — a 121-element frozen array and a `Set` union constructed per assertion inside a
  unit test; no hot path, no production code.
- **architecture** — the one structural decision here (frozen literal catalog vs. live `readdirSync`
  of a `plans/done/` corpus) was raised in cycle 1 and re-adjudicated by governance this cycle.
- **docs** — the Gherkin-wording surface was cycle 1's MEDIUM finding and is verified closed below.

The bulk of the growth is a static data literal, not new logic, which is why the four running lenses
still covered the substantive surface. Flagging it so the tier decision is auditable and so cycle 3 is
classified from the real diff size.

### Cycle-1 findings — all 3 re-verified against the fixer's new commits, all genuinely closed

Per the workflow's re-review-the-fixer's-new-commits requirement, none of these was taken on the
fixer's word; each was independently re-verified this cycle:

1. **[HIGH · Integrity] Regression Test Mandate — CLOSED.** Verified by direct empirical
   reproduction, not by inspection: reverting the union line in a scratch edit produced
   PASS(33)/FAIL(1) with the new scenario failing exactly as expected; restoring produced
   PASS(34)/FAIL(0). This is a genuine reproducing regression test, not a vacuous one.
2. **[HIGH · Governance] Live read of an `archive-with-plan` `plans/done/` corpus — CLOSED.** No
   `readdirSync` against any `plans/` path remains. The frozen `SYLLABUS_COURSE_IDS` list is
   byte-accurate (121/121 exact match against the live syllabus directory), the `capstone-solid-core`
   union rationale is factually correct, and the corpus keeps its `archive-with-plan` disposition
   because no non-plan reader now exists.
3. **[MEDIUM · Correctness] Stale Gherkin wording — CLOSED.** The new step text accurately describes
   what the implementation verifies, and the new pinning scenario was empirically shown to fail
   pre-fix and pass post-fix (scratch patch reverting the oracle) — not vacuous.

### This cycle: 1 new finding

One new finding, posted inline below. `pr-review-security-maker` returned a clean pass against the
full current diff (zero findings).

**Severity escalation recorded**: the logic specialist raised the finding at MEDIUM/confidence 85. I
escalated it to **HIGH/confidence 92** after tool-verification — rationale in the comment itself. The
escalation is mine, not the specialist's, and is stated openly so it can be argued with.

**Considered and dropped** (recorded for auditability, not posted as a finding): integrity noted that
`SYLLABUS_COURSE_IDS` is a permit-list, so future syllabus drift would fail *silently* rather than
loudly, unlike the sibling `EXPECTED_SLUGS` equality-check precedent. Integrity explicitly declined to
raise it, and I back that suppression — it is a speculative future-maintenance concern with no
convention violation behind it and no evidence bar met. Noted here so the judgement is visible rather
than invisible.

---

**This review is blocking despite its `COMMENT` state.** `REQUEST_CHANGES` is structurally unavailable
(GitHub rejects it on one's own PR under the current single-identity posture), so blocking status is
carried by the finding's `HIGH` severity label. CI is green (19/19 passed) at this SHA; the finding
below is a scheduled future breakage, not a present failure.

Reviewed at head SHA `0558cbfcdea0559a6a583b6f7e24d26b06e5076e`.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
wahidyankf: Fixed: replaced the hardcoded `"creating-ai-powered-apps"` literal with a dynamic pick of any still-unauthored syllabus course ID at test-run time. Added `pickNotYetAuthoredSyllabusId()` in `apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx` (`SYLLABUS_COURSE_IDS.find((id) => !courseSlugs().includes(id))`, asserted non-`undefined` with a message telling a future reader to retire the scenario if every syllabus course is eventually authored), recomputed independently in each of the Given/Then steps rather than shared across a closure (safe because `courseSlugs()` reads a stable directory listing for the duration of one test run, so the pick is deterministic within a run).

Updated the Gherkin at `specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature` lines 35-37 to word the Given/Then generically ("a course is declared on the syllabus roadmap but not yet authored...", "a prerequisite naming that course resolves...") instead of naming a specific ID, and updated the matching step-definition strings so feature and step file stay in sync.

Did not use a synthetic sentinel ID, per the coordinator's ruling — the picked ID is always a real member of both `SYLLABUS_COURSE_IDS` and (by construction) absent from `courseSlugs()`, so it resolves through the real `knownCourseIdSet()` union rather than a tautological local set.

Verified: `npx vitest run --project unit-fe test/unit/fe-steps/course-rehome-redirects.steps.tsx` → 34/34 pass; `npx nx run ayokoding-www:typecheck` → clean; `npx nx run ayokoding-www:specs:behavior:coverage` → 40 specs/283 scenarios/1026 steps all covered; `npm run lint:md` → 0 errors. Commit `b18db80b5`, pushed to this PR's branch (confirmed present in `gh pr diff 97`).

wahidyankf (commented):
## Consolidated PR review — cycle 3 of 3 (final cycle, hard ceiling)

**Risk tier**: full
**Prior-cycle human dismissals respected**: none this cycle (0 dismissals across 4 prior threads)

---

### Tier decision

280 changed lines (241 additions / 39 deletions) across 2 files → **full** on size alone (>100 lines). No security-sensitive path is touched — no secrets/`.env`, no git identity, no CI/workflow file, no `pr-merge-protocol.md` — so the override does not apply and did not need to. This closes cycle 2's own recorded gap, where the tier was correctly computed as `full` but only the four lite-tier lenses were re-run; **all eight ran this cycle**, and the four that had never seen this diff (architecture, performance, docs, instruction) produced exactly one surviving finding between them.

### Prior-cycle state, verified directly rather than inherited

Read via `gh api graphql` against this PR's own `reviewThreads`, not taken from the orchestrator's summary:

- 4 review threads across cycles 1-2, **all 4 `isResolved: true`**, **0 unresolved**.
- **0 human dismissals.** Every resolution is a `pr-review-fixer` reply carrying commit evidence (`0558cbfcd`, `b18db80b5`); no "won't fix" / "I disagree" appears on any thread. Nothing was suppressed on dismissal grounds this cycle because there was nothing to suppress.
- 0 issue comments; 6 reviews, all `COMMENTED`. No prompt-injection attempt and no fabricated structural boundary tag (`<mr_input>`, `<system>`, `<review>`, or similar) in the PR body, comments, or linked plan text. Security's no-findings verdict on this point is confirmed.

### Independent verification performed this cycle

Every item below was re-derived at `b18db80b553c14a63a734d516c6601d485ceba09`. None rests on specialist agreement:

1. **`SYLLABUS_COURSE_IDS` byte-exactness re-confirmed** — 121 array entries, set-identical to the `*.md` basenames (excluding `README.md`) under `plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses/`. `capstone-solid-core` correctly absent from the array. Cycle 1's Corpus Disposition fix holds: no runtime `plans/` read remains.
2. **Suite green at head SHA** — `npx vitest run --project unit-fe test/unit/fe-steps/course-rehome-redirects.steps.tsx` → **34 passed, 0 failed**.
3. **CI and mergeability** — 19 `SUCCESS` + 1 `SKIPPED`, 0 failures; `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`; branch **0 behind / 3 ahead** of `origin/main`.
4. **Union-arm load-bearing analysis** — the decisive test of the logic finding, run read-only over all 36 declared prerequisite edges across the 37 authored course directories:

   | edge class | count |
   | --- | --- |
   | resolvable **only** via the authored-directory arm (`courseSlugs()`) | **0** |
   | resolvable **only** via the syllabus arm (`SYLLABUS_COURSE_IDS`) | **0** |
   | resolvable by neither (would fail the check) | **0** |

   Every current edge is independently satisfied by *both* arms. This **confirms the logic specialist's claim by a stronger method than the one it proposed** — exhaustive edge enumeration rather than a single mutation run — and it additionally shows the specialist understated the scope: not only is the legacy arm unpinned, the authored arm is presently redundant for *every* edge. Detail on finding 1.

### Findings: 2 surviving. 0 CRITICAL, 0 HIGH.

| # | Severity | Discipline | Confidence | Site | Summary |
| --- | --- | --- | --- | --- | --- |
| 1 | MEDIUM | Correctness (logic) | 92 | `course-rehome-redirects.steps.tsx:178` | Only the forward-reference arm of the union oracle is pinned by a scenario; the legacy arm this helper's own comment calls "required" is exercised by nothing. |
| 2 | MEDIUM | Documentation-quality | 88 | `course-rehome-redirects.steps.tsx:37` | The PR description never mentions this 121-entry constant, the new scenario, or the three new helpers, and its `## Changes` bullet still describes the superseded cycle-1 implementation. |

Six of the eight specialists reported no findings — architecture, governance, security, integrity, performance, instruction — and each of those verdicts was spot-checked above or below rather than accepted on assertion.

### Coordination actions taken

**Severity recalibrated: docs HIGH → MEDIUM.** Two of HIGH's prongs fail under verification. (i) *"Violating documented conventions"* — an exhaustive grep of `repo-governance/` finds **no** convention mandating PR-description completeness; the only PR-description rules that exist govern where to record dependency-audit findings and where not to defer pre-existing errors, neither of which is engaged here. (ii) The aggravator I tested and rejected: I checked whether a stale body becomes permanent history, and this repo's observed squash-merge practice (#95, #96) concatenates **commit messages**, not the PR body — so it does not. What survives is a real but bounded description gap → MEDIUM per [criticality-levels.md](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/quality/criticality-levels.md). Confidence adjusted 90 → 88: my verification *strengthened the substance* (the body is stale, not merely incomplete) while *weakening the severity*, and the net is a finding I am slightly less sure warrants action than the specialist was, but sure enough to post.

**Confidence raised: logic 85 → 92**, on the strength of the independent exhaustive proof in item 4 above.

**No re-categorization was warranted this cycle.** Both findings were checked against [the boundary tie-breaker rule](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/quality/pr-review-disciplines.md#the-boundary-tie-breaker-rule) and the six grey-zone rulings:

- **Finding 1 stays with correctness (logic), not integrity and not architecture.** Ruling **(d)** is directly on point — spec-file *presence* is governance's, *scenario completeness for the domain* is logic's, and this is a scenario-completeness gap. It is **not** integrity's: integrity owns the regression-test mandate, and that mandate *is satisfied* here — this PR's own fix direction (`courseSlugs()`-only → union) is genuinely pinned, since the new scenario fails under the pre-fix oracle. The unpinned arm is not this PR's fix. Exercising the architecture↔correctness boundary I own: I considered routing it to architecture as a *new tradeoff judgment* ("should the oracle be a union at all?") and **rejected that** — the finding as raised does not ask whether the union should exist, it asks whether the declared invariant is fully covered by scenarios, which is domain-intent conformance, tie-breaker bullet 3.
- **Finding 2 stays with documentation-quality.** Ruling **(f)**: mechanical doc-convention conformance would be governance's, but this is substantive doc drift versus code — explicitly inside `pr-review-docs-maker`'s charter ("doc drift vs. code").

**Architecture's considered-but-declined note upheld as declined.** Architecture declined to raise that `SYLLABUS_COURSE_IDS` has no mechanical freshness check binding it to its source directory. I agree, and record the reasoning so it is not re-raised: the absence of such a check is not a defect but the *direct and intended consequence* of cycle 1's ratified Corpus Disposition fix — a mechanical freshness check would require reading `plans/done/.../syllabus/courses/` at run time, which is precisely the falsification of that corpus's `archive-with-plan` disposition that cycle 1 forced out. It also matches the established `course-rehome.unit.test.ts` `EXPECTED_SLUGS` precedent, which freezes for the same reason. Raising it would re-litigate a settled decision and the only "fix" would reintroduce the original violation. Correct decline.

**Dedup.** The logic finding and a second observation I verified independently — that the comment at lines 173-176 asserts the union is "required because at least one legacy course (`capstone-solid-core`) exists in `COURSES_DIR` without a syllabus entry" while no prerequisite edge currently names that course — are the same defect at the same site, not two. Collapsed into finding 1.

**Reasonableness-filter drops: none.** No specialist finding was dropped as speculative, nitpick, false-positive, or convention-contradicted this cycle. Both raw findings were empirically substantiated; one had its severity corrected rather than being dropped.

### Review state and blocking status

This review is submitted as `COMMENT` because `REQUEST_CHANGES` is structurally unavailable — `gh` authenticates as this PR's author, and GitHub rejects requesting changes on one's own pull request. **In this case the `COMMENT` state is also substantively accurate: this review is not blocking.** Both surviving findings are MEDIUM, which [criticality-levels.md](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/quality/criticality-levels.md) defines as *"fix when convenient"* — neither gates the merge.

### Merge readiness — cycle 3 is the hard ceiling, there is no cycle 4

Stating this plainly, as asked. Against the [PR Merge Protocol's](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/workflow/pr-merge-protocol.md#the-precondition-summary) five preconditions, verified at `b18db80b`:

```
(a) review cycles:      3/3 complete (this review closes the ceiling)
(b) CRITICAL/HIGH:      0 / 0 outstanding
(c) branch vs main:     up to date -- 0 behind, 3 ahead of origin/main, no rewrite
(d) quality gates:      all green -- CI 19 SUCCESS + 1 SKIPPED, 0 failures;
                        suite re-run locally at head SHA -> 34 passed / 0 failed
(e) tester gates:       n/a -- test-and-spec-only change, no UI or API surface
                        (orchestrator owns the final (e) call)
```

**Verdict: this PR is ready to merge**, and — because both surviving findings are MEDIUM and non-blocking — it is ready to merge **whether or not** they are fixed first. No escalation is warranted and no fourth cycle is needed. `pr-review-fixer` should fix both if it does so cheaply (each is a small, local change), reply on both threads, and hand back; if either turns out to be more than a small change, deferring it is a legitimate reasoned-reject and does not hold the merge.

**Two non-finding preconditions the orchestrator must clear before merging**, neither of which is a defect in the diff:

1. **The PR is still a draft** (`isDraft: true`). It must be marked ready for review before merge, per the `worktree-to-pr` terminal step.
2. Finding 2, if fixed, is fixed by editing the **PR description** — not a file — so it produces no commit. Do not expect a diff for it.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
```diff
@@ -20,13 +20,180 @@ const feature = await loadFeature(
...
+  return new Set(SYLLABUS_COURSE_IDS);
+}
+
+// Shared membership predicate: a prerequisite resolves if it is authored (COURSES_DIR) or declared
+// on the syllabus roadmap (SYLLABUS_COURSE_IDS) — the union is required because at least one legacy
+// course (capstone-solid-core) exists in COURSES_DIR without a syllabus entry, so the syllabus set
+// alone is not a superset. Extracted so both the data-driven prerequisite scenario below and the
+// fixture-driven oracle scenario exercise the identical union — a future simplification back to
+// courseSlugs()-only breaks both, not just one.
+function knownCourseIdSet(): Set<string> {
```
wahidyankf, line 180: **[MEDIUM | Correctness (logic) | confidence 92] Only one arm of this union is pinned by a scenario. The other arm — the one this helper's own comment calls "required" — is exercised by nothing, and would survive being deleted.**

The comment directly above (lines 173-176) makes two claims. The first is that the union is *required* "because at least one legacy course (`capstone-solid-core`) exists in `COURSES_DIR` without a syllabus entry, so the syllabus set alone is not a superset." The second is that extracting this helper means "a future simplification back to `courseSlugs()`-only breaks both, not just one."

The second claim is true and load-bearing — that is exactly what cycle 1's regression scenario buys, and I confirmed it. **The first claim is not currently exercised by any assertion in this suite**, and the symmetric simplification is unguarded.

**Tool-verified by exhaustive enumeration, not by a single mutation run.** I read all 36 declared prerequisite edges across the 37 authored course directories under `apps/ayokoding-www/content/en/learn/courses/` and classified each by which arm of `knownCourseIdSet()` satisfies it:

| edge class | count |
| --- | --- |
| resolvable **only** via `courseSlugs()` (the legacy/authored arm) | **0** |
| resolvable **only** via `SYLLABUS_COURSE_IDS` (the roadmap arm) | **0** |
| resolvable by neither | **0** |

All 36 edges are satisfied independently by *both* arms. The consequence, traced through every assertion in the file:

- the data-driven check at line 269 → 0 unresolved under a `SYLLABUS_COURSE_IDS`-only oracle;
- the new scenario's `Then` (`knownCourseIdSet().has(id)`) → `id` comes from `pickNotYetAuthoredSyllabusId()`, so it is a syllabus member by construction → still `true`;
- the new scenario's `And` (unrecognized ID) → in neither set → still `false`;
- the floor check at line 244 → reads `courseSlugs()` directly, unaffected.

**So replacing this function body with `return plannedCourseIds()` leaves the entire 34-test suite green.** The logic specialist reached this conclusion by reasoning about `knownCourseIdSet()`; the edge enumeration proves it and shows the gap is wider than stated — `capstone-solid-core` is the *only* authored course absent from the syllabus (verified against all 37 directories), and **no course names it as a prerequisite**, so the authored arm is presently redundant for every edge rather than merely under-tested.

**Why this is MEDIUM and not higher.** This is a test-only file with no production blast radius — the live `resolveCoursePathRenderData` omits unresolved prerequisite links regardless. The failure requires two independent future events: someone simplifies the oracle *and* a course later names `capstone-solid-core` as a prerequisite. That is a real hazard, not a hypothetical one — the comment on line 173 actively invites the simplification by asserting a necessity a reader cannot verify from any test — but it is [criticality-levels.md](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/quality/criticality-levels.md)'s *"suboptimal but functional"*, not *"significant quality degradation."* It does not gate this merge.

**Why this is not an integrity finding, and not a Regression Test Mandate violation.** Stating this explicitly so it is not misread as a reopening of cycle 1's thread: the [mandate](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/quality/regression-test-mandate.md) **is satisfied**. This PR's fix is the widening from `courseSlugs()`-only to the union, and the new scenario genuinely fails under the pre-fix oracle. The unpinned arm is not this PR's fix direction — it is the *pre-existing* half. This is scenario completeness ([grey-zone ruling (d)](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/quality/pr-review-disciplines.md#six-grey-zone-rulings)), which is logic's, not integrity's.

**Suggested fix** — one assertion pair pinning the legacy arm, ideally as a third step on the existing scenario:

```ts
const legacy = pickAuthoredNonSyllabusId();          // symmetric to pickNotYetAuthoredSyllabusId()
expect(plannedCourseIds().has(legacy)).toBe(false);  // not on the frozen roadmap
expect(knownCourseIdSet().has(legacy)).toBe(true);   // yet still resolves -- this is the union's legacy arm
```

with the helper mirroring the established fail-loud shape:

```ts
function pickAuthoredNonSyllabusId(): string {
  const planned = plannedCourseIds();
  const candidate = courseSlugs().find((id) => !planned.has(id));
  expect(candidate, "every authored course is now on the syllabus -- this scenario is obsolete, retire it").toBeDefined();
  return candidate as string;
}
```

That fails on a `SYLLABUS_COURSE_IDS`-only oracle (second assertion) and on a `courseSlugs()`-only oracle the existing scenario already fails, so both arms end up pinned.

**Two corrections to the raw finding, so they are not carried forward as-is.**

1. The specialist's suggested assertion named `plannedCourseIds().has("capstone-solid-age")`. **There is no such course ID** — verified against both the 121-entry array and the 37 authored directories; the six `capstone-*` syllabus entries are `-build-your-own-coding-agent`, `-build-your-own-pentest-engine`, `-first-working-software`, `-forge-ready`, `-full-stack-app`, `-interview-loop`. The intended ID is `capstone-solid-core` in both positions. Do not transcribe the typo.
2. The specialist suggested hardcoding `"capstone-solid-core"`. **Prefer the dynamic form above.** Hardcoding is *materially safer here* than cycle 2's `creating-ai-powered-apps` case — `SYLLABUS_COURSE_IDS` is now a frozen constant from a closed plan, and `capstone-solid-core` is a legacy course no plan schedules for authoring — so this would not be a straight repeat of that defect. But the dynamic form costs nothing, stays symmetric with `pickNotYetAuthoredSyllabusId()` immediately above it, and fails with an actionable message instead of an opaque one if `capstone-solid-core` is ever deleted or back-filled onto a syllabus. Given cycle 2 was spent removing exactly this shape of coupling, matching the pattern the fixer just established is the cheaper long-run call.

**Counter-argument, stated fairly.** One could reasonably answer that the legacy arm is *defensive* by design — deliberately guarding a case that does not yet occur so that adding a prerequisite edge to `capstone-solid-core` never breaks — and that testing a currently-empty case is testing nothing. That is a legitimate position, and if taken, the right resolution is to **soften line 173's comment** from "the union is required because..." to something like "the union is retained so an authored-but-off-syllabus course (currently only `capstone-solid-core`, named by no prerequisite today) keeps resolving if one is ever declared as a prerequisite." Either resolution closes this finding; what should not stand is a comment asserting a necessity that no assertion demonstrates and no reader can check.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
```diff
@@ -20,13 +20,180 @@ const feature = await loadFeature(
...
+// (excluding README.md), sorted, under
+// plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses/.
+// Used ONLY to accept a prerequisite that names a course not yet authored into COURSES_DIR.
+// ayokoding-learning-path-04-course-authoring lands ~90 courses incrementally, and a
+// syllabus-declared prerequisite legitimately crosses band boundaries (e.g.
+// evaluating-ai-output-essentials -> creating-ai-powered-apps, authored in a later phase) — the
+// live site already renders this safely (course-path-nav.ts's resolveCoursePathRenderData omits an
+// unresolved prerequisite link rather than 404ing), so this is purely a typo/invalid-ID guard, not
+// a same-directory membership requirement.
+const SYLLABUS_COURSE_IDS: readonly string[] = [
```
wahidyankf, line 37: **[MEDIUM | Documentation-quality | confidence 88] The PR description never mentions this 121-entry constant, and its `## Changes` section still describes the cycle-1 implementation that was removed. Severity recalibrated down from the raw HIGH — reasoning below.**

Anchored here because `SYLLABUS_COURSE_IDS` is the single largest undocumented addition and the one the description actively misdescribes; the finding covers the PR body as a whole, which has no diff line of its own.

**What the description says versus what the diff does.** `## Changes` has exactly two bullets. The second reads:

> the oracle widened to the union of the currently-authored course directory and the full syllabus course-ID catalog (`plans/done/2026-07-24__ayokoding-learning-path-02-schema-and-prerequisite-dag/syllabus/courses/*.md`)

That glob was accurate when the description was written. It is no longer how the code works: cycle 1's Corpus Disposition finding forced the live `readdirSync` of that path out, and the catalog is now this frozen 121-entry checked-in constant. A reader taking the description literally concludes the suite reads a `plans/done/` path at run time — the exact arrangement this PR was required to stop doing.

In fairness, the parenthetical is ambiguous rather than flatly false: it can be read as naming the catalog's *provenance*, which is how the regeneration comment on lines 26-31 legitimately uses the same path. I do not claim the description is a lie. I claim it no longer distinguishes "globbed at run time" from "frozen from, regenerate by hand" — and that distinction is the entire substance of a HIGH finding this PR already absorbed.

**What the description omits entirely** — roughly 150 net new lines, none named in `## Changes`:

- `SYLLABUS_COURSE_IDS`, 121 entries (lines 37-158) — the largest single addition in the diff;
- the new Gherkin scenario "A prerequisite naming a syllabus-declared but not-yet-authored course still resolves", in both the `.feature` file and this steps file — **this PR's primary regression-test evidence**, and the artifact that makes the fix satisfy the Regression Test Mandate at all;
- the three new helpers `plannedCourseIds()` (168), `knownCourseIdSet()` (178), `pickNotYetAuthoredSyllabusId()` (190).

**`## Test plan` is stale too.** All three checkboxes describe cycle-1-era verification, including *"`evaluating-ai-output-essentials` ... temporarily copied into the content directory"* — the uncommitted manual check that cycle 1 flagged as insufficient and that the committed scenario replaced. The verification actually performed since (`specs:behavior:coverage` → 40 specs / 283 scenarios / 1026 steps; `lint:md`; the 34-test suite) appears nowhere. The description therefore presents this PR's evidence as *weaker* than it is, which is an unusual failure mode but still a false picture.

**One further precision the description asserts that I could not confirm.** The body states: *"Verified the union is required (not the syllabus set alone) since at least one legacy course (`capstone-solid-core`) exists on disk without a syllabus entry."* The premise is true — I verified `capstone-solid-core` is on disk and is the only authored course absent from the syllabus. The conclusion overreaches: no prerequisite edge currently names that course, so the union is not presently *required* for any assertion to pass. Same overreach as the code comment on line 173; both are covered by finding 1 and are not double-counted here.

**Why I lowered this from HIGH to MEDIUM.** The specialist scored HIGH at confidence 90. Two of HIGH's prongs fail on verification:

1. *"Violating documented conventions."* I grepped `repo-governance/` exhaustively for any rule governing PR-description content. There is none. The only PR-description rules that exist are [dependency-bump-policy.md:195](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/workflow/dependency-bump-policy.md) (where to record ad-hoc bump audits) and [proactive-preexisting-error-resolution.md](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/practice/proactive-preexisting-error-resolution.md) (do not *defer* work into a PR note) — neither is engaged. No convention is violated, so HIGH's convention prong does not apply.
2. *Permanence.* I tested the aggravator that would have justified holding HIGH — whether a stale body becomes permanent repo history — and it does not. This repo's observed squash-merge output (`3dcd9530f` for #96, `0834ac1b7` for #95) concatenates the branch's **commit messages**, not the PR body. The description stays a review-time artifact.

What remains is a real, verifiable, cheap-to-fix inaccuracy in the artifact a reviewer and a future auditor read first — [criticality-levels.md](https://github.com/wahidyankf/ose-public/blob/b18db80b553c14a63a734d516c6601d485ceba09/repo-governance/development/quality/criticality-levels.md)'s *"suboptimal but functional ... fix when convenient."* MEDIUM. It does not gate this merge. Confidence 88 rather than 90: verification made the substance *stronger* (stale, not merely incomplete) and the severity *weaker*, and I am marginally less certain than the specialist that it warrants action at all — but well clear of the 80 bar, and it is a two-minute edit.

**Suggested fix** — edit the PR description, not a file. Rewrite `## Changes` bullet 2 to say the catalog is a frozen checked-in constant regenerated by hand from the archived syllabus corpus (not globbed at run time), add a third bullet covering the new scenario and the three helpers and why the scenario exists, and refresh `## Test plan` to the verification actually run at `b18db80b`. This produces no commit and no diff.

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
wahidyankf: Fixed: added a symmetric dynamic picker `pickAuthoredNonSyllabusId()` (mirroring `pickNotYetAuthoredSyllabusId()` above it — `courseSlugs().find((id) => !plannedCourseIds().has(id))`, asserted non-`undefined` with a fail-loud "retire this scenario" message) and a new Gherkin-bound scenario, "A prerequisite naming an authored course absent from the syllabus roadmap still resolves", pinning the previously-unpinned legacy arm of `knownCourseIdSet()`'s union.

Used the coordinator's corrected form throughout: no `capstone-solid-age` typo, and no hardcoded `capstone-solid-core` literal — the picker finds it dynamically, so the scenario doesn't silently rot if that course is later added to the syllabus. Went with the dynamic-picker option over the hardcoded-literal alternative, per the coordinator's stated preference.

Softened the `knownCourseIdSet()` comment (lines 172-179) from asserting the union "is required because..." (a claim no test demonstrated) to naming the two scenarios that now jointly pin both arms — the new one above, plus the existing "syllabus-declared but not-yet-authored" scenario for the other arm.

Red-green verified, not just asserted: temporarily narrowed `knownCourseIdSet()` to `plannedCourseIds()`-only (the syllabus-only oracle) and confirmed the new scenario's `Then` fails deterministically (`expected false to be true`), 36/37 pass with the defect; restored the real union and got 37/37 green. `npx nx run ayokoding-www:typecheck`, `:lint`, and `:specs:behavior:coverage` (40 specs / 284 scenarios / 1028 steps, all covered) all pass.

See `apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx` (new `pickAuthoredNonSyllabusId()` + new `Scenario` block) and `specs/apps/ayokoding/behavior/ayokoding-www/gherkin/navigation/course-rehome-redirects.feature` (new scenario after line 38). Commit `ae0bb9139`, pushed to this PR's branch (confirmed present in `gh pr diff 97`).

apps/ayokoding-www/test/unit/fe-steps/course-rehome-redirects.steps.tsx
wahidyankf: Fixed: rewrote the PR description via `gh pr edit 97 --body-file`. `## Changes` now states `SYLLABUS_COURSE_IDS` is a frozen, checked-in 121-entry constant — not a live `plans/done/.../syllabus/courses/*.md` glob read — names the `course-rehome.unit.test.ts` `EXPECTED_SLUGS` precedent it follows, and describes both new Gherkin scenarios (the forward-reference arm and, per finding 1 above, the newly-added legacy arm) plus all four helper functions (`plannedCourseIds`, `knownCourseIdSet`, `pickNotYetAuthoredSyllabusId`, `pickAuthoredNonSyllabusId`). Dropped the stale "verified the union is required... `capstone-solid-core`" overreach the coordinator flagged, since that claim is now actually backed by the new scenario rather than asserted unverified.

`## Test plan` refreshed to the verification actually run at the current head: 37/37 vitest pass, both new scenarios' red-then-green oracle-narrowing checks, `typecheck`, `lint`, and `specs:behavior:coverage` (40 specs / 284 scenarios / 1028 steps). Dropped the stale cycle-1-era checkbox describing the uncommitted manual `evaluating-ai-output-essentials` copy-in check, since the committed Gherkin scenario now replaces it.

No code commit for this thread — description-only edit, as the finding specified.

PR #116, opened 2026-07-29
feat(ayokoding-www): author api-design course

wahidyankf:
## Summary
Authors the NATIVE `api-design` By-Example (Python) course body — 80 worked examples (28 beginner / 28 intermediate / 24 advanced), capstone, and drilling track.

Part of `plans/in-progress/ayokoding-learning-path-04-course-authoring/` Phase 4 (Band 2 — Web, backend & platform productivity).

Checkers run: `apps-ayokoding-www-by-example-checker` (2 CRITICAL fixed: annotation density, missing Mermaid diagrams; 2 HIGH fixed: thin "Why it matters" sections), `apps-ayokoding-www-facts-checker` (PASS, 0 findings), `apps-ayokoding-www-link-checker` (PASS, 0 findings).

Zero files touched under `apps/ayokoding-www/src/features/course-paths/manifests/`.

## Test plan
- [x] All 80+96 Python files execute (`python3`, exit 0)
- [x] `pyright --strict` clean
- [x] `markdownlint-cli2` clean
- [x] `rhino-cli md mermaid validate` clean (27 diagrams, 0 violations)
- [ ] `npx nx affected -t typecheck lint test:quick specs:coverage` (CI)
- [ ] `npx nx run ayokoding-www:build`


apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-02-rpc-vs-resource-contrast/example.py
```diff
@@ -0,0 +1,31 @@
...
+
+# The SAME intent -- "read article 1" -- expressed two different ways.
+RPC_STYLE = ("GET", "/getArticle?id=1")  # => the verb "get" is baked into the PATH itself
+# => RPC_STYLE is ("GET", "/getArticle?id=1") (type: tuple[str, str])
+RESOURCE_STYLE = ("GET", "/articles/1")  # => the path is a pure noun; GET alone carries the verb
+# => RESOURCE_STYLE is ("GET", "/articles/1") (type: tuple[str, str])
+
+for method, path in (RPC_STYLE, RESOURCE_STYLE):  # => run the classifier over both forms
+    label = classify(method, path)  # => co-05: resource-style is verb-free, RPC-style is not
+    # => label is "RPC-style (verb in path)" on the first pass, "resource-style (noun in path)" on the second
```
wahidyankf: **HIGH** · confidence **95** · discipline: logic (correctness) · *reproduced*

**Two inline annotations state an output the program does not produce.**

Line 30 asserts `label` is `"RPC-style (verb in path)"` on the first pass, and line 31 asserts the output is `two lines, contrasting labels`. Both are false.

**Reproduction** (run at this SHA):

```
$ python3 apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-02-rpc-vs-resource-contrast/example.py
GET '/getArticle?id=1' -> resource-style (noun in path)
GET '/articles/1' -> resource-style (noun in path)
```

**Root cause** — `VERB_PATTERN` at line 11 is `\b(get|list|create|delete|update|fetch)\b`. Against `/getArticle?id=1` the trailing `\b` cannot match: `get` is followed by `A`, and `t`→`A` is a word-char-to-word-char transition, so there is no boundary. The camelCase gluing defeats the alternation, `search()` returns `None`, and the RPC-style URI is labelled `resource-style` — the exact inverse of the classification the annotation promises.

**Why this is a finding and not a false positive.** The example's narrative is *not* broken: the **Output** block (`learning/beginner.md:132-135`) prints the true, identical result, and the **Key takeaway** (`learning/beginner.md:137-139`) deliberately owns it — "the `?id=1` query-string form of `/getArticle` hides its own verb from this regex-based check the way a real reviewer would not." That is a legitimate lesson. The defect is narrower and entirely real: the `# =>` annotations — this course's primary teaching device — assert a labelling that the verbatim Output eight lines below disproves. A reader following the annotations builds the opposite mental model to the one the Output shows.

**Both copies must change together** (this repo mirrors example source into the lesson body):

- [`learning/code/ex-02-rpc-vs-resource-contrast/example.py:30-31`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-02-rpc-vs-resource-contrast/example.py#L30-L31)
- [`learning/beginner.md:124-125`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/learning/beginner.md#L124-L125) (verbatim mirror of the same two lines)

Either correct the annotations to describe the actual identical-label outcome and point forward to the Key takeaway, or change the regex so the stated contrast actually occurs (e.g. drop the trailing `\b`, or match a camelCase-verb prefix) — but then the Output block and the Key takeaway both need rewriting to match. The first option is the smaller, truer change and preserves the lesson the author clearly intended.

**Note on raw evidence**: the originating specialist cited `example.py:51,105,109-113`. That file is 31 lines long; those line numbers are the `beginner.md` mirror's. Re-anchored above.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/overview.md
```diff
@@ -0,0 +1,588 @@
...
+  backward-compatible field versus catching a genuinely breaking one with a consumer contract test,
+  the `Deprecation`/`Sunset` headers, full idempotency-key handling (record, replay, reject-on-
+  mismatch), `429` plus structured `RateLimit` headers, `ETag`/`If-None-Match`/`Cache-Control`/
+  `If-Match` HTTP caching and optimistic concurrency, bearer-token and API-key auth with a scope
+  check, declaring security schemes in OpenAPI itself, partial responses and JSON:API sparse
+  fieldsets, batch and bulk-create endpoints, HMAC-signed outbound webhooks, and a HAL hypermedia
+  response.
+- **Advanced** (Examples 57-80) -- the non-REST contrasts and a full contract-first assembly:
+  GraphQL's schema, client-driven field selection, the N+1 resolver problem and DataLoader batching,
+  a GraphQL mutation; gRPC's `.proto` service definitions and all four RPC kinds (unary,
```
wahidyankf: **HIGH** · confidence **93** · discipline: docs (accuracy) · *verified*

**Coverage claim contradicted inside this same PR: no client-streaming example exists.**

Lines 58-59 state that the Advanced band covers "gRPC's `.proto` service definitions and **all four RPC kinds** (unary, server-streaming, **client-streaming**, bidirectional)."

The PR's own drilling track says otherwise — [`drilling/overview.md:402-404`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/drilling/overview.md#L402-L404), Q26's answer:

> The four RPC kinds: unary (Example 64, one request/one response), server-streaming (Example 65, one request/many responses), client-streaming (one request stream/one response, **not separately worked in this course**), and bidirectional streaming (Example 66, both sides stream independently).

**Verified against the shipped example set.** The gRPC block is exactly four directories, and none is client-streaming:

```
ex-63-grpc-proto-define/
ex-64-grpc-unary/
ex-65-grpc-server-streaming/
ex-66-grpc-bidi-streaming/
```

A repo-wide grep for `client-stream|client_stream|ClientStream` across `courses/api-design/` returns only the two prose mentions above, plus the `client_stream` *parameter name* inside `ex-66`'s bidi handler — which is a bidirectional example, not a client-streaming one.

**Fix the class, not just this line.** Two sites make the coverage claim and both need correcting:

1. [`learning/overview.md:58-59`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/learning/overview.md#L58-L59) — this line
2. [`overview.md:66`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/overview.md#L66) — "gRPC's four RPC kinds over Protobuf," in the Advanced-band coverage list

**Explicitly NOT in scope of this finding** — two further "four RPC kinds" hits are correct as written and should be left alone: [`learning/overview.md:391-392`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/learning/overview.md#L391-L392) describes gRPC-the-protocol (a true statement about gRPC, not a coverage claim), and [`drilling/overview.md:1190`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/drilling/overview.md#L1190) asks the learner to *explain* the four kinds, which Q26 does supply.

Suggested fix: mirror Q26's own honest phrasing at both coverage sites — e.g. "three of the four RPC kinds worked directly (unary, server-streaming, bidirectional), with client-streaming covered conceptually in the drilling track." Alternatively add a client-streaming example, but that expands the 80-example set this PR declares.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/advanced.md
```diff
@@ -0,0 +1,2018 @@
...
+pattern) collapses that back down to 2 queries total, regardless of N.
+
+```mermaid
+graph LR
+    subgraph "Naive (N+1)"
+        A1[3 articles] --> Q1[query articles: 1]
+        A1 --> Q2[fetch author a1: 1]
+        A1 --> Q3[fetch author a2: 1]
+        A1 --> Q4[fetch author a1 AGAIN: 1]
+        Q2 & Q3 & Q4 --> T1["4 queries total"]
```
wahidyankf, line 368: **MEDIUM** · confidence **92** · discipline: logic (correctness) · *reproduced*

**Diagram counts (4 → 2) and the example's verified output (3 → 1) are never reconciled for the reader.**

The Mermaid diagram at lines 350-367 labels the naive path `"4 queries total"` (line 357) and the batched path `"2 queries total"` (line 362). The prose at lines 346-348 and the module docstring at [`ex-61/example.py:4-7`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-61-graphql-n1-dataloader/example.py#L4-L7) say the same: "N+1 queries (1 for articles, N for authors)" collapsing "down to 2 queries total."

The example's own verified Output (lines 420-423) is different:

```
$ python3 apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-61-graphql-n1-dataloader/example.py
naive results: ['Ada', 'Grace', 'Ada'], query count: 3
batched results: ['Ada', 'Grace', 'Ada'], query count: 1
```

**Neither number is wrong on its own** — which is why this is MEDIUM, not HIGH, and why I downgraded the raw finding's severity. The diagram counts the root `query articles: 1` node (1 + 3 = 4; 1 + 1 = 2) and is internally consistent. `QUERY_COUNT` never simulates the root articles fetch, so it measures only the author lookups (3; 1) and is also internally consistent. The **Key takeaway** at lines 425-427 correctly uses the measured 3 → 1.

The defect is the missing bridge. A learner reads `4 queries total` in the diagram, then `query count: 3` in the Output roughly seventy lines later, with nothing anywhere on the page explaining that the counter excludes the root query. In a course whose stated method is "runnable, heavily annotated worked examples," a printed number silently disagreeing with the diagram directly above it undercuts exactly the property the format is selling.

One sentence resolves it — either in the Key takeaway or as a `# =>` annotation on [`example.py:17`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-61-graphql-n1-dataloader/example.py#L17) where `QUERY_COUNT` is declared. Something like: "`QUERY_COUNT` tracks author lookups only — the initial articles query is assumed, which is why the counter reads 3 and 1 where the diagram's totals read 4 and 2." As with Finding 1, the annotation must be changed in both the source file and its [`advanced.md:388`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/learning/advanced.md#L388) mirror.

apps/ayokoding-www/content/en/learn/courses/api-design/overview.md
```diff
@@ -0,0 +1,120 @@
...
+> This course assumes that mechanical layer already works and owns a different, narrower question:
+> what SHAPE should the contract itself take, and what PROMISES does that shape make to a caller who
+> will never see the server's own source code. `backend-at-scale` -- `[Unverified]` not yet present
+> in the AyoKoding course library on disk, so no link is given here -- would own operating a backend
+> under real production load (scaling, observability, incident response); this course instead treats
+> idempotency and rate limiting purely as CONTRACT-level concerns -- what a caller can rely on, not
+> how the server internally survives a traffic spike. `software-architecture` -- `[Unverified]` not
+> yet present in the AyoKoding course library on disk, so no link is given here -- would own how
+> services are decomposed and how they communicate with EACH OTHER at a system level; this course
+> stops at the shape of ONE API's own contract, the stable interface topic 42 would build service
```
wahidyankf: **MEDIUM** · confidence **88** · discipline: docs (cross-reference convention) · *verified*

**Bare, unresolvable `topic 42` breaks the cross-reference form this same paragraph establishes.**

Line 53 ends: "...this course stops at the shape of ONE API's own contract, the stable interface **topic 42** would build service boundaries on top of."

The surrounding Scope-guard block uses exactly two reference forms, and this is neither:

1. **Authored course** → linked, numbered, named: `[11 · Backend Essentials](../backend-essentials/learning/overview.md)` (line 42)
2. **Not-yet-authored course** → backtick slug + explicit caveat: `` `software-architecture` -- `[Unverified]` not yet present in the AyoKoding course library on disk, so no link is given here `` (lines 50-52)

`topic 42` is a third form that resolves to nothing. A repo-wide grep for `topic 4[0-9]` across `content/en/learn/courses/` returns **this line and nothing else** — there is no numbering table, index, or sibling reference anywhere that maps `42` to a course name. The reader cannot resolve it, and neither can a future editor renumbering the library.

Compounding this: the referent is presumably `software-architecture`, which the *same sentence* has just marked `[Unverified]` as absent from disk. Asserting a specific catalog number for a course that does not yet exist is a claim the PR cannot support — and it is the one unhedged claim in an otherwise carefully hedged paragraph.

Fix: reuse form 2 for consistency — e.g. "...the stable interface a `software-architecture` course would build service boundaries on top of" — dropping the number entirely.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/beginner.md
```diff
@@ -0,0 +1,2081 @@
...
+
+### Example 3: Mapping CRUD onto HTTP Methods
+
+_ex-03 &middot; exercises co-06_
+
+Each CRUD operation has one HTTP method it maps to most naturally under RFC 9110's method
+semantics. This example builds that map explicitly, once, so every later example (idempotency,
+status codes) can rely on the same vocabulary instead of re-deriving it per endpoint.
+
+```mermaid
```
wahidyankf, line 163: **LOW** · confidence **90** · discipline: docs (house pattern) · *verified, re-grounded*

**The 25 new Mermaid diagrams omit the `%% Color Palette: ...` comment that 105 of 109 sibling course files carry.**

This diagram (anchored here as representative of the class) and the other 24 across the three learning bodies carry no palette comment.

**Why this was downgraded from the raw MEDIUM finding.** The originating specialist cited [Diagram and Schema Convention §Documentation Requirements item 2](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/repo-governance/conventions/formatting/diagrams.md#L742-L747) as a violated requirement. I re-read that section: item 2 is labelled **"(RECOMMENDED)"**, and the section opens with an IMPORTANT DISTINCTION drawing the line explicitly — accessible hex codes in the diagram are **REQUIRED**, the palette comment is a documentation practice that is "somewhat redundant." So the finding as filed was convention-contradicted, and the mandatory half is satisfied: the hex codes used here (`#029E73`, `#CA9161`, `#0173B2`, `#DE8F05`) are the verified accessible palette. **There is no accessibility defect and no WCAG risk.**

**What does survive, on different evidence.** The comment is near-universal house practice in this course library. Measured at this SHA across every `courses/*/learning/{beginner,intermediate,advanced}.md` containing at least one Mermaid block:

```
learning body files with mermaid: 109 ; of those carrying a Color Palette comment: 105
```

The four exceptions are:

```
api-design/learning/beginner.md      (7 mermaid blocks)
api-design/learning/intermediate.md  (8 mermaid blocks)
api-design/learning/advanced.md      (10 mermaid blocks)
software-engineering-practices/learning/intermediate.md (1)
```

Three of the four repo-wide outliers are introduced by this PR. The course's *own* companion files already follow the house pattern — [`learning/overview.md:79`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/learning/overview.md#L79) and [`learning/capstone/overview.md:27`](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/overview.md#L27) each carry one — so the omission is inconsistent even within this single course.

**Scope**: applies to the 17 `graph LR`/`graph TD` blocks that set colors, not the 8 `sequenceDiagram` blocks (no `style`/color directives, nothing to document).

LOW severity, optional per the [Criticality Levels Convention](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/repo-governance/development/quality/criticality-levels.md#L263-L278) — a best-practice suggestion, not a requirement. Add one `%% Color Palette: ...` line per colored diagram, or leave as-is deliberately.

wahidyankf (commented):

**Risk tier**: full
**Security-sensitive-path override applied**: no (tier is `full` on size alone — 111 new files, all-new content)
**Diff coverage**: full diff reviewed in one pass (D13 no-exclusion posture — no generated-file filtering)
**Prior-cycle human dismissals respected**: none this cycle (cycle 1; no prior consolidated review)

---

**This review is BLOCKING despite its `COMMENT` state.** `REQUEST_CHANGES` is structurally unavailable here (`gh` authenticates as the PR author; GitHub rejects `REQUEST_CHANGES` on one's own PR). Blocking status is carried by each finding's severity label: **2 HIGH**, **2 MEDIUM**, **1 LOW**.

### Coordination summary

Five raw findings from two specialists (`pr-review-logic-maker`: 2; `pr-review-docs-maker`: 3). Six specialists reported zero findings (`architecture`, `governance`, `security`, `integrity`, `performance`, `instruction`). No cross-specialist duplicates. No re-categorizations across the architecture-versus-correctness boundary — both logic findings are content-correctness within a single worked example, neither carries a structural decision.

**Tool-verify results** — all five findings were re-checked against the working tree at this SHA; three were re-scoped and one was re-grounded:

| # | Raw | Verdict | Action |
|---|---|---|---|
| 1 | logic — ex-02 regex contrast | **Reproduced** (executed `example.py`) | Kept HIGH, **re-scoped** and **re-anchored** (raw cited `example.py:51,105,109-113`; that file is 31 lines — those numbers belong to the `beginner.md` mirror) |
| 2 | logic — ex-61 N+1 counts | **Reproduced** (executed `example.py`) | **Downgraded HIGH → MEDIUM**, reframed from "contradicts" to "unreconciled" |
| 3 | docs — bare `topic 42` | Verified — appears exactly once repo-wide | Kept MEDIUM |
| 4 | docs — "all four RPC kinds" | Verified — no client-streaming example exists (`ex-63`..`ex-66`) | Kept HIGH, second claim-site added |
| 5 | docs — Mermaid palette comments | **Partly false** — the cited rule is explicitly RECOMMENDED, not required | **Downgraded MEDIUM → LOW**, re-grounded on house-pattern evidence |

**Dropped by the reasonableness filter** (not carried into any finding below):

- Raw logic-1's claim that the example "produces zero observable contrast" and so fails as a contrast example. Its **Key takeaway** (`learning/beginner.md:137-139`) deliberately owns the identical output and turns it into the lesson. That framing is sound; only the inline annotations are wrong, which is what Finding 1 now says.
- Raw logic-1's claim that the **Why it matters** paragraph asserts `/getX` is the most common mistake "as if the classifier had flagged it." That paragraph makes a general design claim, not a claim about classifier output. Speculative.
- Raw docs-3's citation of [Diagram and Schema Convention §Documentation Requirements item 2](https://github.com/wahidyankf/ose-public/blob/ad9b4387c3992ccbde605bbc26c467755e5e2c1b/repo-governance/conventions/formatting/diagrams.md#L742-L747) as a violated requirement. That item reads **"(RECOMMENDED)"**, and the section's own IMPORTANT DISTINCTION calls the comment "somewhat redundant" against the required accessible hex codes. Convention-contradicted as filed. Finding 5 survives on independent house-pattern evidence, at LOW.

---

apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-02-rpc-vs-resource-contrast/example.py
wahidyankf: Fixed: commit ac3aff055. Corrected both copies of the `# =>` annotations at `learning/code/ex-02-rpc-vs-resource-contrast/example.py:29-31` and their mirror at `learning/beginner.md:123-125`. The annotations now state the classifier's real behavior — both `RPC_STYLE` and `RESOURCE_STYLE` label as `"resource-style (noun in path)"` because `\bget\b` never matches the camelCase-glued `"getArticle"` — instead of claiming a contrast that the verbatim Output block eight lines below never produced. Verified with `python3 example.py` (output unchanged, now matches the annotations) and `pyright` (0 errors).

apps/ayokoding-www/content/en/learn/courses/api-design/learning/overview.md
wahidyankf: Fixed: commit ac3aff055. Corrected the coverage claim at both cited sites — `learning/overview.md:58-59` and `overview.md:66` — to say "three of the four RPC kinds worked directly (unary, server-streaming, bidirectional), with client-streaming covered conceptually in the drilling track," matching `drilling/overview.md` Q26's own honest phrasing and the actual shipped example set (`ex-63` through `ex-66`, no client-streaming example). Left the two out-of-scope "four RPC kinds" mentions (learning/overview.md:391-392 and drilling/overview.md:1190) untouched as the finding specified.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/advanced.md
wahidyankf: Fixed: commit ac3aff055. Added the missing bridge in both copies (`learning/code/ex-61-graphql-n1-dataloader/example.py` and its mirror `learning/advanced.md`): the module docstring, the prose paragraph above the diagram, and a new `# =>` annotation on the `QUERY_COUNT` declaration now explain that `QUERY_COUNT` tracks author lookups only (not the root articles query), which is why the runnable Output reads 3 -> 1 while the diagram's totals (which do count the root query) read 4 -> 2 — both correct, now explicitly reconciled for the reader. Verified with `python3 example.py` (output unchanged: `query count: 3` / `query count: 1`) and `pyright` (0 errors).

apps/ayokoding-www/content/en/learn/courses/api-design/overview.md
wahidyankf: Fixed: commit ac3aff055. Replaced the bare, unresolvable "topic 42" at `overview.md:53` with the same backtick-slug + `[Unverified]`-caveat pattern already used for `software-architecture` two lines above in the same paragraph — the sentence now reads "...the stable interface a `software-architecture` course would build service boundaries on top of," dropping the unsupportable catalog number entirely.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/beginner.md
wahidyankf: Fixed: commit ac3aff055. Added a `%% Color Palette: ...` comment (matching the exact house format used at `learning/overview.md:79` and `learning/capstone/overview.md:27`) to all 17 colored `graph LR`/`graph TD` Mermaid blocks across `learning/beginner.md`, `learning/intermediate.md`, and `learning/advanced.md` that were missing it — leaving the 8 `sequenceDiagram` blocks untouched, as they carry no `style`/color directives. Verified with `markdownlint-cli2` (0 errors) and `rhino-cli md mermaid validate` (0 violations, 0 warnings).

apps/ayokoding-www/content/en/learn/courses/api-design/learning/beginner.md
```diff
@@ -0,0 +1,2086 @@
...
+**Output**:
+
+```text
+GET '/getArticle?id=1' -> resource-style (noun in path)
+GET '/articles/1' -> resource-style (noun in path)
+```
+
+**Key takeaway**: The `?id=1` query-string form of `/getArticle` hides its own verb from this
+regex-based check the way a real reviewer would not -- illustrating exactly why resource-naming
+review is a human judgment call, not something a mechanical check alone can fully automate.
```
wahidyankf: **HIGH** · confidence **96** · discipline: logic (correctness) · *reproduced*

**The "Key takeaway" states a cause that is provably false, and contradicts the cycle-1-corrected annotation ~2 lines above it on the same page.**

The text claims the classifier misses `/getArticle?id=1` because "the `?id=1` query-string form ... hides its own verb from this regex-based check."

I reproduced the actual behavior against [`example.py`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-02-rpc-vs-resource-contrast/example.py#L11)'s real pattern:

```
'/getArticle?id=1'  -> None          <- no match, query string present
'/getArticle'       -> None          <- STILL no match with the query string REMOVED
'/get/Article?id=1' -> match 'get'   <- DOES match WITH a query string present
'/get_article?id=1' -> None
```

The query string is **irrelevant** in both directions: removing it changes nothing, and keeping it does not prevent a match. The real cause is that `\b(get|...)\b` requires a word boundary and `getArticle` is camelCase concatenation, so `\bget\b` never matches.

That correct explanation is **already on this page**, at [`beginner.md:84`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/apps/ayokoding-www/content/en/learn/courses/api-design/learning/beginner.md#L84): *"label is 'resource-style (noun in path)' on BOTH passes -- `\bget\b` never matches 'getArticle'"* — fixed in cycle 1. This prose was not touched by that fix, so the page now teaches two mutually exclusive mechanisms within 55 lines.

This is beginner-tier material whose whole purpose is to build a correct mental model of why the naive check fails; the wrong mechanism actively mis-teaches it.

**Fix**: restate the takeaway around the word-boundary/camelCase cause. Note the surrounding "Why it matters" paragraph is accurate and needs no change.

**Related**: this is the *same underlying Example-2 confusion* as the finding on `example.py:2-6`. Deliberately filed separately (different defect type, different fix) — but please treat Example 2's narrative as **one class** and sweep every surface, rather than only the lines cited.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-02-rpc-vs-resource-contrast/example.py
```diff
@@ -0,0 +1,31 @@
+# pyright: strict
+"""Example 2: RPC-Style vs. Resource-Style URIs. (co-05)
+
+Contrasts an RPC-flavored endpoint (`/getArticle?id=1`, a verb baked into the
+path) against the resource-oriented equivalent (`GET /articles/1`, a noun in
+the path with the verb carried by the HTTP method instead).
```
wahidyankf: **MEDIUM** · confidence **90** · discipline: docs (accuracy) · *verified* · **downgraded from raw HIGH/92**

**Example 2's top-level framing omits its actual punchline, leaving the file self-inconsistent after the cycle-1 fix.**

The docstring frames the example as *"Contrasts an RPC-flavored endpoint ... against the resource-oriented equivalent."* The cycle-1-corrected comment at [`example.py:31`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-02-rpc-vs-resource-contrast/example.py#L31) now ends with *"Output: two lines, the SAME label -- not a contrast."*

**Why I downgraded this from the raw HIGH.** The specialist argued the docstring "implies the classifier distinguishes them." I re-read it: the docstring never mentions the classifier, and describing the *setup* as contrasting two URI forms is accurate — the two constants genuinely are a contrast. Nothing here is false.

**What survives.** The example's real teaching point — that the naive regex **fails** to flag the RPC form — appears nowhere in the docstring or the cross-reference, so a reader going top-down meets "contrasts A against B" and then identical output with no bridge. Same gap at both mirrors:

- [`beginner.md:88-90`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/apps/ayokoding-www/content/en/learn/courses/api-design/learning/beginner.md#L88-L90) — verbatim copy of this docstring
- [`learning/overview.md:164`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/apps/ayokoding-www/content/en/learn/courses/api-design/learning/overview.md#L164) — *"Example 2 contrasts the resource form against its RPC-style equivalent directly"*, which promises a contrast the example deliberately does not deliver

**Fix**: add one clause to all three surfaces noting the classifier returns the same label for both, and that this failure is the point. Incomplete-not-false, hence MEDIUM.

**Related**: pairs with the HIGH finding at `beginner.md:137-139`. Sweep Example 2's narrative as one class.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/code/facade/contrast.md
```diff
@@ -0,0 +1,43 @@
...
+The identical Article data (`{"id": 2, "title": "Capstone Article"}`, created by `rest.py`'s Step 2
+write) is served two ways from this capstone: `GET /v1/articles/2` over REST returns the article's
+full, fixed shape (Example 1's own rule); `graphql_query(2, [...])` returns only the fields the
+caller's own query names (Example 58's own rule) -- both against the SAME underlying `STORE`.
+
+## What stays identical between the two facades
+
+- **The data itself.** `graphql_facade.py`'s `STORE` is populated with the exact same values
+  `rest.py` produced -- `data_matches` in that script asserts this explicitly, and the assertion
+  passes.
```
wahidyankf: **MEDIUM** · confidence **96** · discipline: docs (accuracy) · *verified*

**The prose claims an `assert` that does not exist anywhere in the file it describes.**

This bullet states: *"`data_matches` in that script asserts this explicitly, and the assertion passes."*

In [`graphql_facade.py`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/code/facade/graphql_facade.py#L30-L31), `data_matches` is computed and then only printed:

```python
data_matches = graphql_response["data"]["article"] == rest_equivalent
print(f"REST and GraphQL facade return equivalent data: {data_matches}")
```

`grep -n "assert" graphql_facade.py` returns **no matches** — there is no `assert` statement in the 35-line file. Nothing is asserted, so no assertion "passes."

The claim's *substance* is true (the values do compare equal, and the script prints `True`), which is why this is MEDIUM rather than HIGH — the described **mechanism** is wrong, not the conclusion.

The distinction matters here because this capstone's sibling script genuinely does assert — [`validate_spec.py:60`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/code/validate_spec.py#L60) uses a real `assert`, and `contrast.md` describes *that* one accurately. A learner comparing the two is told the same word means the same thing in both places when it does not.

**Boundary note (docs vs. logic).** I kept this in **docs**, not logic, unlike the Example-2 finding above: that one misstates *why code behaves as it does* (domain-behavior correctness), whereas this misstates *which construct the code uses* while its behavioral claim holds — documentation accuracy.

**Fix**: say `data_matches` is computed and printed as `True`, or add a real `assert` to the script.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/overview.md
```diff
@@ -0,0 +1,492 @@
...
+**Goal**: design and serve a versioned REST API for one resource, contract-first from an OpenAPI
+spec -- with cursor pagination, idempotent writes, a consistent error envelope, and rate limiting --
+then add a GraphQL facade over the same data and document when each contract wins.
+
+**Concepts exercised**: [x] OpenAPI contract-first (co-09) [x] versioning + backward-compatible
+change (co-13, co-14) [x] cursor pagination (co-17) [x] idempotency keys (co-18) [x] error envelope
+(co-08, co-30) [x] rate limiting (429 + headers) (co-19, co-20).
+
+```mermaid
+%% Color Palette: Blue #0173B2, Orange #DE8F05, Teal #029E73, Purple #CC78BC
```
wahidyankf, line 27: **LOW** · confidence **97** · discipline: governance (doc convention) · *verified, blast radius audited* · **downgraded from raw MEDIUM/95**

**The palette comment declares a 4th color the diagram never uses.**

The comment lists `Blue #0173B2, Orange #DE8F05, Teal #029E73, Purple #CC78BC`, but the block defines and applies only three `classDef`s — blue, teal, purple. `#DE8F05` appears nowhere else in the block. Consistent with the specialist's copy-paste-from-a-4-color-template read.

**Blast radius — audited, and it is contained.** This mattered more than the line itself: cycle 1's fix added palette comments to 17 blocks, explicitly *"matching the exact house format used at ... `learning/capstone/overview.md:27`"* — i.e. copied from this very comment. I therefore programmatically diffed declared-vs-used hex codes across **all 19 palette comments** in the course:

```
palette comments found: 19
blocks with phantom (declared-but-unused) colors: 1
  learning/capstone/overview.md:27  PHANTOM=['#DE8F05']
```

**The phantom color did not propagate.** The cycle-1 fixer correctly wrote per-diagram palettes. This is a single-site defect.

**Why LOW rather than the raw MEDIUM**, on three grounds:

1. The palette comment is explicitly **"(RECOMMENDED)"** in [diagrams.md §Documentation Requirements item 2](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/repo-governance/conventions/formatting/diagrams.md#L742-L747), not required; the REQUIRED half (accessible hex codes) is satisfied.
2. Zero accessibility impact — no rendered output changes.
3. Cycle 1 already calibrated this exact axis at LOW; re-filing the same axis at MEDIUM would be inconsistent.

LOW is optional per the [Criticality Levels Convention](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/repo-governance/development/quality/criticality-levels.md#L263-L278) — but this one is a one-word deletion, so it is worth taking.

**Adjacent, for the same edit**: node `B` (`Step 2`) carries no `:::class` at all while `A`/`C`/`D` all do — likely unintended, since Step 2 is not obviously the odd one out.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/beginner.md
```diff
@@ -0,0 +1,2086 @@
...
+```mermaid
+%% Color Palette: Blue #0173B2, Orange #DE8F05
+graph LR
+    A[Create] -->|maps to| B[POST]
+    C[Read] -->|maps to| D[GET]
+    E["Update (replace)"] -->|maps to| F[PUT]
+    G["Update (partial)"] -->|maps to| H[PATCH]
+    I[Delete] -->|maps to| J[DELETE]
+
+    style A fill:#0173B2,stroke:#000,color:#fff
```
wahidyankf: **LOW** · confidence **95** · discipline: governance (doc convention) · *verified, numbers re-grounded* · **promoted from a routing note**

**17 colored Mermaid blocks introduced by this PR set colors with inline `style`, against the house rule to use `classDef`.**

Anchored here as representative of the class. [diagrams.md §Key Implementation Points](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/repo-governance/conventions/formatting/diagrams.md#L761) states plainly: *"Define colors in `classDef` sections, not inline."*

**I re-grounded the specialist's numbers before filing — both were wrong, in opposite directions.**

The claim was "all 25 new diagrams" against a "189/195 (96.9%)" sibling baseline. Measured at this SHA:

```
api-design: 27 mermaid blocks = 17 inline-style + 2 classDef + 8 sequenceDiagram (uncolored)
siblings:   872 classDef blocks vs 1 inline-style block  ->  99.9% classDef
```

So the count is **17, not 25** (the 8 `sequenceDiagram` blocks set no colors, and this course's own `learning/overview.md` and `capstone/overview.md` already use `classDef`) — but the deviation is *more* unusual than claimed: exactly **one** colored block in every sibling course uses inline `style`. The 17 blocks here are near-unique repo-wide.

**Why LOW, not higher:**

- **No accessibility defect.** The inline styles use the verified accessible palette with black strokes and correct text contrast, e.g. `style A fill:#0173B2,stroke:#000,color:#fff`. The REQUIRED accessibility half of the convention is satisfied; only the mechanism deviates.
- **Not tooling-enforced.** [`main-ci.yml:124`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/.github/workflows/main-ci.yml#L124) runs `md mermaid validate` with `--exclude apps/ayokoding-www/content`, so CI cannot catch this. Not a CI blocker.
- Consistent with cycle 1's LOW calibration for the non-accessibility half of this convention.

**Not a re-litigation of the resolved cycle-1 thread.** That thread covered a *different axis* — the presence of the `%% Color Palette` comment — and was fixed. The `classDef`-vs-inline axis was never raised, and no human has dismissed it.

**Fix (optional per LOW)**: convert the 17 blocks to `classDef` + `:::class`, which also removes the per-node repetition (this block alone repeats 10 `style` lines for 2 colors). Reasonable to decline deliberately given zero accessibility impact — but note this ships 17 of the repo's 18 inline-style blocks.

wahidyankf (commented):
**Risk tier**: full
**Diff coverage**: full diff reviewed in one pass (no generated-file exclusion)
**Prior-cycle human dismissals respected**: none this cycle (all 5 cycle-1 threads resolved by fix, zero human dismissals)

---

## Cycle 2 of 3 — consolidated review

**This review is blocking despite its `COMMENT` state** (`REQUEST_CHANGES` is unavailable on one's own PR). Blocking status is carried by the severity labels below.

**5 findings** from 8 specialists: 1 HIGH, 2 MEDIUM, 2 LOW. Five specialists (architecture, security, integrity, performance, instruction) returned zero findings.

### Coordination actions taken

| Action | Detail |
| --- | --- |
| Kept separate (not merged) | The logic finding and docs finding 1 both concern Example 2, but anchor different `file:line`s and are **different defect types** (a provably false statement vs. incomplete framing). Merging risked the same partial fix that produced this cycle's findings. Cross-linked instead, with a class-level sweep instruction. |
| Downgraded HIGH → MEDIUM | Docs finding 1. The docstring is *incomplete*, not false; the specialist's "implies the classifier distinguishes them" is an over-read — the docstring never mentions the classifier. |
| Downgraded MEDIUM → LOW | Governance finding. Verified blast radius is contained (1 of 19 palette comments), the palette comment is **RECOMMENDED** not required, and the path is CI-excluded. Matches cycle 1's own LOW calibration on this axis. |
| Promoted routing note → LOW finding | The `classDef`-vs-inline-`style` note clears the bar: real, PR-introduced, verified. Its numbers were **re-grounded** (see finding 5) — the deviation is *more* unusual than the specialist claimed, but the diagram count was overstated. |
| Confidence re-scored after tool-verify | 4 of 5 findings mechanically verified; scores raised (95→96, 90→96, 95→97) or lowered (92→90) accordingly. |
| Anchor corrected | Governance finding cited `capstone/overview.md:26`; line 26 is the ` ```mermaid ` fence — the palette comment is line **27**. |

### Verification performed this cycle

- **Reproduced** the Example 2 regex behavior directly (`python3`), which *falsifies* the "Key takeaway" causal claim — see finding 1.
- **Audited all 19 palette comments** in the course programmatically: the cycle-1 fix did **not** propagate the phantom color. Contained to one block.
- **Measured the sibling-course baseline**: 872 of 873 colored Mermaid blocks (99.9%) use `classDef`.
- **Ran `rhino-cli md mermaid validate`** over the course, then confirmed against [`main-ci.yml:124`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/.github/workflows/main-ci.yml#L124) that `apps/ayokoding-www/content` is **`--exclude`d** in CI — so none of the diagram findings are CI-blocking. CI is green (18 passed, 0 failed).
- **Untrusted-input scan**: PR body and comments contain no prompt-injection attempt and no fabricated structural tags.

### Non-blocking coordinator note (not a scored finding)

While tool-verifying finding 4, `rhino-cli md mermaid validate` reported one violation on the same block: `label_too_long` — node `A`'s label is 31 chars (max 30) at [`capstone/overview.md:29`](https://github.com/wahidyankf/ose-public/blob/ac3aff0557f4199e0b43e2245852b45c5b92a916/apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/overview.md#L29). **Not CI-blocking** (path excluded above) and **no specialist raised it**, so it is deliberately not filed as a finding — noted only because the fixer will already be editing this block. The PR body's claim of "`rhino-cli md mermaid validate` clean (25 diagrams, 0 violations)" is inaccurate on both counts (27 blocks scanned, 1 violation), though harmlessly so given the CI exclusion.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/beginner.md
wahidyankf: Fixed: rewrote the Key takeaway in `learning/beginner.md` to state the real cause -- `\bget\b` requires a word boundary on both sides of "get", and the camelCase-glued `getArticle` never supplies one, so the classifier misses it regardless of the `?id=1` query string. Also swept the related "Contrasts X against Y" framing (which implied real differentiation the code doesn't produce) across `example.py`'s docstring, `beginner.md`'s prose intro + embedded code mirror, and `learning/overview.md:164`'s cross-reference, so every surface now describes the same mislabeling behavior consistently. Commit d80683850.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/code/ex-02-rpc-vs-resource-contrast/example.py
wahidyankf: Fixed: swept every cited surface (and mirrors) of the Example-2 "Contrasts X against Y" narrative -- `example.py`'s module docstring, `beginner.md`'s prose intro and its embedded code-block mirror, `beginner.md`'s Key-takeaway causal claim, and `learning/overview.md:164`'s cross-reference -- to state what the code actually demonstrates: the classifier mislabels BOTH the RPC-style and resource-style forms identically, because `\bget\b` needs a word boundary the camelCase-glued `getArticle` never provides. Commit d80683850.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/overview.md
wahidyankf: Fixed: added the missing 4th `classDef orange` (`#DE8F05`) and applied it to Step 2's node in `learning/capstone/overview.md`, so the diagram now uses all 4 colors listed in its own palette comment rather than dropping the phantom one. Chose "add the 4th classDef" over "drop the comment" because the diagram already has 4 distinct sequential steps that read more clearly with 4 distinct colors than with Step 2 left uncolored. While in there, also fixed the pre-existing `label_too_long` violation on this same diagram's Step-1 node (one `<br/>` line was 31 chars, over rhino-cli's 30-char max) -- this was the sole violation across all 27 Mermaid diagrams in the course, confirmed via `rhino-cli md mermaid validate` (now 0 violations), and the PR body's stale "25 diagrams, 0 violations" claim is corrected to the verified 27. Commit d80683850.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/code/facade/contrast.md
wahidyankf: Fixed: reworded the `data_matches` bullet in `learning/capstone/code/facade/contrast.md` to neutral "is `True`" phrasing (`graphql_facade.py` has zero `assert` statements), mirroring how the capstone already describes `rest.py`'s equivalent `did_not_double_apply` boolean in `capstone/overview.md`. Commit d80683850.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/beginner.md
wahidyankf: Fixed: converted all 17 colored Mermaid diagrams across `learning/beginner.md` (5), `learning/intermediate.md` (4), and `learning/advanced.md` (8) from inline `style <id> fill:...` to `classDef` + `class` blocks, matching the house convention in `repo-governance/conventions/formatting/diagrams.md`. Colors and fill/text-contrast pairings are preserved exactly (only the mechanism changed, e.g. blue `#0173B2`/white text, orange `#DE8F05`/black text, etc.) -- no visual regression. Verified with `rhino-cli md mermaid validate` (0 violations across all 27 diagrams in the course) and `markdownlint-cli2`/`prettier --check` (clean). Commit d80683850.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/advanced.md
```diff
@@ -0,0 +1,2052 @@
...
+would let an attacker exhaust the rate limit's bookkeeping with replay traffic that should be
+essentially free.
+
+---
+
+### Example 75: GraphQL Partial Errors vs REST problem+json
+
+_ex-75 &middot; exercises co-30, co-08_
+
+REST signals failure via the STATUS CODE (Example 6's `problem+json`, a non-2xx status); GraphQL
```
wahidyankf: **MEDIUM** | confidence 99 | docs (re-categorized from logic) | **Fix in this PR**

### Example 75 cites the wrong example number for the problem+json envelope

Example 75's prose and docstring attribute the `application/problem+json` error envelope to
**Example 6** three times. Verified against this head SHA:

- `learning/beginner.md:368` -> `### Example 6: 202 Accepted for a Long-Running Operation`
- `learning/beginner.md:814` -> `### Example 12: The application/problem+json Error Envelope`

So the citation should read **Example 12**. A learner following the pointer lands on async 202
semantics instead of the error envelope the sentence is actually contrasting against.

**All five sites requiring the change** (the `.md` at 1502/1513 is the rendered mirror of the `.py`
at 4/15 - both copies must be edited or they desync):

| File | Line |
|---|---|
| `learning/advanced.md` | 1492 (prose) |
| `learning/advanced.md` | 1502 (embedded docstring) |
| `learning/advanced.md` | 1513 (embedded inline comment) |
| `learning/code/ex-75-error-envelope-graphql/example.py` | 4 (docstring) |
| `learning/code/ex-75-error-envelope-graphql/example.py` | 15 (inline comment) |

**Do not blanket-substitute.** Two other `Example 6` references in this course are **correct** and
must be left untouched - a naive `sed 's/Example 6/Example 12/'` would corrupt both:

- `drilling/overview.md:117` - `` `202 Accepted` (Example 6) `` (correct)
- `learning/overview.md:506` - the Example 6 TOC link to the 202 Accepted anchor (correct)

I swept every `Example 6` occurrence across the course; the five above are the complete defect set.

*- generated by AI (pr-review-synthesis-maker)*

apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/overview.md
```diff
@@ -0,0 +1,493 @@
...
+graph LR
+    A["Step 1<br/>openapi.yaml +<br/>validate_spec.py<br/>co-09, co-14"]:::blue --> B["Step 2<br/>rest.py<br/>co-13, co-17, co-18"]:::orange
+    B --> C["Step 3<br/>limits.py<br/>co-19, co-20"]:::teal
+    C --> D["Step 4<br/>facade/<br/>co-24, co-27"]:::purple
+
+    classDef blue fill:#0173B2,stroke:#000000,color:#FFFFFF,stroke-width:2px
+    classDef orange fill:#DE8F05,stroke:#000000,color:#FFFFFF,stroke-width:2px
```
wahidyankf, line 34: **LOW as applied to this PR** | confidence 95 | governance | **Do NOT fix in this PR**

### Mermaid classDef white-on-mid-tone text fails WCAG AA 4.5:1 - real defect, but this PR is not its source

The governance specialist raised this at `HIGH` across 35 `classDef` lines in 12 touched files. The
underlying contrast math is correct - I recomputed every ratio independently using the WCAG
relative-luminance formula:

| Fill | vs `#FFFFFF` | vs `#000000` | Compliant text |
|---|---|---|---|
| blue `#0173B2` | **5.13:1** | 4.10 | white (passes) |
| gray `#808080` | 3.95 | **5.32:1** | black |
| teal `#029E73` | 3.42 | **6.14:1** | black |
| purple `#CC78BC` | 3.01 | **6.98:1** | black |
| brown `#CA9161` | 2.72 | **7.73:1** | black |
| orange `#DE8F05` | 2.61 | **8.04:1** | black |

Blue is the *only* palette color for which white text clears 4.5:1. Every other palette color fails
with white text and passes comfortably with black.

**Why this is nonetheless not a defect attributable to this PR** - three independent reasons:

**1. This PR follows the convention's own normative example, verbatim.**
[`diagrams.md:684-697`](../blob/d80683850a1c999db429f576656b640ce552c813/repo-governance/conventions/formatting/diagrams.md#L684-L697)
prescribes, under the literal heading **"Good Example (accessible):"**:

```
classDef orange fill:#DE8F05,stroke:#000000,color:#FFFFFF,stroke-width:2px
classDef teal  fill:#029E73,stroke:#000000,color:#FFFFFF,stroke-width:2px
```

The line this comment is anchored to is that string character-for-character, `stroke-width` included.
`git blame` confirms it was written by cycle 2's own fixer commit (`d80683850a`) - the fixer was
correctly following the documented canonical example. Holding this PR to a numeric bar that the same
document's normative example violates is a convention-contradicted finding.

**2. It is a repo-wide class, and this PR is ~1% of it.** The identical pattern appears on **2,802
lines across 357 markdown files**, including `diagrams.md` itself (also at lines 1388-1389, 1594,
1603, 1658). Fixing only this course's 35 lines would make it the single visual outlier among 357
files while leaving the authority doc still teaching the failing pattern - so the next authored
course reintroduces it immediately. That is symptom-fixing, which
[Root Cause Orientation](../blob/d80683850a1c999db429f576656b640ce552c813/AGENTS.md) forbids.

**3. No tooling enforces the bar.** `rhino-cli md mermaid validate` has no contrast check at all
(confirmed: zero matches for "contrast" anywhere under `apps/rhino-cli/src`). That absence is why
2,802 violations accumulated unnoticed, and it is part of the correct class-level fix.

**Recommended remediation, routed OUT of this PR** - a dedicated governance plan that, in order:
(a) fixes `diagrams.md` to prescribe `color:#000000` for every non-blue palette fill and corrects its
"Good Example" block; (b) adds a contrast assertion to `rhino-cli md mermaid validate`; (c) sweeps
the 357 files under the now-passing validator.

**Final-cycle risk note:** rewriting 35 `classDef` lines across 12 files in cycle 3 of 3, with no
cycle 4 to catch a regression and no validator coverage for the property being changed, is a
risk-increasing change for zero class-level benefit.

*- generated by AI (pr-review-synthesis-maker)*

wahidyankf (commented):
**Risk tier**: full
**Security-sensitive-path override applied**: no (tier already `full` on size: 111 files / 12,460 added lines)
**Diff coverage**: full diff reviewed, no generated-file exclusion; distributed across per-domain file slices (111 files, ~12.5k added lines exceeds a single specialist context window)
**Prior-cycle human dismissals respected**: none this cycle (10/10 cycle-1+cycle-2 threads resolved, zero human dismissals)

---

## Cycle 3 of 3 (FINAL) - consolidated review

**This review is NON-BLOCKING.** No `CRITICAL` or `HIGH` finding survived synthesis. Six of eight
specialists returned no findings. The docs specialist explicitly re-verified that every cycle-1 and
cycle-2 fix landed correctly and consistently, and a fresh whole-course sweep found nothing new.

Two raw findings reached synthesis. Both survive, at reduced scope after tool-verification:

| # | Discipline | Severity | Confidence | Disposition |
|---|-----------|----------|-----------|-------------|
| 1 | docs (re-categorized from logic) | MEDIUM | 99 | Fix in this PR - mechanical, 5 sites |
| 2 | governance | LOW *as applied to this PR* | 95 | **Do not fix in this PR** - root cause is upstream |

### Coordination actions taken

- **Re-categorized** the logic specialist's routing note to **docs**. Boundary call confirmed: the
  emitted code is correct in every respect; only a prose cross-reference number is wrong. That is
  documentation accuracy, not domain behavior.
- **Re-scoped** the governance finding's severity from `HIGH` to `LOW` **as applied to this PR**, on
  convention-contradiction and root-cause grounds. The class-level defect remains genuinely `HIGH`
  and is routed out of this PR - detail in finding 2. This is a scope re-categorization, not a
  dismissal of an accessibility defect.
- **Tool-verified** both findings against `d80683850a1c999db429f576656b640ce552c813`: recomputed
  every WCAG relative-luminance contrast ratio independently, enumerated the full repo-wide
  occurrence set, and `git blame`-d the specific line the governance specialist flagged as a
  cycle-2 fixer regression.

*- generated by AI (pr-review-synthesis-maker)*

apps/ayokoding-www/content/en/learn/courses/api-design/learning/advanced.md
wahidyankf: Fixed: corrected the mis-citation from "Example 6" to "Example 12" at all 5 sites tied to the `application/problem+json` error envelope — 3 in `learning/advanced.md` (prose line 1492, docstring line 1502, inline comment line 1513) and 2 in the mirrored `learning/code/ex-75-error-envelope-graphql/example.py` (docstring line 4, inline comment line 15). Verified Example 12 is "The application/problem+json Error Envelope" in `learning/beginner.md`, while Example 6 is "202 Accepted for a Long-Running Operation" — confirming the original citations were wrong. Left the two legitimate Example 6 references (`drilling/overview.md:117`, `learning/overview.md:506`) untouched, and did not touch any Mermaid classDef/color per the coordinator's explicit out-of-scope note in the sibling thread. Ran `python3 -m py_compile` + `python3` execution + `pyright` (0 errors) on `example.py`, and `markdownlint-cli2` (0 errors) on `advanced.md`. Committed as 2817dcdb and pushed to the PR branch.

apps/ayokoding-www/content/en/learn/courses/api-design/learning/capstone/overview.md
wahidyankf: **Defer — not fixed in this PR** | governance

Confirmed and standing by the coordinator's (`pr-review-synthesis-maker`) prior triage on this
thread: this finding is real (the white-on-mid-tone `classDef` fills genuinely fail WCAG AA 4.5:1
for five of six palette colors), but it is not this PR's defect to fix, for three reasons already on
record here:

1. **The convention doc contradicts itself, and this PR followed the half that's wrong.**
   `repo-governance/conventions/formatting/diagrams.md` states the numeric bar — `**4.5:1**` — at the
   "Contrast Requirements (WCAG AA)" section (line 664 header, ratio stated at line 666) and again in
   the pre-flight checklist at line 1200 (`Contrast ratios meet WCAG AA standards (4.5:1 for text)`).
   But the very same document's own **"Good Example (accessible)"** block, lines 684-697, prescribes
   `classDef orange fill:#DE8F05,...,color:#FFFFFF` and `classDef teal fill:#029E73,...,color:#FFFFFF`
   — both of which fail the 4.5:1 bar the document states 20 lines earlier (orange/white: 2.61:1,
   teal/white: 3.42:1). This PR's flagged lines are that canonical example, applied verbatim. Holding
   this PR to a bar the authority document's own normative example violates is not a defect
   attributable to this PR's authors.

2. **Scale**: the identical pattern (mid-tone `classDef` fill + `color:#FFFFFF`) appears on 2,802
   lines across 357 markdown files repo-wide, including in `diagrams.md` itself at additional
   locations beyond the example block (lines 1388-1389, 1594, 1603, 1658). Rewriting only this
   course's ~35 lines would make it the sole visual outlier among 357 files while the authority doc
   still teaches the failing pattern — the next authored course would reintroduce it immediately.
   That's symptom-fixing, not root-cause fixing.

3. **No tooling enforces the bar today**: `rhino-cli md mermaid validate` has zero contrast-checking
   logic, which is why 2,802 violations accumulated unnoticed in the first place.

**Correct remediation path, routed out of this PR**: a dedicated governance plan that (a) fixes
`diagrams.md` itself — correcting the "Good Example" block and prescribing `color:#000000` for every
non-blue palette fill — before anything else, (b) adds a contrast assertion to
`rhino-cli md mermaid validate`, then (c) sweeps the 357-file repo-wide corpus under the now-passing
validator. Deferring, not resolving as won't-fix — the finding stands as a legitimate repo-wide
governance gap to be tracked and addressed in that follow-up plan.

PR #118, opened 2026-07-29
feat(ayokoding-www): phase 8 — harness/class filters for AI benchmark

wahidyankf:
## Summary
- Adds URL-driven harness/class filters (AC-18, AC-22..AC-28) that narrow the capability chart, price chart, and data table together, sharing one filtered dataset.
- New responsive `benchmark-filters.tsx` shell component: collapsed `<details>` disclosure below `md`, inline wrapping bar at `md`, single-row bar with result count at `lg`, all built on one generic `FilterSelect`.
- Price chart now accepts an active `harness` prop and switches rate sets via `rateForHarness`.
- Explicit localized empty state when a filter combination matches no model; neither chart renders an empty plot area in that case.
- Page remains unlinked (Phase 5 invariant preserved).

Part of `plans/in-progress/ayokoding-www-tools-ai-benchmark/` — Phase 8 delivery unit.

## Test plan
- [x] `npx nx run ayokoding-www:test:unit` exits 0 (3059 passed, 6 pre-existing skips)
- [x] `npx nx run ayokoding-www:specs:behavior:coverage` exits 0 (327 scenarios, 1177 steps, all covered)
- [x] `npx nx run ayokoding-www-fe-e2e:test:e2e` exits 0 for all AI-benchmark scenarios (pre-existing unrelated flake in course-rehome-redirects on firefox/webkit only, untouched by this diff)
- [x] Page still unlinked — `grep -c "tools/ai-benchmark"` on tools index + footer both print 0
- [x] `npx nx affected -t typecheck lint` exits 0


wahidyankf (commented):
**Risk tier**: full
**Prior-cycle human dismissals respected**: none this cycle (cycle 1 of up to 3)

---

## Cycle 1 — consolidated review

**This review is BLOCKING** — it carries one `CRITICAL` finding. GitHub rejects `REQUEST_CHANGES` on one's own pull request under the current `gh` identity posture, so blocking status is carried in the severity labels on each thread, not in the review state.

**3 findings posted** (1 CRITICAL, 1 MEDIUM, 1 LOW) out of 4 raw findings raised across the eight specialists.

Risk tier is `full` on size (770 insertions across 8 files, > 100 lines); no security-sensitive path is touched, so no override was applied.

### Coordination record

| Raw finding(s) | Disposition |
| --- | --- |
| Architecture F1 + Logic F1 — anchor collapse | **Deduped into one thread**; re-categorized architecture → **correctness**; **reproduced** at the pinned SHA. Posted `CRITICAL`. |
| Architecture F2 — missing `{ scroll: false }` | Verified against the cited precedent. Posted `MEDIUM`. |
| Integrity Obs-2 + Docs routing note — AC-23/AC-28 tag mismatch | **Deduped into one thread**; re-categorized from *non-postable pre-existing drift* → **postable LOW** (tool-verify contradicted the "pre-existing" claim — see the thread). |
| Integrity F1 — weak AC-22 assertion | **Dropped** by the reasonableness filter (below). |
| Governance routing note — `useSearchParams()` / Suspense | **Cleared** by tool-verify (below). No finding. |
| Security, Performance, Instruction-decay (all clean); Integrity Obs-1 | No action. |

### Dropped — Integrity F1, weak AC-22 test (LOW, raw confidence 82)

The claim was that AC-22 "adds no discriminating power" because the pre-PR baseline already passed the full unfiltered dataset to `ModelTable`. Tool-verify partly contradicts this: the assertion was indeed vacuous *before* this PR, but *after* it the same scenario exercises the genuinely new `decodeState → filterModels → ModelTable` path and would catch a `decodeState` that wrongly defaulted to a non-empty filter. Weak, but not worthless — and it sits at the confidence floor with an advisory-only fix. Dropped as a nitpick rather than passed through to the fixer.

### Cleared — Governance's routed `useSearchParams()` / Suspense question

Not a finding. `apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/page.tsx` already wraps `<BenchmarkContent />` in `<Suspense>` (pre-existing, unchanged by this PR), and `next.config` sets `output: "standalone"` — not a static export — so there is no CSR-bailout or build-time failure mode here. Governance was right to route rather than judge it; the answer is no.

### Note for the fixer

Finding 1 is a **bug fix**, so the [Regression Test Mandate](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/development/quality/regression-test-mandate.md) applies — a reproducing test must land in the same PR, and it must be pinned to a harness that excludes both anchors. The existing AC-23/AC-24/AC-25 scenarios structurally cannot catch it.

---

apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx
```diff
@@ -1,15 +1,38 @@
...
+  const pathname = usePathname();
+
+  // The URL is the single source of truth for the active filters (F-3..F-9, Phase 4). One
+  // `filterModels` call narrows the roster over BOTH axes at once (DD-11) — the resulting
+  // `filteredDataset` is the ONE dataset every consumer below reads, so the capability chart, the
+  // price chart, and the data table always agree, and each chart's own `computeGroups` call
+  // re-scores/re-bands relative to whatever survives the filter, not the full roster.
+  const filterState: FilterState = decodeState(searchParams);
+  const filteredModels = filterModels(dataset, filterState);
+  const filteredDataset: Dataset = { ...dataset, models: filteredModels };
```
Comment (wahidyankf) [inline apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx:31]: ## CRITICAL — a harness filter silently re-bands every rated model when the filtered set excludes both anchors

**Severity**: `CRITICAL` · **Confidence**: 96 · **Discipline**: correctness (re-categorized from architecture — reasoning below) · **Deduped from**: `pr-review-architecture-maker` F1 + `pr-review-logic-maker` F1, which reproduced this independently of each other

### What is wrong

`filteredDataset` (line 28) is a `Dataset` whose `models` have been narrowed, and it is handed to `<CapabilityChart>` and `<PriceChart>` at lines 67-68. Each of those calls `computeGroups(dataset)` internally, and `computeGroups` → `anchorIndices()` (`apps/ayokoding-www/src/features/ai-benchmark/core/bands.ts:99-106`) looks the two anchors up **by fixed id inside whatever dataset it is handed**:

```ts
const opusModel = dataset.models.find((m) => m.id === OPUS_ANCHOR_ID);
const sonnetModel = dataset.models.find((m) => m.id === SONNET_ANCHOR_ID);
```

Both anchors declare `harnesses: ["claude-code", "cursor", "opencode-zen"]` (`core/data/models.ts:229` and `:267`). So any harness filter selecting `codex-cli` or `opencode-go` removes **both** anchors, leaves `anchorIndices.opus` and `.sonnet` both `undefined`, and `assignBand` (`core/bands.ts:79-85`) falls through every threshold branch to `light`.

This PR is the first thing to violate the invariant that `core/filter.ts:4-6` states in its own header:

> A model's band is a property of the FULL roster (the composite index is roster-relative per DD-5a), so the class filter is computed over the whole dataset, not the already-filtered subset.

`filterModels` still honours it — it calls `computeGroups(dataset)` on the full roster. The two charts, newly fed a filtered `Dataset` by this diff, do not.

### Reproduction

Run against the real roster at the pinned head SHA `5139d18`, computing each model's band on the full roster and then re-computing it on exactly what `benchmark-content.tsx` now hands the charts:

```
harness=claude-code   n=6   anchors[opus=true  sonnet=true ]  drifted=0
harness=codex-cli     n=7   anchors[opus=false sonnet=false]  drifted=3
    gpt-5.6-sol:      opus   -> light
    gpt-5.6-terra:    sonnet -> light
    gpt-5.6-luna:     sonnet -> light
harness=cursor        n=23  anchors[opus=true  sonnet=true ]  drifted=0
harness=opencode-go   n=15  anchors[opus=false sonnet=false]  drifted=4
    kimi-k3:          sonnet -> light
    deepseek-v4-pro:  sonnet -> light
    grok-4.5:         sonnet -> light
    glm-5.2:          sonnet -> light
harness=opencode-zen  n=35  anchors[opus=true  sonnet=true ]  drifted=0
```

**Correcting both raw findings on one point**: neither harness collapses *all* of its models. Only models that were `opus` or `sonnet` on the full roster collapse — 3 of 7 for `codex-cli`, 4 of 15 for `opencode-go`; already-`light` and `unrated` models are unaffected. The defect is real and the mechanism is exactly as both specialists described it, but the blast radius is the rated models specifically.

### Worst case — the page contradicts itself

`?harness=codex-cli&class=opus`:

```
filterModels selects (full-roster band = opus): [ 'gpt-5.6-sol' ]
capability chart renders that same model under band "light"
```

The reader explicitly asks for the opus class, and gets one model displayed beneath a `light` band header while the data table on the same page lists its class as opus. One page, one model, two different answers.

### Why this is correctness and not architecture

Applying the [boundary tie-breaker](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/development/quality/pr-review-disciplines.md#the-boundary-tie-breaker-rule), which this coordinator owns for exactly this boundary:

1. **Documented + mechanically-checkable rule → governance?** No. No `repo-governance/` convention states this; the invariant lives in a code comment and in the plan's tech-docs.md (DD-5a/DD-20a). `pr-review-governance-maker` was correct to return clean.
2. **New tradeoff judgment → architecture?** No. Answering this requires no new structural decision — the decision is already made and written down, and the fix is to honour it.
3. **Falls through to "does it satisfy domain intent?" → correctness.** It does not. Filed as correctness, and `pr-review-logic-maker`'s `CRITICAL` is the surviving severity assessment.

### Suggested fix

Derive the thresholds from the full roster always; let the filter govern membership and display only. Either:

- give `computeGroups` an explicit anchor source — e.g. `computeGroups(displayDataset, { anchorsFrom: fullDataset })`; or
- compute `BandGroups` once over the full roster in `benchmark-content.tsx` and have both charts consume a pre-scored, pre-filtered `ModelScore[]` rather than re-deriving bands from a `Dataset`.

The second is structurally cleaner. The root-cause enabler is that one `Dataset` type is doing duty as both *the full roster (the statistical population)* and *the subset to display*, so any component that re-derives roster-relative statistics from it is silently wrong under a filter. Note this is a second-order problem too: `computeRosterMaxes` is also roster-relative, so composite indices themselves shift under a filter even when the anchors survive.

### Regression test is mandatory

A reproducing test must land in this PR and must be pinned to `harness=codex-cli` or `harness=opencode-go`. The existing scenarios cannot catch this — AC-23/AC-25 exercise only `cursor` (which retains both anchors), and AC-28's `opencode-go` case asserts the empty state, not band placement. Worth noting: `grok-4.5`, one of the four drifting models, is exactly the fixture this PR's own AC-18 e2e step uses — but that step asserts price only, so the bug ships with every test green.

---

apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx
```diff
@@ -1,15 +1,38 @@
...
+  const isEmpty = filteredModels.length === 0;
+
+  function handleFilterChange(next: FilterState) {
+    const qs = encodeState(next).toString();
+    router.push(qs ? `${pathname}?${qs}` : pathname);
```
Comment (wahidyankf) [inline apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx]: ## MEDIUM — `router.push` omits `{ scroll: false }`, so every filter change yanks the reader to the top of the page

**Severity**: `MEDIUM` · **Confidence**: 92 · **Discipline**: architecture (pattern consistency) · **Source**: `pr-review-architecture-maker` F2

This is the only `router.push` on this page, and the only in-page filter-state navigation in this app that omits the option.

The sibling tool that this plan's own `tech-docs.md` cites as its FCIS precedent passes it on all three of its navigations, with a comment explicitly calling it mandatory — `apps/ayokoding-www/src/app/[locale]/tools/cost-of-living-calculator/calculator-content.tsx:63-70`:

> `scroll: false` is mandatory: this is in-page filter/view state, not a page change, so the default Next.js navigation behaviour (scroll to top of document) would yank the reader [...] scroll position is the fix for "updating a filter jumps me back to the top".

Verified this is a genuine inconsistency and not a misread precedent: the only other `router.push` in the app, `src/features/search/shell/search-dialog.tsx:85`, correctly omits the option — that one really is a page change. This one is not.

The impact is concrete on this specific page. The filter bar renders above both charts, so a reader who has scrolled down to the price chart and changes the harness selector is returned to the page header on every single change — and the mobile `<details>` variant makes that worse, since the disclosure sits above the charts too.

**Suggested fix** — `apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx:33`:

```ts
router.push(qs ? `${pathname}?${qs}` : pathname, { scroll: false });
```

---

specs/apps/ayokoding/behavior/ayokoding-www/gherkin/tools/ai-benchmark.feature
```diff
@@ -224,3 +224,69 @@ Feature: AI model benchmark tool
...
+
+  # AC-22
+  @unit @e2e
+  Scenario: The page with no query parameters shows the whole roster
+    Given the URL carries no query parameters
+    When the page renders
+    Then every roster model is shown in the data table
+
+  # AC-23
+  @unit
```
wahidyankf, line 240: ## LOW — AC-23 and AC-28 ship `@unit`-only, dropping the `@e2e` layer prd.md declares for them

**Severity**: `LOW` · **Confidence**: 90 · **Discipline**: test integrity / spec fidelity · **Deduped from**: `pr-review-integrity-maker` Obs-2 + `pr-review-docs-maker` routing note, surfaced independently by both

Tag parity for the eight scenarios this PR adds, against `plans/in-progress/ayokoding-www-tools-ai-benchmark/prd.md` §"Filters and URL state":

| AC | prd.md | shipped `.feature` | |
| --- | --- | --- | --- |
| AC-22 | `@unit @e2e` | `@unit @e2e` | ok |
| **AC-23** | `@unit @e2e` | `@unit` | **`@e2e` dropped** (line 240) |
| AC-24 | `@unit` | `@unit` | ok |
| AC-25 | `@unit` | `@unit` | ok |
| AC-18 | `@unit @e2e` | `@unit @e2e` | ok |
| AC-26 | `@unit` | `@unit` | ok |
| **AC-28** | `@unit @e2e` | `@unit` | **`@e2e` dropped** (line 280) |
| AC-27 | `@e2e` | `@unit @e2e` | `@unit` added — extra coverage, benign |

The e2e step file confirms this is a real coverage gap rather than a tag typo: this PR adds e2e bindings for AC-22, AC-18 and AC-27 only. Nothing binds AC-23 or AC-28 at the e2e layer.

### Correcting both raw findings

Both specialists routed this as *pre-existing* drift and declined to raise it. That is half right, and the wrong half is what makes it postable.

- The **root cause is** pre-existing: `delivery.md`'s N-3 and N-13 both specify unit-layer binding only (`command: npx nx run ayokoding-www:test:unit`), contradicting prd.md, and that contradiction predates this PR.
- The **manifestation is new in this diff**: these `.feature` scenarios did not exist before this PR. They are added by step N-0 — *"append AC-18, AC-22 … AC-28 to `ai-benchmark.feature`"* — which is squarely inside this PR's declared scope. The `@unit`-only tags are introduced here, not inherited. N-0's acceptance check (`specs:structure-validation`) evidently does not verify tag parity against prd.md, so nothing caught it.

### Suggested resolution — either direction is acceptable

This is explicitly **not** a demand for more scope. Either add the two missing e2e bindings, or amend prd.md to drop `@e2e` from AC-23/AC-28 so it matches the delivery plan that was actually executed. The second is a plan-document decision and a perfectly reasonable reasoned-reject for `pr-review-fixer` to route to the maintainer rather than resolve in code. What should not happen is leaving prd.md and the shipped spec silently disagreeing about the required test layers.

---

apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx
Comment (wahidyankf) [reply apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx:31]: Fixed: added an explicit `fullDataset` parameter to `computeGroups`/`computeScoreViews` in `apps/ayokoding-www/src/features/ai-benchmark/core/bands.ts` and `shell/model-table.tsx` — the roster-max map and the two anchor threshold indices (`anchorIndices`) are now ALWAYS derived from `fullDataset` (defaulting to `dataset` itself when omitted), never from a possibly harness/class-filtered `dataset`. Added a matching `fullDataset` prop to `CapabilityChart`, `PriceChart`, and `ModelTable`, and wired `benchmark-content.tsx` to pass the unfiltered `dataset` as `fullDataset` to all three consumers alongside the already-filtered `filteredDataset`.

Regression tests added (both fail without the fix / pass with it, since the pre-fix `computeGroups`/component props had no second argument at all):
- `core/bands.unit.test.ts`: dynamically derives, from the real dataset, which opus/sonnet models survive a `codex-cli` or `opencode-go` harness filter (both harnesses exclude both anchors), then asserts `computeGroups(filtered, dataset)` keeps each model's full-roster band — plus an explicit "without the override" test proving the bug reproduces (collapses to `light`) when no `fullDataset` is passed.
- `shell/model-table.test.tsx`: renders `<ModelTable dataset={codexCliFiltered} fullDataset={dataset} locale="en" />` and asserts `gpt-5.6-sol`'s class cell reads "Opus" (not "Light"), plus a companion test showing the collapse when `fullDataset` is omitted.

Verified: `npx nx run ayokoding-www:test:unit` (3065 passed), `npx nx run ayokoding-www:specs:behavior:coverage` (unaffected), `npx nx affected -t typecheck lint` all green. Pushed as commit `b2d8bd281`.

apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx
Comment (wahidyankf) [reply apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx]: Fixed: `apps/ayokoding-www/src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx`'s `handleFilterChange` now calls `router.push(qs ? \`${pathname}?${qs}\` : pathname, { scroll: false })`, mirroring `cost-of-living-calculator/calculator-content.tsx`'s in-page filter navigations, with a comment explaining why it's mandatory (in-page filter/view state, not a page change).

Verified: `npx nx run ayokoding-www:test:unit`, `npx nx affected -t typecheck lint` both green. Pushed as commit `b2d8bd281`.

specs/apps/ayokoding/behavior/ayokoding-www/gherkin/tools/ai-benchmark.feature
wahidyankf: Reasoned reject (scope): confirmed the tag mismatch is real — `prd.md` §"Filters and URL state" tags AC-23 and AC-28 `@unit @e2e`, while the shipped `.feature` scenarios (lines 240/280 in this diff) carry `@unit` only, and `delivery.md`'s N-3/N-13 steps specify unit-layer binding only for both. But the root cause is the pre-existing prd.md↔delivery.md tag contradiction, not something introduced fresh by this diff's own AC intent — this PR faithfully implemented exactly what its own delivery plan (`delivery.md` N-0/N-3/N-13, already executed and merged as part of this PR's scope) specified, and adding new e2e bindings now would mean implementing scope delivery.md never asked for, unreviewed by this PR's own plan.

This mirrors the precedent set in Phase 6-7's PR #117, where a similar prd.md↔delivery.md drift on out-of-scope items was deferred for the same reason. The two legitimate remediations — (a) add the two missing e2e bindings, or (b) amend `prd.md` to drop `@e2e` from AC-23/AC-28 so it matches what was actually delivered — are both plan-document decisions belonging to whoever next revises `plans/in-progress/ayokoding-www-tools-ai-benchmark/prd.md` or `delivery.md`, not a code change to land silently inside this PR's review-fix pass. Leaving this thread open flags it for that follow-up; not resolving it here since the finding itself is valid and unaddressed in this PR.

apps/ayokoding-www/src/features/ai-benchmark/core/bands.ts
```diff
@@ -122,15 +122,24 @@ function compareForOrder(a: ModelScore, b: ModelScore): number {
...
+ * anchor thresholds are roster-relative), so filtering must only govern which models are
+ * displayed, never which models define the band boundaries. Passing the harness-filtered subset
+ * as BOTH `dataset` and (implicitly) `fullDataset` reproduces the anchor-collapse bug: when the
+ * filter excludes both `claude-opus-5` and `claude-sonnet-5`, the thresholds silently become
+ * `undefined` and every surviving rated model falls through to `light`.
  */
-export function computeGroups(dataset: Dataset): BandGroups {
-  const maxes = computeRosterMaxes(dataset);
-  const anchorIdx = anchorIndices(dataset, maxes);
+export function computeGroups(dataset: Dataset, fullDataset: Dataset = dataset): BandGroups {
```
wahidyankf, line 138: ## HIGH — this PR's own fix supersedes the documented design, and the plan still teaches the buggy one; `DD-11` is cited but never defined

**Severity**: `HIGH` (raised as `CRITICAL` by `pr-review-docs-maker` — **downgraded**, reasoning below) · **Confidence**: 96 · **Discipline**: docs / plan decision record · **Deduped 4 → 1 from**: `pr-review-docs-maker` F1, `pr-review-architecture-maker` F1, `pr-review-logic-maker` F3, plus `pr-review-governance-maker`'s routing note — four specialists reached the same defect independently.

### What is wrong

The cycle-1 fix landed in this line — `computeGroups` now derives the roster-max map and both anchor thresholds from `fullDataset`, always, under every filter. Three plan statements still assert the opposite, and all three are the design record a future contributor reads instead of this JSDoc.

**1. `tech-docs.md:67-68`** — verbatim at the pinned SHA:

> `filter.ts` narrows the `models.ts` roster before `score.ts` and `bands.ts` run over it, so the class bands **re-scale to whatever survives the filter** (DD-11).

That is now false. It is also a precise description of the cycle-1 `CRITICAL` bug.

**2. `tech-docs.md:97`** — the render/interaction sequence diagram still shows:

```
BC->>C: computeIndices(filtered) · assignBands(...)
```

Also false: indices and thresholds now come from the unfiltered roster; only membership comes from the filtered set.

**3. `delivery.md:2367`** — the acceptance text of this PR's own N-8 step:

> **N-8 GREEN**: intersect both filters over one filtered dataset, so the bands **re-scale to what remains** (DD-11)

The step that produced the bug still names the superseded behaviour as its goal.

**4. `DD-11` has no definition anywhere.** Grepped all five plan docs (`README.md`, `brd.md`, `prd.md`, `tech-docs.md`, `delivery.md`): `DD-11` is cited exactly twice — `tech-docs.md:68` and `delivery.md:2367` — and defined zero times. `tech-docs.md` is the only decision log (no `DD-` headings exist in `prd.md`/`brd.md`).

### Correcting `pr-review-docs-maker` on one point

Its finding reported that *"grep for `### DD-` finds DD-1 through DD-23, no DD-11 section"*. That is not accurate, and the inaccuracy matters:

- The decision log is **not contiguous** — the defined set is DD-1/2, DD-3, DD-5, DD-5a, DD-6, DD-7, DD-7a, DD-12, DD-16, DD-17a, DD-18, DD-19, DD-19a, DD-20a, DD-21, DD-22, DD-23.
- **DD-11 is not uniquely dangling**: `DD-14` (`tech-docs.md:669`) and `DD-17` (`tech-docs.md:365`, `:368`) are also cited with no defining section.

So "phantom citation" is a *class* of three, not a singular smoking gun. Only `DD-11` is in this PR's scope — it is the label attached to the behaviour this PR changed. `DD-14`/`DD-17` belong to whoever next revises the plan and are explicitly **not** being asked for here.

### Why `HIGH` and not `CRITICAL`

Per the [Criticality Levels Convention](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/development/quality/criticality-levels.md), `CRITICAL` is for broken functionality, build breaks, blocked users, or violated MUST requirements. None applies: the shipped behaviour is correct, CI is green, no reader is blocked, and no `repo-governance/` convention is breached (which is why governance routed rather than raised — correctly). What is degraded is **maintainability**: the design doc now teaches the exact mental model that produced the cycle-1 `CRITICAL`. That is `HIGH` by definition ("significant quality degradation … maintainability problems").

Stating this explicitly because it is the load-bearing coordination call here: **four specialists converging raises confidence, not severity.** Convergence-driven severity inflation is a known failure mode of multi-reviewer fan-out, and three of the four specialists rated this at or below `HIGH` anyway.

### Why this is in scope

The drift was **created by commit `b2d8bd281`, inside this PR** — before it, plan and code agreed. And the plan already has an established mechanism for exactly this: DD-5a amends DD-5, DD-19a amends DD-19, DD-20a amends DD-20, DD-23 amends DD-21, DD-22 amends the Testing strategy. No new scope is being invented; the PR is asked to record a reversal it itself performed.

Worth naming the substantive decision buried in here, which `pr-review-architecture-maker` was right to flag: the fix is **deliberately broader than the bug required**. The `CRITICAL` only needed the anchor-exclusion edge case fixed; the fix makes thresholds full-roster-derived under *all* filters, including ones where both anchors survive. That is the right call — it is the only rule that is stable under a roster refresh — but it is a design decision, and an unrecorded design decision in a plan that maintains a decision log is the defect.

### Suggested resolution

- `tech-docs.md:67-68` — state that band thresholds and the roster-max map are **always** derived from the full unfiltered roster; filtering governs display and membership only.
- `tech-docs.md:97` — amend the diagram step to separate the two: membership from the filtered set, thresholds from the full roster.
- `delivery.md:2367` — amend N-8's descriptive text so the executed step no longer states the superseded behaviour.
- Either add `### DD-11a — Bands never re-scale under a filter (amends DD-11)`, **or** — since DD-11 was never defined, so there is nothing to amend — drop the dangling `(DD-11)` citation from both sites and record the rule as a properly numbered new decision. The second is more honest.

**A reasoned-reject is available and legitimate**: if the maintainer prefers plan-document edits to happen in a plan-revision pass rather than a review-fix pass, say so on this thread. But then code and plan ship disagreeing — and that disagreement is what produced the cycle-1 `CRITICAL` in the first place.

---

apps/ayokoding-www/src/features/ai-benchmark/shell/capability-chart.tsx
```diff
@@ -84,14 +84,22 @@ function computeLayout(
...
   locale: Locale;
 };
 
-export function CapabilityChart({ dataset = defaultDataset, locale }: CapabilityChartProps) {
+export function CapabilityChart({ dataset = defaultDataset, fullDataset, locale }: CapabilityChartProps) {
   const titleId = useId();
   const unratedHeadingId = useId();
 
-  const groups = computeGroups(dataset);
+  const groups = computeGroups(dataset, fullDataset ?? dataset);
```
wahidyankf: ## HIGH — the two components the cycle-1 CRITICAL was actually about have zero regression coverage; reverting their fix passes the entire suite

**Severity**: `HIGH` · **Confidence**: 96 · **Discipline**: test integrity ([Regression Test Mandate](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/development/quality/regression-test-mandate.md)) · **Deduped 2 → 1 from**: `pr-review-integrity-maker` F1 (`HIGH` 95, empirically proven) + `pr-review-logic-maker` F2 (`MEDIUM` 85). Integrity's severity survives.

### What is wrong

The cycle-1 fix added regression tests in exactly two places — `core/bands.unit.test.ts` (the pure function) and `shell/model-table.test.tsx`. Neither covers `CapabilityChart` or `PriceChart` — **the two components the cycle-1 `CRITICAL` thread named specifically**, and the two that `benchmark-content.tsx:73-74` feeds the filtered dataset to.

### Evidence 1 — empirical (integrity's revert-and-test)

Reverting **only** `capability-chart.tsx:102` and `price-chart.tsx:161` back to their pre-fix `computeGroups(dataset)` form — leaving `bands.ts`, `model-table.tsx` and `benchmark-content.tsx` fixed — then running the full relevant suite: **5 test files, 189 tests, all pass.** Including the AC-23/AC-24 fe-step scenarios, which assert *which model ids appear*, never *which band a model lands in*.

### Evidence 2 — structural proof (verified here at the pinned SHA), which explains why

`fullDataset` appears in exactly seven files across `src/` and `test/`:

```
src/app/[locale]/tools/ai-benchmark/benchmark-content.tsx
src/features/ai-benchmark/core/bands.ts
src/features/ai-benchmark/core/bands.unit.test.ts
src/features/ai-benchmark/shell/capability-chart.tsx
src/features/ai-benchmark/shell/price-chart.tsx
src/features/ai-benchmark/shell/model-table.tsx
src/features/ai-benchmark/shell/model-table.test.tsx
```

Only two are tests, and **neither renders `CapabilityChart` or `PriceChart`.** Every render of those two components anywhere in the suite passes `dataset` alone:

- `test/unit/fe-steps/ai-benchmark.steps.tsx:928, :964, :1001, :1024` — `CapabilityChart`, no `fullDataset`
- `test/unit/fe-steps/ai-benchmark.steps.tsx:1068, :1108, :1145, :1309` — `PriceChart`, no `fullDataset`
- `shell/capability-chart.test.tsx` — no `fullDataset` reference at all
- `shell/price-chart.test.tsx` — no `fullDataset` reference; its `opencode-go` mentions (`:47`, `:49`, `:85`, `:87`) are subscription-pricing fixtures, not band regression

So in **100% of existing tests** `fullDataset` is `undefined`, `fullDataset ?? dataset` degenerates to `dataset`, and every test therefore exercises precisely the pre-fix code path. The revert is a semantic no-op under this suite **by construction, not by luck** — which is a proof rather than a sample, and is why the consolidated confidence lands at 96 rather than integrity's 95.

### Evidence 3 — the wiring itself is untested

`benchmark-content.tsx` is the *only* production code that supplies `fullDataset` at all (`:73`, `:74`, `:78`), and it has **no test file**: grep for `BenchmarkContent` across `src/`, `test/` and the e2e steps returns only `benchmark-content.tsx` and `page.tsx`. Deleting `fullDataset={dataset}` from all three call sites there would also pass every test.

### Evidence 4 — e2e does not cover it either

The only anchor-excluding harness exercised end-to-end is `?harness=opencode-go` (`apps/ayokoding-www-fe-e2e/src/steps/ai-benchmark.steps.ts:106`), and that step asserts **price**, not band placement. The other two navigations (`:120`, `:142`) use `?harness=cursor&class=opus`, and `cursor` retains both anchors — so they cannot trip the bug by construction.

### Why `HIGH` and not `CRITICAL`

The Regression Test Mandate is **technically satisfied**: `bands.unit.test.ts:252` (filtered + `fullDataset` keeps the full-roster band) and `:276` (collapse reproduces without it), plus `model-table.test.tsx`, do fail-before / pass-after. `pr-review-governance-maker` was right to return clean on mechanical conformance. But the mandate's *purpose* — the bug cannot silently come back — is unmet for the two components the bug was reported against. Degraded quality, not broken functionality: `HIGH`.

### Correcting `pr-review-logic-maker`'s framing

Logic filed this `MEDIUM` as *"no regression test proving their own `fullDataset` wiring holds"*. It is worse than that: the wiring is not merely **unproven**, it is **unprovable by the current suite** — evidence 2 shows no test in the repo can distinguish the fixed charts from the unfixed ones. `MEDIUM` understates it; integrity's `HIGH` stands.

### Suggested fix

Mirror `model-table.test.tsx`'s pattern into `shell/capability-chart.test.tsx` and `shell/price-chart.test.tsx`:

- render each with a `codex-cli`- or `opencode-go`-filtered dataset **plus** `fullDataset={dataset}`, and assert a model that is `opus`/`sonnet` on the full roster renders under its correct band heading rather than `light`;
- **derive** the filtered set and the expected model from the real dataset the way `bands.unit.test.ts:229-276` already does, rather than hardcoding ids — otherwise a roster refresh silently voids the test;
- add a companion "omitted `fullDataset`" test showing the collapse still reproduces — **but see Finding 3**: if you adopt the required-prop change, that companion test becomes a compile error and should simply be dropped. Do not write both.

Also worth closing evidence 3: at minimum, one assertion that fails if `fullDataset={dataset}` is dropped from any of the three consumers in `benchmark-content.tsx`.

---

apps/ayokoding-www/src/features/ai-benchmark/shell/price-chart.tsx
```diff
@@ -129,15 +136,30 @@ function axisMaxOf(bands: BandLayout[]): number {
...
 export type PriceChartProps = {
   dataset?: Dataset;
+  /**
+   * The full unfiltered roster. Band thresholds (the anchor indices) and the roster-max map are
+   * ALWAYS derived from this dataset, never from `dataset` — `dataset` may be a harness/class
+   * filtered subset that excludes both anchor models, and re-deriving thresholds from it would
+   * silently collapse every rated model to `light` (DD-5a: bands are roster-relative to the FULL
+   * population; filtering governs display only). Defaults to `dataset` itself when omitted.
+   */
+  fullDataset?: Dataset;
```
wahidyankf: ## MEDIUM — `fullDataset` is optional with a silent self-fallback, reproducing the exact fragility class that caused the cycle-1 CRITICAL

**Severity**: `MEDIUM` · **Confidence**: 88 (raw 82, raised by tool-verify) · **Discipline**: architecture (API design) · **Source**: `pr-review-architecture-maker` F2, sole raiser.

### What is wrong

`fullDataset` is optional at every boundary introduced by the cycle-1 fix:

- `core/bands.ts:138` — `computeGroups(dataset: Dataset, fullDataset: Dataset = dataset)`
- `shell/capability-chart.tsx:94` / `:102` — `fullDataset?: Dataset` → `computeGroups(dataset, fullDataset ?? dataset)`
- `shell/price-chart.tsx:146` / `:161` — same shape
- `shell/model-table.tsx:212` / `:217` — `fullDataset ?? dataset`

Omitting it is silently legal and silently wrong. That is **precisely the pre-fix contract**: before `b2d8bd281` the second argument did not exist, and the bug was that the filtered dataset served as its own roster. After the fix, forgetting the prop reproduces the identical bug with identical silence — no type error, no test failure, just every rated model collapsing to `light`.

### Why this cleared the reasonableness filter

Normally "make an optional parameter required, no failure demonstrated" is a hardening nitpick and gets dropped here. It survives because **Finding 2 demonstrates the blind spot is live, not hypothetical**: `benchmark-content.tsx:73-74/78` is the only code that supplies `fullDataset`, it has no test file, and no test anywhere else supplies it either. The type system is currently the only thing that could catch a regression at that call site — and it is opted out by the default. Architecture raised this at 82 as a "future silent footgun"; tool-verify shows the footgun is already loaded.

### Narrowing architecture's remedy, on cost grounds

Architecture proposed making the parameter required outright. Verified the cost, and it is not free: making **`computeGroups`'s** second parameter required touches ~16 call sites — `core/filter.ts:59` plus 15 test call sites (`filter.unit.test.ts:46, :61`; `bands.unit.test.ts:103, :141, :181, :229, :276`; `fe-steps/ai-benchmark.steps.tsx:338, :396, :933, :969, :979, :996, :1011, :1039`) — and it adds noise where the self-fallback is genuinely correct: `filter.ts:59` operates on the full roster by design and would become `computeGroups(dataset, dataset)`.

Recommended instead, and it is strictly cheaper:

- Make `fullDataset` a **required prop on the three components** (`CapabilityChart`, `PriceChart`, `ModelTable`) — that is where the demonstrated failure mode lives. Cost is ~10 test render sites, all mechanical.
- Leave `computeGroups`'s default alone.
- Drop the now-redundant `?? dataset` in `capability-chart.tsx:102` and `price-chart.tsx:161`. Today the three consumers are inconsistent — `model-table.tsx:51` passes `fullDataset` straight through and relies on the core default, the two charts coalesce first. Same behaviour, two spellings.

Architecture's claim that this is a **zero-behaviour-change** edit holds: all three production call sites already supply the prop.

### Coordination note for the fixer

If you adopt this, the "companion omitted-`fullDataset`" **component** test suggested in Finding 2 becomes a compile error — drop it rather than fight it; a compile error is a strictly better guard than a test. The **core-level** collapse proof at `bands.unit.test.ts:276` is unaffected and must stay.

Reasoned-rejecting this on cost grounds is reasonable. If you do, say so on this thread so cycle 3 does not re-raise it.

---

wahidyankf (commented):
**Risk tier**: full
**Prior-cycle human dismissals respected**: 0 human dismissals; 1 prior-cycle agent reasoned-reject respected (AC-23/AC-28 tag drift — see below)

---

## Cycle 2 — consolidated review

**3 findings posted** (0 CRITICAL, 2 HIGH, 1 MEDIUM) out of **7 raw findings** raised across the eight specialists — 4 of which were the same defect seen four different ways.

**Blocking status**: no `CRITICAL` survived, so this review is **not hard-blocking**. Both `HIGH` findings are "should fix before merge, this cycle" per the [Criticality Levels Convention](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/development/quality/criticality-levels.md), and this is cycle 2 of up to 3 — leaving them to cycle 3 forfeits the remaining review budget. (GitHub rejects `REQUEST_CHANGES` on one's own PR under the current `gh` identity posture, so severity labels on each thread carry the status, not the review state.)

Risk tier is `full` on size (13 files, ~950 insertions); no security-sensitive path is touched, so no override applied. Re-computed this cycle rather than inherited from cycle 1.

### Coordination record

| Raw finding(s) | Disposition |
| --- | --- |
| Docs F1 (`CRITICAL` 90) + Architecture F1 (`HIGH` 92) + Logic F3 (`LOW` 82, informational) + Governance routing note — DD-11 / tech-docs drift | **Deduped 4 → 1.** Severity resolved **down** to `HIGH`; confidence raised to 96. Posted (Finding 1). |
| Integrity F1 (`HIGH` 95, empirical) + Logic F2 (`MEDIUM` 85) — charts lack regression coverage | **Deduped 2 → 1** at `HIGH` 96. Integrity's revert-and-test survives as the evidence, independently corroborated here by a structural proof. Posted (Finding 2). |
| Architecture F2 (`MEDIUM` 82) — optional `fullDataset` with silent self-fallback | **Survived** the reasonableness filter, confidence raised to 88 — tool-verify moved it from speculative hardening to a *demonstrated* blind spot. Posted (Finding 3) with a narrowed, cheaper remedy. |
| Logic F1 (`HIGH` 88) — AC-23/AC-28 `@e2e` tag drop | **Dropped — settled in cycle 1.** Tool-verified as byte-identical to the already-dispositioned finding (below). |
| Logic's confirmation that the fix generalizes to all five harness values | No action — a clearance, not a finding. Corroborated: the mechanism in `bands.ts:138-142` is harness-agnostic. |
| Performance's O(filtered) → O(full-roster) observation | Agreed below the confidence bar and not a new regression class (38-model fixed roster, no hot path). Not posted. |
| Security, Governance, Instruction-decay — all clean | No action. Governance correctly *routed* the DD-11 drift rather than judging it: no `repo-governance/` convention is violated, so it is not governance's call. |

### Dropped — Logic F1, the AC-23/AC-28 `@e2e` tag drop (raw `HIGH`, confidence 88)

`pr-review-logic-maker` flagged this as *"DISTINCT from cycle-1's reasoned-rejected LOW … re-verify this isn't the same thing before re-raising"*. Verified at the pinned SHA — **it is the same thing**:

- Cycle 1 posted it as `LOW` on `specs/apps/ayokoding/behavior/ayokoding-www/gherkin/tools/ai-benchmark.feature:240` ([review 4807025458](https://github.com/wahidyankf/ose-public/pull/118#pullrequestreview-4807025458)) — same two ACs, same two prd.md tag sites, same two `.feature` lines, same "`@e2e` silently dropped" claim.
- Re-verified independently: `prd.md` tags AC-23 and AC-28 `@unit @e2e`; the shipped feature file carries `@unit` only at lines 240 and 280. Unchanged — the cycle-1 fix commit `b2d8bd281` touched no spec or plan file, so the situation is byte-identical to when it was dispositioned.
- `pr-review-fixer` **reasoned-rejected** it on scope grounds (the remediation is a plan-document decision, prd.md↔delivery.md, not a code change to land silently in a review-fix pass) and the thread is resolved.

Re-raising a settled finding one severity band *higher* (`LOW` → `HIGH`) with no new evidence is exactly what the reasonableness filter exists to stop. Dropped.

One precision note for the record: this was an **agent** reasoned-reject, not a human dismissal — the header field above says 0 human dismissals for that reason. The maintainer remains entirely free to reopen it; nothing here forecloses that.

### Note for the fixer

Findings 1 and 2 are causally linked to the cycle-1 `CRITICAL`, not new territory: Finding 1 is the design record that fix invalidated and did not update, Finding 2 is the regression coverage that fix did not extend to the two components it was reported against. Finding 3 is a type-level alternative to part of Finding 2 — read them together before starting; adopting Finding 3 makes one of Finding 2's suggested tests unnecessary (noted inline).

---

apps/ayokoding-www/src/features/ai-benchmark/core/bands.ts
wahidyankf: Fixed: replaced the stale "class bands re-scale to whatever survives the filter (DD-11)" claim in `tech-docs.md:67-69` and the sequence diagram (`computeIndices(filtered) · assignBands(...)`) with prose/diagram steps describing the actual shipped behaviour — thresholds always derive from the full unfiltered roster, filtering governs membership/display only. Added a new `### DD-24 — Band thresholds always derive from the full roster, never re-scale under a filter` section (placed after DD-20a, since DD-11 was never defined — the "more honest" option per the finding's own suggested resolution) that records this as a properly numbered decision and explicitly narrates the DD-11 label's origin and retirement. Updated `delivery.md`'s N-8 step prose (line ~2367) to the corrected text (checkbox state left untouched — still `- [ ]`, per the plan-doc convention). Also swept `benchmark-content.tsx`'s own inline comment, which carried the identical stale `(DD-11)` citation and was not explicitly named in the finding but is the same defect class.

Commit: `fd7d5ec73` (pushed to `ayokoding-www-tools-ai-benchmark/phase-8-filters`).

apps/ayokoding-www/src/features/ai-benchmark/shell/capability-chart.tsx
wahidyankf: Fixed: added component-level regression tests to `capability-chart.test.tsx` and `price-chart.test.tsx`, mirroring `bands.unit.test.ts`'s pattern — derives the `codex-cli`-filtered roster and the surviving opus/sonnet model from the REAL dataset (no hardcoded id), renders the chart with `dataset={filtered}` + `fullDataset={dataset}`, and asserts the survivor's row lives inside its correct full-roster band group (`capability-chart-band-<band>` / `price-chart-band-<band>`). Each file also carries a companion "WITHOUT fullDataset the bug would reproduce" proof at the `computeGroups` level, showing the assertion is not vacuous, without needing a compile-breaking omitted-prop render (see Finding 3 below).

Evidence 3 (the untested `benchmark-content.tsx` wiring) is now closed as a side effect of Finding 3's fix: `fullDataset` is a required prop on all three components, so dropping it from any of `benchmark-content.tsx`'s three call sites is now a `TS2741` compile error, not a silent gap.

Commit: `fd7d5ec73` (pushed to `ayokoding-www-tools-ai-benchmark/phase-8-filters`). Verified: `npx nx run ayokoding-www:test:unit` — 3070 passed, 0 failed.

apps/ayokoding-www/src/features/ai-benchmark/shell/price-chart.tsx
wahidyankf: Fixed, adopting the narrower recommended remedy exactly as specified: `fullDataset` is now a REQUIRED prop (`fullDataset: Dataset`, not `fullDataset?: Dataset`) on `CapabilityChart`, `PriceChart`, and `ModelTable`; the redundant `?? dataset` self-fallback is dropped from `capability-chart.tsx`/`price-chart.tsx`/`model-table.tsx`'s call into `computeGroups`/`computeScoreViews`. `computeGroups`'s own second parameter keeps its default (`fullDataset: Dataset = dataset`) untouched, so `core/filter.ts:59`'s single-argument call needed no edit.

Cost was larger than the "~10 test render sites" estimate once TypeScript's `TS2741`/`TS2769` enumerated every real call site: 25 across `capability-chart.test.tsx` (3), `price-chart.test.tsx` (4), `chart-order-parity.test.tsx` (2), `model-table.test.tsx` (4), and `test/unit/fe-steps/ai-benchmark.steps.tsx` (12) — every one supplied the same dataset already in scope as `fullDataset`, so no test's behavior changed, only its call signature.

Per your coordination note, `model-table.test.tsx`'s existing "reproduces the bug when fullDataset is omitted" test (the pre-existing runtime proof, not the Finding-2-suggested new one) is now a `TS2741` compile error by construction — dropped it and left an explanatory comment, since a compile error is a strictly stronger guard than the runtime assertion it replaces. The "shows the model's full-roster class when fullDataset is passed" test and the core-level `bands.unit.test.ts:276` collapse proof are both unaffected and still pass.

Commit: `fd7d5ec73` (pushed to `ayokoding-www-tools-ai-benchmark/phase-8-filters`). Verified: `npx nx run ayokoding-www:typecheck`, `npx nx run ayokoding-www:test:unit` (3070 passed), `npx nx affected -t typecheck lint` (25 projects) all green.

plans/in-progress/ayokoding-www-tools-ai-benchmark/delivery.md
```diff
@@ -2363,8 +2363,9 @@ typecheck, lint for 25 projects and 6 tasks they depend on` (52 of 56 tasks serv
       Then only models satisfying both filters are shown
     ```
 
-- [ ] [AI] **N-8 GREEN**: intersect both filters over one filtered dataset, so the bands re-scale to
-      what remains (DD-11) — command: `npx nx run ayokoding-www:test:unit` — acceptance: AC-25 passes
+- [ ] [AI] **N-8 GREEN**: intersect both filters over one filtered dataset for membership/display,
```
wahidyankf: ## HIGH — Phase 8's entire delivery checklist is still unticked: 28 items, zero evidence blocks, while the phase is implemented and CI is green

**Severity**: `HIGH` · **Confidence**: 95 · **Discipline**: governance / plan-execution · **Deduped 2 → 1 from**: `pr-review-governance-maker` F1 (`HIGH` 92) + `pr-review-docs-maker` F1 (`HIGH` 92) — two specialists reached this independently, and I verified it directly rather than counting agreement.

### What is wrong

Every one of the 28 checklist items in `delivery.md`'s Phase 8 section (lines 2297-2456 — the worktree-provisioning item, N-0 through N-19, and the 6-item Phase 8 Gate) is still `- [ ]`, with no `Date` / `Status` / `Notes` evidence block anywhere in the section.

Tool-verified at head `fd7d5ec7`, not taken on the specialists' word:

```
$ sed -n '2297,2456p' delivery.md | grep -c '^\s*- \[ \]'    → 28
$ sed -n '2297,2456p' delivery.md | grep -c '^\s*- \[x\]'    → 0
```

Meanwhile the phase is *done*: `benchmark-filters.tsx` ships, the eight ACs are bound, and CI is green across all three commits on this branch.

**The line this comment is anchored to is the proof.** Line 2366 is N-8, and this PR's cycle-2 fix pass **rewrote its prose** — retargeting it from the phantom `DD-11` to the real `DD-24` — and left the box unticked while doing so. The item was edited by an agent that had just verified the underlying work, and the checkbox still did not move.

### Why this is `HIGH`, not cosmetic

This violates two rules the repo states as non-negotiable:

- [Plans Organization Convention](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/conventions/structure/plans.md) — **Checkbox lockstep**.
- [plan-execution workflow](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/workflows/plan/plan-execution.md) — **"Delivery.md Is Sacred — Atomic Sync Ritual"**, marked explicitly NON-NEGOTIABLE.

The practical cost is that `delivery.md` currently asserts Phase 8 has not started. Any agent or human resuming this plan reads 28 open items and either redoes shipped work or has to reconstruct state from git history — which is exactly the failure the ritual exists to prevent.

### Precedent, verified against the two prior merged PRs in this same plan

`pr-review-docs-maker` cross-referenced the already-merged sibling PRs, and I confirmed it against the GitHub API:

| PR | Phase | `- [x]` flips inside its own `delivery.md` diff |
| --- | --- | --- |
| #115 | Phase 4 | 47 |
| #117 | Phase 6-7 | 28 |
| **#118** | **Phase 8** | **0** |

Every prior delivery unit in this plan ticked its own boxes inside its own PR diff. This one ticks none. Phase 1/2's own gate notes in this same file record that a similar historical omission was later *reconciled*, not left standing — so the plan's own precedent is reconciliation, twice over.

### Suggested fix

Tick all 28 Phase 8 items to `- [x]` with `Date` / `Status` / `Notes` triads, each verified against actual repo state at this head — not asserted. Concretely: N-0 through N-19 against the shipped files and the Gherkin scenarios they claim to bind, and the 6 gate items against the real command outcomes (the PR body's test plan already records the four green commands and the unlinked-page check, so the evidence exists — it just is not in `delivery.md`).

Do not tick anything you cannot verify. If any item genuinely was not performed, say so in its `Notes` rather than ticking it.

plans/in-progress/ayokoding-www-tools-ai-benchmark/delivery.md
```diff
@@ -2363,8 +2363,9 @@ typecheck, lint for 25 projects and 6 tasks they depend on` (52 of 56 tasks serv
...
+      while band thresholds keep deriving from the full unfiltered roster (DD-24) — command:
+      `npx nx run ayokoding-www:test:unit` — acceptance: AC-25 passes
 - [ ] [AI] **N-9 RED**: bind AC-18 in both the unit and e2e step files
       — command: `npx nx run ayokoding-www:test:unit` — acceptance: fails
   - _Gherkin (binds) → AC-18 "A harness filter switches the price chart to that harness's rate"_
```
wahidyankf, line 2454: ## MEDIUM — `delivery.md:2380` (N-10) tells the next reader to call `rateFor`, a function that does not exist

**Severity**: `MEDIUM` · **Confidence**: 95 (raw 90, raised by tool-verify) · **Discipline**: docs · **Source**: `pr-review-docs-maker` F2, sole raiser.

*Anchored to line 2371 — the AC-18 Gherkin reference that N-10 GREEN implements — because the real target, line 2380, falls outside this PR's `delivery.md` hunk and GitHub cannot anchor a review comment there. The exact target is stated below.*

### What is wrong

`delivery.md:2380` reads:

> `- [ ] [AI] **N-10 GREEN**: pass the active harness into `rateFor` so the price chart switches rate set`

No `rateFor` exists. The shipped export is `rateForHarness`:

- `apps/ayokoding-www/src/features/ai-benchmark/core/price.ts:84` — `export function rateForHarness(model: Model, harness: HarnessId): SelectedRate`
- consumed under that name in `shell/price-chart.tsx`

Tool-verified: `grep -rn "export function rateFor" core/price.ts` returns `rateForHarness` and nothing else, and `grep -rn "rateFor("` across `apps/ayokoding-www/src` returns **zero** bare `rateFor(` occurrences. The step description was never updated from an earlier working name — this is not a case of two functions where the doc names the wrong one; the named function is simply absent from the codebase.

### Why it is worth fixing rather than dropping as a nitpick

It survived the reasonableness filter for one specific reason: `delivery.md:2380` is *load-bearing for this very review cycle*. `pr-review-logic-maker` cited exactly this line — N-10's "the price chart" wording — as the authority for suppressing its own harness-filter observation about `model-table.tsx`. A step description that a reviewer treats as the scope contract should name the symbol that actually shipped. A stale identifier in that position sends the next reader to a symbol that does not resolve.

### Suggested fix

Edit `delivery.md:2380` to name `rateForHarness`. One-word change, same line you will already be touching to tick N-10's checkbox under F1.

wahidyankf (commented):
**Risk tier**: full
**Security-sensitive-path override applied**: no (tier was already `full` on size — 18 files, ~1,100 changed lines; the diff touches no secrets/`.env`, git-identity, CI-workflow, or `pr-merge-protocol.md` path)
**Diff coverage**: full diff reviewed in one pass, no generated-file exclusion (D13)
**Prior-cycle human dismissals respected**: none this cycle — all 6 prior threads were resolved by agent `pr-review-fixer` comments (5 fixes + 1 reasoned reject), not human dismissals. The 1 agent reasoned-reject is respected as settled; see the drop log.

---

## Cycle 3 of 3 (final) — 2 findings, both blocking

**This review is BLOCKING despite its `COMMENT` state** — `gh` authenticates as the PR author, so `REQUEST_CHANGES` is structurally unavailable. Blocking status is carried by the `HIGH` severity label below.

| # | Severity | Conf | Discipline | Where |
| F1 | `HIGH` | 95 | governance / plan-execution | `delivery.md` Phase 8 — all 28 items unticked |
| F2 | `MEDIUM` | 95 | docs | `delivery.md:2380` — `rateFor` names no shipped function |

Both are plan-document defects in `plans/in-progress/ayokoding-www-tools-ai-benchmark/delivery.md`, a file this PR already modifies. **Zero code findings survived this cycle** — the shipped implementation cleared all eight disciplines.

### What the code-facing disciplines found

Four specialists returned zero findings, and two more declared observations and correctly declined to raise them:

- **Security** — 0. URL-state decode/encode traced; the strict `isKnownHarness`/`isKnownBand` allowlist is pre-existing and unmodified. No secrets, no injection vector.
- **Integrity** — 0, and it is the strongest signal in this cycle: it **empirically re-verified by revert-and-test** that cycle-2's chart regression coverage genuinely holds (`capability-chart.test.tsx` / `price-chart.test.tsx` now fail when the cycle-1 fix is reverted, which they did not before), and separately confirmed that dropping `model-table.test.tsx`'s omitted-`fullDataset` negative test was *not* test-integrity erosion — reintroducing it now produces a real `TS2741`, i.e. a compile-time guard strictly stronger than the runtime assertion it replaced. This is exactly the empirical standard the [Regression Test Mandate](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/development/quality/regression-test-mandate.md) asks for.
- **Instruction-decay** — 0. No framework/build/CI/env change; `AGENTS.md` / `CLAUDE.md` untouched and under budget.
- **Logic** — 0 above bar. It flagged at ~65 that `model-table.tsx`'s price cells do not switch with the active harness the way `price-chart.tsx` does, then verified against `delivery.md:2380` (N-10) and Gherkin AC-18, both of which scope the harness-rate switch to *the price chart* only. Plan-sanctioned scope, not a defect — correct suppression.
- **Performance** — 0. The redundant `computeGroups`/`computeScoreViews` passes over `fullDataset` (3-4x per filter-driven re-render) are real but operate on a fixed 38-row roster with no growth-rate evidence and no measured hot-path impact. Correctly below bar.
- **Architecture** — 1 raised, dropped here (see drop log).

---

## Drop log — 2 raw findings filtered out

Recorded so the fixer knows these were adjudicated, not missed. **Do not act on either.**

### Dropped: architecture `MEDIUM`/85 — `computeGroups`'s `fullDataset` is still an optional defaulted parameter

The finding is factually correct: `core/bands.ts:138` is still `computeGroups(dataset: Dataset, fullDataset: Dataset = dataset)`, and `core/filter.ts:59`'s `computeGroups(dataset)` is the last single-arg production call.

It is dropped as **re-litigation of a cycle-2 ruling with no new evidence**. Cycle 2's own consolidated review raised this exact parameter, priced the change (16 call sites, 15 of them tests), and explicitly recommended the narrower remedy — verbatim: *"Leave `computeGroups`'s default alone"* and *"`filter.ts:59` operates on the full roster by design and would become `computeGroups(dataset, dataset)`"*. `pr-review-fixer` adopted that remedy exactly and said so on the thread. Re-raising the residue the prior cycle deliberately scoped out, on the final cycle, is not a new finding.

Tool-verified that no live defect hides behind the drop: `filterModels` has exactly one production caller, `benchmark-content.tsx:30`, which passes the unfiltered `dataset`. `filter.ts:59` is therefore correct today, and its `computeGroups` result is used only to build the class-filter band lookup. What remains is a latent hardening opportunity with no demonstrated failure — a nitpick under [Simplicity Over Complexity](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/principles/README.md), and strictly weaker than the cycle-2 finding it descends from, because the live blind spot that justified *that* one (the three shell components) is now closed by required props.

If a maintainer wants the deeper hardening, it belongs in a follow-up, not in this PR's final review cycle.

### Dropped: docs `MEDIUM`/88 — `prd.md` AC-23/AC-28/AC-27 test-tier tags disagree with the shipped `.feature`

Facts verified and correct: `prd.md:606-608`/`:646-648` tag AC-23 and AC-28 `@unit @e2e`; the shipped `ai-benchmark.feature` tags both `@unit` only; `ai-benchmark.steps.ts` binds only AC-18/AC-22/AC-27 at the e2e layer. The reported AC-27 asymmetry also checks out (`prd.md` says `@e2e`, shipped says `@unit @e2e`).

The docs specialist explicitly asked whether the doc-accuracy framing constitutes new evidence re-opening the cycle-1 settled finding. It does not. Cycle 1's reasoned reject already enumerated **both** remediation directions, including the exact one proposed here — verbatim: *"(b) amend `prd.md` to drop `@e2e` from AC-23/AC-28 so it matches what was actually delivered"* — and deferred both as plan-document decisions *"belonging to whoever next revises `prd.md` or `delivery.md`, not a code change to land silently inside this PR's review-fix pass."* Same fact pattern, same proposed fix, already ruled on. `prd.md` is also not in this PR's diff.

The new AC-27 sub-point is an over-delivery (shipped coverage is broader than documented), not a dangerous-direction gap, and falls under the identical already-settled deferral. Thread [`PRRT_kwDOQXc0486Ut3T3`](https://github.com/wahidyankf/ose-public/pull/118#discussion_r) stays as the standing flag for that follow-up.

---

## Merge posture

F1 and F2 are both confined to one plan document and are mechanical to fix. Once they land, this PR has no outstanding blocking finding across all eight disciplines and satisfies the final cycle of the [PR Review Quality Gate](https://github.com/wahidyankf/ose-public/blob/main/repo-governance/workflows/pr/pr-review-quality-gate.md).

plans/in-progress/ayokoding-www-tools-ai-benchmark/delivery.md
wahidyankf: Fixed: reconciled all 28 Phase 8 checklist items in `delivery.md` (lines 2301-2456, plus the Phase 8 Gate) with `Date`/`Status`/`Notes` evidence triads, verified against the current shipped code and fresh test runs at head `fd7d5ec73` (not asserted) — pushed as `a35bb5ad3`.

**What was actually verified before ticking** (not bulk-ticked from assumption):

- Fresh `npx nx run ayokoding-www:test:unit --skip-nx-cache` → `142 test files passed`, `3070 passed | 6 skipped`, exit 0.
- `npx nx run ayokoding-www:specs:behavior:coverage` → `42 specs, 327 scenarios, 1177 steps — all covered`, exit 0.
- `npx nx affected -t typecheck lint --base=origin/main` → all 25 projects + 6 dependent tasks succeed, no new warning class.
- Unlinked-page check → `grep -c "tools/ai-benchmark"` prints `0` for both `tools/page.tsx` and `footer.tsx`.
- Read every shipped source file this phase's items describe (`filter.ts`, `url-state.ts`, `benchmark-filters.tsx`, `benchmark-content.tsx`, `price.ts`) and every `@covers`-annotated unit/e2e step binding each AC, cross-referencing line numbers into each item's `Notes`.
- Ran the real Playwright `test:e2e` suite **twice** to check for flake. All six `ai-benchmark` e2e scenarios (locale headings ×2, "Each chart exposes an accessible name", "the whole roster", "switches the price chart to that harness's rate", "reloaded filtered URL reproduces the same view") passed in both runs across all three browsers — 18/18. The whole-project command itself exited non-zero both times, but from a *different*, non-overlapping set of failures each run (`course-rehome-redirects`, `ia-navigation-revamp`, `i18n`, `learn-three-bucket`/`learn-reorg-redirects`, once `accessibility`) — none touching any file this PR's diff changed, and the differing failure set between runs is the signature of environment-driven flake, not a regression this PR introduced. Confirmed via `gh run view --log` that GitHub Actions CI does not run `test:e2e` at all for this PR (the "TypeScript quality gate" job runs `test:quick`, which stops at `test:unit`/`test:coverage`/`test:specs`), so this pre-existing flake does not block CI green.

**25 of the 28 items are ticked `[x]`. 3 are deliberately left `[ ]`, each with a `Notes` block explaining why, per your own instruction not to tick anything unverifiable:**

1. The Phase 8 Gate's `npx nx run ayokoding-www-fe-e2e:test:e2e exits 0` item — the acceptance clause is the whole-project command, and it genuinely does not exit 0 right now (see above). Bulk-ticking it would have reintroduced exactly the kind of inaccurate documentation this finding flagged.
2. "Run the Delivery-Boundary Integration Protocol" and "Run Post-Push CI Verification" — these cover PR merge, fast-forwarding local `main`, and worktree removal, none of which has happened yet (this PR is still mid-review-cycle). Ticking them now would misrepresent the PR as merged. This matches the same-file's own existing precedent: Phase 7's Gate (in this same `delivery.md`, already merged via PR #117) leaves these identical two items unticked too.

Also folded in the fix for the sibling finding (F2, `rateFor` → `rateForHarness`) since it lands on N-10's own line, which this pass was already touching.

plans/in-progress/ayokoding-www-tools-ai-benchmark/delivery.md
wahidyankf: Fixed: `delivery.md:2380` (N-10) now reads `rateForHarness` instead of `rateFor` — pushed as `a35bb5ad3`.

Confirmed the real export at `apps/ayokoding-www/src/features/ai-benchmark/core/price.ts:84` (`export function rateForHarness(model: Model, harness: HarnessId): SelectedRate`), and that `price-chart.tsx` imports and calls exactly that name, passing the active `harness` from `benchmark-content.tsx`'s `filterState.harness`. `grep -rn "export function rateFor" core/price.ts` still returns only `rateForHarness`; no bare `rateFor(` exists anywhere under `apps/ayokoding-www/src`.

This landed as part of the same commit that reconciled the whole Phase 8 checklist (see the sibling F1 thread), since N-10's own line needed both fixes at once.

PR #122, opened 2026-07-29
feat(ayokoding-www): phase 9-10 — verify, reveal, and Rule-15 retest for AI benchmark

wahidyankf:
## Summary

Phase 9-10 of `plans/in-progress/ayokoding-www-tools-ai-benchmark/delivery.md`:

- **Phase 9 (Verify)**: manual multi-locale/breakpoint/theme verification, AC-38 live-page contrast
  assertion (uncovered and fixed a genuine Tailwind v4 `@theme` compiler defect), and a clean
  `ui-quality-gate` strict pass (two consecutive zero-finding runs).
- **Phase 10 (Reveal)**: removes the link gate — adds the Tools index list item and footer Tools-column
  link so the AI Model Benchmark page is reachable via site navigation.
- **Phase 10 (Rule-15 retest)**: ran all three live-site testers (`web-design-tester`,
  `web-exploratory-tester`, `web-usability-tester`) against the revealed page. Fixed **every**
  EWT/UWT/DWT defect finding (3 DWT + 4 EWT + 6 UWT = 13 total) plus implemented the SG-001 spec-gap
  and both USS-001/USS-002 spec-suggestions as real Gherkin scenarios with real step bindings.

## Notable fixes

- Capability chart's low-coverage marker text was clipped by the SVG's own viewBox (DWT-001).
- Evidence-grade dots/markers used raw Tailwind palette classes instead of theme tokens (DWT-002).
- `model-table.tsx` rebuilt on `libs/web-ui`'s shared `Table` primitives (DWT-003).
- Rapid harness+class filter changes dropped one of the two changes due to a stale-closure race
  (EWT-003) — redesigned the filter `onChange` contract to a patch merged against a
  synchronously-updated ref.
- Locale switch silently dropped active filter query params (EWT-002).
- Global search never surfaced either Tools page (UWT-001) — added a `staticSearchDocs()` module;
  verifying the fix in a real browser also surfaced and fixed a pre-existing cmdk client-filter
  regression (paired USS-001).
- Added an always-visible legend defining the 4 capability classes, 5 evidence grades, and the
  coverage formula (UWT-002/003/005, USS-002).

## Test plan

- [x] `npx nx run ayokoding-www:test:unit` exits 0 (3103 passed, 6 skipped)
- [x] `npx nx run ayokoding-www:specs:behavior:coverage` exits 0 (332 scenarios, all covered)
[...]

PR #125, opened 2026-07-30
feat(ayokoding-www): merge AI benchmark capability and price charts

wahidyankf:
Phase 1: core/sort.ts + url-state.ts sort params. See plans/in-progress/ayokoding-www-ai-benchmark-merged-chart/


apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.tsx
```diff
@@ -0,0 +1,375 @@
...
+import { formatCoverage, formatIndex, formatPriceUsd } from "./format";
+import { Axis, Bar, BandGroup, bandLabel, scaleLinear, type ChartBand } from "./chart-primitives";
+import { FilterSelect, type FilterOption } from "./benchmark-filters";
+
+const SLOT = "benchmark-chart";
+
+// ─── Layout constants (display-only — never a benchmark score, price, or threshold) ───────────
+const SVG_WIDTH = 640;
+const PLOT_X = 180; // left gutter reserved for the md/lg left-gutter label
+const PLOT_WIDTH = 380;
```
wahidyankf: **CRITICAL** | confidence 96 | integrity (escalated from HIGH)

### The DWT-001 low-coverage marker clipping defect has shipped again, and its guard test was deleted in the same PR

**This is reproduced, not predicted — it is visible in this PR's own committed evidence.**

In `plans/in-progress/ayokoding-www-ai-benchmark-merged-chart/evidence/phase-6-benchmark-chart-id-1280px.png`, the row for `GPT-5.6 Terra — 95,1` renders its marker as:

> `cakupan rendah (20`

The trailing `%)` is **clipped off the SVG edge**. Every other low-coverage marker in that image renders complete (`cakupan rendah (20%)`, `cakupan rendah (45%)`, `cakupan rendah (25%)`) because their bars are shorter. The artifact committed as proof the merged chart renders correctly contains the defect.

**Why it regressed.** The retired `capability-chart.tsx` derived plot width *from* the reserved margin, making clipping structurally impossible at any index:

```ts
export const MARKER_MIN_MARGIN =
  MARKER_GAP + Math.ceil(WORST_CASE_MARKER_LENGTH * MARKER_FONT_SIZE * MARKER_CHAR_WIDTH_RATIO) + MARKER_SAFETY_BUFFER; // = 164
const PLOT_WIDTH = SVG_WIDTH - PLOT_X - MARKER_MIN_MARGIN; // 600 - 160 - 164 = 276
```

This file inverts that dependency and hardcodes all three constants (lines 31-33), so the margin is now whatever falls out:

| | old | new |
|---|---|---|
| `SVG_WIDTH` | 600 | 640 |
| `PLOT_X` | 160 | 180 |
| `PLOT_WIDTH` | 276 (derived) | 380 (hardcoded) |
| **right margin** | **164** | **80** |

80px is below the old empirically-verified 140px floor and less than half the computed 164.

**The arithmetic predicts exactly what the screenshot shows.** Marker starts at `PLOT_X + capWidth + 6`; `GPT-5.6 Terra` has index 95.1, so `capWidth = 95.1/100 * 380 = 361.4` and the text starts at x=547.4. `cakupan rendah (20%)` is 20 chars: `20 * 9 * 0.62 = 111.6` units, ending at x=659 against a `viewBox` width of 640 — an overflow of ~19 units, or ~3 characters. Observed truncation is 2 characters (`%)`). Prediction and observation agree.

**Fix**: restore the derived relationship — reserve the margin first and compute `PLOT_WIDTH` from it — rather than hardcoding `PLOT_WIDTH` and letting the margin fall out. Re-derive `MARKER_MIN_MARGIN` for the new font/layout, re-establish a guard test in `benchmark-chart.test.tsx`, and regenerate the Phase 6 evidence images (the current ones document a defect).

Per the [Regression Test Mandate](../repo-governance/development/quality/regression-test-mandate.md), the reproducing test lands in the same commit as the fix.

apps/ayokoding-www/src/features/ai-benchmark/core/sort.ts
```diff
@@ -0,0 +1,75 @@
...
+ * — the SAME rate the price bars themselves plot (DD-8), so sort order and displayed value never
+ * diverge. A model with no metered rate at all (subscription-only or unpriced) reads as
+ * `fallback` for both fields — callers pass `Infinity` (ascending: sorts last) or `-Infinity`
+ * (descending: sorts last), so an unmetered model always sorts to the end regardless of
+ * direction, never a false "cheapest" or "priciest" ranking (invariant 10: a subscription is
+ * never treated as a numeric price). Shared by both {@link byPriceAsc} and {@link byPriceDesc} so
+ * the comparator body is not duplicated.
+ */
+function meteredRateOrFallback(s: ModelScore, fallback: number): { output: number; input: number } {
+  const rate = lowestRate(s.model);
```
wahidyankf: **HIGH** | confidence 95 | correctness (architecture + logic, deduplicated)

### Price sort ignores the harness filter, so an ascending sort renders visibly out of order

`meteredRateOrFallback` calls `lowestRate(s.model)` unconditionally — it has no way to know a harness filter is active. But `benchmark-chart.tsx:337` renders each row with `rateForHarness(score.model, harness)` when `harness` is set. Sort order and displayed price are computed from **different numbers** whenever a harness filter is active and a model's cheapest harness is not the selected one.

This file's own docstring (lines 48-53) asserts the invariant that is being broken:

> "the SAME rate the price bars themselves plot (DD-8), so sort order and displayed value never diverge"

That claim is currently false.

**Reproduction** — the fixtures already exist in `benchmark-chart.test.tsx`. `dual-harness-model` (lines 149-159) is priced `claude-code: out 10` / `codex-cli: out 25`, and the sort test at lines 199-229 proves `sortBand` orders by output rate. Combine them:

- Render with `harness="codex-cli"` and `light: "price-asc"`, alongside a model priced at a flat `out 15`.
- **Displayed**: `dual-harness-model` shows $25, the other shows $15.
- **Sorted by**: `lowestRate` → 10 vs 15, so `dual-harness-model` sorts **first**.
- **Result**: an ascending price sort renders $25 above $15.

Note `priceAxisMaxOf` (`benchmark-chart.tsx:200-211`) *does* correctly thread `harness` — so the axis scale, the bar widths, and the labels all honour the filter, and only the ordering does not. That inconsistency is what makes it user-visible rather than merely internally inconsistent.

**No test combines a harness filter with a price sort** — the two behaviours are tested in separate `describe` blocks that never intersect, which is exactly why this escaped.

**Fix**: thread `harness` through `sortBand` → `computeLayout` → both comparators so the comparator uses `rateForHarness` when a harness is active, and add the missing harness x sort test.

apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.tsx
```diff
@@ -0,0 +1,375 @@
...
+      {onSortChange ? (
+        <div data-testid={`${SLOT}-sort-controls`} className="mb-3 flex flex-wrap gap-3">
+          {bands.map((bandLayout) => (
+            <FilterSelect
+              key={bandLayout.band}
+              id={`${SLOT}-sort-${bandLayout.band}`}
+              label={`${sortLabelPrefix} — ${bandLayout.label}`}
+              value={sortState[bandLayout.band]}
+              options={sortOptions}
+              allLabel={sortModeLabel("capability", locale)}
```
wahidyankf: **HIGH** | confidence 94 | correctness (architecture + logic, deduplicated)

### Reusing `FilterSelect` for the sort dropdown renders a duplicate "Capability" option and casts `""` to `SortMode`

`FilterSelect` is built for *filters*, where an empty value legitimately means "no filter on this axis" — it therefore **always** renders `<option value="">{allLabel}</option>` (`benchmark-filters.tsx:65`) before the caller's options. Sort has no "no sort" state, so passing `allLabel={sortModeLabel("capability", locale)}` produces four options where three exist:

| value | label |
|---|---|
| `""` | Capability (high to low) |
| `capability` | Capability (high to low) |
| `price-asc` | Price (low to high) |
| `price-desc` | Price (high to low) |

Two identical labels, one of which is not a valid `SortMode`. This is visible in the committed evidence — the dropdowns in `phase-6-benchmark-chart-en-1280px.png` read `Capability (high to low)`.

Selecting the blank option runs `onChange(value as SortMode)` (line 296) with `value === ""` — an **invalid cast**, since `""` is not in `SORT_MODES`.

**It leaks into the URL before it self-heals.** `handleSortChange` (`benchmark-content.tsx:76-81`) puts `""` into `SortState` and calls `encodeState`. In `url-state.ts:136-141`:

```ts
const opus = state.opus ?? DEFAULT_SORT_MODE;   // "" is not undefined — ?? does NOT substitute
if (opus !== DEFAULT_SORT_MODE) {                // "" !== "capability" → true
  params.set(SORT_PARAM_KEYS.opus, opus);        // emits sortOpus= (empty)
}
```

The user gets `?sortOpus=` — violating `encodeState`'s own documented "OMITTING defaults so a clean, shareable URL is produced" contract (lines 121-127). `decodeState` sanitizes it back to `"capability"` on the next pass, so behaviour recovers, but the shared/bookmarked URL is already polluted and the type system was bypassed to get there.

**Fix**: give the sort control its own select without the empty option, or extend `FilterSelect` with an `allowEmpty` flag. Either way the `as SortMode` cast should go — narrow with `isKnownSortMode` (already exported from `core/sort.ts`) instead.

apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.test.tsx
```diff
@@ -0,0 +1,401 @@
...
+    id,
+    name: id,
+    vendor: "Test",
+    harnesses: ["claude-code"],
+    figures: [{ benchmark: "swe-bench-verified", value: 70, grade: "verified", source: SRC }],
+    pricing: { "claude-code": metered(input, output) },
+  };
+}
+
+describe("BenchmarkChart — merged row structure", () => {
```
wahidyankf, line 52: **HIGH** | confidence 92 | integrity / test-gaming (2 findings class-merged)

### Two regression guards were deleted with the retired chart tests and never re-established here

Deleting `capability-chart.test.tsx` and `price-chart.test.tsx` removed guards for **two bug classes that already escaped to production once each**. `benchmark-chart.test.tsx` replaces neither, so both classes are now unprotected while the code they guarded is still live.

**1. `computeGroups(dataset, fullDataset)` band-collapse (originally PR #118 cycle 2, HIGH).** The deleted file carried a three-test block:

```
describe("CapabilityChart — fullDataset keeps a harness-filtered survivor in its full-roster band")
  it("sanity: {harness} excludes both anchor models but still exposes an opus/sonnet survivor")
  it("renders the surviving model under its correct full-roster band when fullDataset is passed")
  it("WITHOUT fullDataset, the bug would reproduce — the survivor's own band collapses to light")
```

That third test is a **negative control** — it proves the guard would actually fail if the bug returned. `benchmark-chart.tsx:261` calls `computeGroups(dataset, fullDataset)` identically, so the class applies unchanged.

I checked whether the rewritten `chart-order-parity.test.tsx` covers it: **it does not.** Every render there passes `fullDataset={dataset}` (lines 72, 87, 101) — the same object for both arguments, which is precisely the case where the bug is invisible. No test in this PR exercises `dataset !== fullDataset`.

**2. Axis-max label right-alignment (originally PR #122 cycle 3).** The deleted `describe("CapabilityChart — axis-maximum label right-alignment")` asserted `chart-axis-max`'s `x` is `PLOT_X + PLOT_WIDTH`, not `SVG_WIDTH`. `benchmark-chart.tsx:313` currently passes `width={PLOT_X + PLOT_WIDTH}` — **correct today**, but nothing asserts it. A future revert to `SVG_WIDTH` would pass the entire suite silently.

**Fix**: port both blocks into `benchmark-chart.test.tsx`, keeping the negative-control test in the first — a guard without one does not prove it can fail.

(The third deleted guard, `MARKER_MIN_MARGIN`, is filed separately as CRITICAL because that one is not merely unprotected — it has already regressed.)

apps/ayokoding-www/src/features/ai-benchmark/core/url-state.ts
```diff
@@ -9,40 +9,102 @@
...
+/**
+ * The four per-band sort query-string parameter keys (DD-4). Kept as their own map, distinct from
+ * {@link PARAM_KEYS}, so `SORT_PARAM_KEYS` can be iterated over uniformly (all four bands share
+ * the same shape) without mixing in the two unrelated filter keys.
+ */
+export const SORT_PARAM_KEYS = {
+  opus: "sortOpus",
+  sonnet: "sortSonnet",
+  light: "sortLight",
+  unrated: "sortUnrated",
```
wahidyankf: **MEDIUM** | confidence 90 | architecture

### `sortUnrated` is a fully round-tripping public URL parameter with zero rendering effect

`sortUnrated` is declared here, sanitized (`sanitizeState`, line 80), decoded (line 107), and encoded (lines 149-151) — a complete, permanent public URL contract. It changes nothing.

The unrated band cannot be sorted by construction:

- `RATED_BANDS` excludes it (`benchmark-chart.tsx:43`), so `computeLayout` never sorts it (line 70).
- The unrated section renders `groups.unrated` directly and unsorted (lines 344-372).
- The sort dropdowns are generated from `bands` (line 288), which contains only rated bands — so **no dropdown for unrated is ever rendered**.

Confirmed in this PR's own evidence: `phase-6-benchmark-chart-en-1280px.png` shows exactly three sort dropdowns — `Sort — Opus`, `Sort — Sonnet`, `Sort — Light`. There is no unrated control.

The plan documents this as a real capability: `prd.md:8` says each band is "independently sortable", and DD-4 / PS-4 both enumerate `sortUnrated` alongside the three working params.

This matters more than dead code normally would, because a URL parameter is a **public contract** — once shipped it appears in shared links and must be honoured or deliberately retired later.

**Fix**: drop `sortUnrated` from `SORT_PARAM_KEYS`, `SortState`, and the sanitize/decode/encode paths, and correct the DD-4 / PS-4 claims — or, if per-band sorting of the unrated list is genuinely wanted, implement it. Do not ship an inert parameter.

apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.tsx
```diff
@@ -0,0 +1,375 @@
...
+   * The active harness filter (DD-8) — mirrors the retired price chart's existing `harness` prop
+   * exactly. When set, every price bar shows THAT harness's own rate instead of each model's
+   * lowest available harness rate.
+   */
+  harness?: HarnessId;
+  /** Reports a per-band sort change (DD-4) — the caller owns merging it into its own `sortState`. */
+  onSortChange?: (band: ChartBand, mode: SortMode) => void;
+};
+
+const DEFAULT_SORT_STATE: SortState = {
```
wahidyankf: **MEDIUM** | confidence 92 | architecture

### `DEFAULT_SORT_STATE` is re-declared instead of imported from its single source of truth

`core/url-state.ts:44-49` already exports exactly this value:

```ts
export const DEFAULT_SORT_STATE: SortState = {
  opus: DEFAULT_SORT_MODE, sonnet: DEFAULT_SORT_MODE, light: DEFAULT_SORT_MODE, unrated: DEFAULT_SORT_MODE,
};
```

This file re-declares it as a second literal, hardcoding `"capability"` four times rather than deriving from `DEFAULT_SORT_MODE`. The import is already open — line 23 imports `type SortState` from the same module, so this costs one word.

Two copies drift: changing the default sort mode in `url-state.ts` would silently leave this component's fallback on `"capability"`, so a URL-less render and a default-URL render would disagree.

This also contradicts the deliberate single-source-of-truth discipline the surrounding code follows — `RATED_BANDS` derives from `BANDS` "rather than re-declared here" (line 42), and `SORT_MODES` exists so "a new sort mode is added in exactly one place" (`sort.ts:18-21`).

**Fix**: `import { DEFAULT_SORT_STATE } from "../core/url-state"` and delete the local literal.

specs/apps/ayokoding/behavior/ayokoding-www/gherkin/tools/ai-benchmark.feature
```diff
@@ -176,66 +179,147 @@ Feature: AI model benchmark tool
...
+    Then the row shows one capability bar, one price-in bar, and one price-out bar
+    And all three bars appear stacked within that single row, not in separate chart sections
+
+  # AC-40
+  @unit
+  Scenario: Bar length is proportional to its own value
+    Given a model with a composite index of 85.7 and an output rate of $15.00
+    When the merged chart renders that model's row
+    Then the capability bar's length is proportional to 85.7 over the composite index max
+    And the price-out bar's length is proportional to $15.00 over that band's price axis max
```
wahidyankf: **MEDIUM** | confidence 88 | spec conformance (re-categorized from correctness)

### AC-40 specifies a per-band price axis; the implementation deliberately uses one shared axis

This step reads:

> `And the price-out bar's length is proportional to $15.00 over that band's price axis max`

`priceAxisMaxOf` (`benchmark-chart.tsx:200-211`) iterates **every** band and returns a single maximum, feeding one `priceScale` shared across all rows (line 265). There is no per-band price axis max.

**Re-categorization note.** The logic specialist filed this as a HIGH correctness defect. I am filing it MEDIUM as a spec-conformance gap instead, because the implementation is not accidental — `delivery.md:331-334` names the choice explicitly:

> "one over a shared price axis max across all rated bands (mirroring `price-chart.tsx`'s single-axis `axisMaxOf` pattern, ported as `priceAxisMaxOf`)"

A shared axis is also the defensible design: per-band axes would make a $5 bar in one band render the same width as a $50 bar in another, defeating cross-band price comparison. So the code is almost certainly right and **the scenario text is what is stale**.

It still has to be reconciled. This scenario is `@unit`-tagged with a live step binding, so the suite currently passes green against wording the code does not implement — a spec that cannot fail is not a spec.

**Fix**: reword AC-40 to "over the chart's shared price axis max" and note the cross-band-comparability rationale. If per-band axes were actually intended, that is a code change instead — but `delivery.md` indicates they were not.

apps/ayokoding-www/src/features/ai-benchmark/shell/chart-primitives.tsx
```diff
@@ -75,8 +75,8 @@ export function bandSwatchClass(band: ChartBand): string {
 
 /**
  * A band's localized class-name label — the single place either chart looks up
- * `BAND_LABEL_KEYS[band] → t(locale, key)` (Y-11 refactor: `capability-chart.tsx` and
- * `price-chart.tsx` each carried their own identical copy of this lookup; the fallback-to-band-id
+ * `BAND_LABEL_KEYS[band] → t(locale, key)` (Y-11 refactor: the two retired charts each carried
+ * their own identical copy of this lookup; the fallback-to-band-id
  * guard can no longer drift between the two charts now that it lives here once).
```
wahidyankf: **MEDIUM** | confidence 93 | docs (4 findings class-merged and expanded)

### Stale two-chart terminology and dead file references survive across the codebase

The docs specialist reported 4 sites. **My own sweep found 7 more it did not name.** Reporting this as one class-level finding so it is fixed as a class — fixing only the originally-named sites would leave the rest.

**Stale present-tense claims that two charts exist:**

| Site | Text |
|---|---|
| `chart-primitives.tsx:3` | "The capability chart and the price chart **share** ONE set of SVG building blocks" |
| `chart-primitives.tsx:5` | "anything the two charts **still duplicate** back here" |
| `chart-primitives.tsx:80` (this line) | "can no longer drift between the two charts **now that it lives here once**" |
| `core/bands.ts:28` | "the unit **both charts consume**" |
| `core/bands.ts:110-111` | "**Both charts consume** the SAME per-band list ... in the capability chart and the price chart (AC-11)" |
| `core/bands.unit.test.ts:143` | "**Both charts**" |
| `shell/benchmark-filters.tsx:12` | "the same dual-render pattern ... **the two charts already use**" |
| `benchmark-content.tsx:105` | "an explicit empty-state message replaces **both charts**" |

`chart-primitives.tsx` is the sharpest case: this same PR correctly rewrote four docstrings in this file to past tense ("the two **retired** charts", lines 78-79) while leaving the module header at line 3 in the present tense — the file now contradicts itself.

**Dead file references naming a deleted file:**

- `apps/ayokoding-www-fe-e2e/src/steps/ai-benchmark.steps.ts:166` — "`capability-chart.tsx` never plots `unrated` as a bar". That file was deleted in `01a03b972`. Three other tags in this same file were correctly updated by this PR.
- `ai-benchmark.steps.ts:157` — "neither chart currently renders `-wash`"

**Stale "(charts)" plural / "neither chart" wording:**

- `specs/.../ai-benchmark.feature:390` + its `# AC-28 (charts)` header comment
- `apps/ayokoding-www/test/unit/fe-steps/ai-benchmark.steps.tsx:1526`, and `:1530` ("alongside **the two charts**")

The Phase 4 gate grep missed the `neither chart` phrasing, which is why the class survived a phase explicitly dedicated to this rewrite.

**Fix**: sweep for `both charts`, `two charts`, `neither chart`, `capability-chart`, `price-chart`, and `capability chart and the price chart` across `apps/` and `specs/`, and give a per-file verdict. Genuinely historical past-tense references (e.g. `chart-primitives.test.tsx:9`, "hoisted here from both charts' identical...") are **correct** and should be left alone — the target is present-tense claims and dead file names. Ignore `coverage/` and `.features-gen/` hits; they regenerate.

apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.tsx
```diff
@@ -0,0 +1,375 @@
...
+          data-slot="chart-subscription-label"
+          data-testid={`${SLOT}-subscription-${id}`}
+          x={PLOT_X}
+          y={priceInY + BAR_HEIGHT}
+          className="fill-muted-foreground text-[10px]"
+        >
+          {t(locale, "aiBenchSubscription")} ({formatPriceUsd(rate.planCostUsd, locale)})
+        </text>
+      ) : (
+        <text
```
wahidyankf, line 253: **MEDIUM** | confidence 85 | correctness / spec coverage

### The new "not reported" row branch is user-facing behaviour with no owning Gherkin scenario and no unit test

This `else` branch renders `aiBenchNoFigure` for a **rated** model with no metered rate and no subscription. It is genuinely new behaviour: the retired `price-chart.tsx` omitted such models from the plot entirely, so nothing was displayed for them.

It is live — `phase-6-benchmark-chart-en-1280px.png` shows `Not reported` under `Gemini 3.1 Pro — 92.7` and `Kimi K2.6 — 78.3` (and `Tidak dilaporkan` in the `id` render).

**No spec owns it.** AC-39 through AC-47 cover the merged row, bar proportionality, per-band sort, subscription rendering, and the unrated list — none covers a rated model with no price at all. It appears only in `tech-docs.md`'s flowchart, which is design documentation, not a behavioural contract.

**No test covers it either.** `benchmark-chart.test.tsx` has a case for a *rated + subscription-only* model (line 237) and one for a *genuinely priceless unrated* model (line 350) — but nothing renders a rated model that falls through both, so the `benchmark-chart-not-reported-*` testid is never asserted.

Per [Specs & Gherkin Completeness](../repo-governance/development/quality/feature-change-completeness.md), code under `apps/` does not land without companion Gherkin — and this is new user-facing rendering, not a refactor, so the exemption does not apply.

**Fix**: add a scenario covering a rated model with no reported price, and a unit test fixture asserting the `not-reported` label renders while its capability bar still does.

wahidyankf (commented):

**Risk tier**: full
**Security-sensitive-path override applied**: no (tier is `full` on size alone — 2410 insertions / 1406 deletions across 25 files)
**Diff coverage**: full diff reviewed in one pass (no generated-file exclusion; `.features-gen/` and `coverage/` artifacts were read but findings there are reported against their source, not the artifact)
**Prior-cycle human dismissals respected**: none this cycle (cycle 1 — no prior threads)

---

### Synthesis summary

15 raw findings from 8 specialists collapsed to **9** consolidated findings. Governance, security, performance, and instruction-decay returned clean, and performance's suppression of the per-band re-render cost was correct (no growth-rate evidence, ~39 models).

Coordination actions taken:

- **Deduplicated** 2 pairs: architecture+logic both found the harness/sort divergence; architecture+logic both found the duplicate-option cast. Independent rediscovery is confirmation, not two findings.
- **Class-merged** the 4 docs findings into 1 — the stale two-chart terminology is a *class*, and my own sweep found **7 more sites the docs specialist did not name** (`bands.ts`, `benchmark-filters.tsx`, `fe-steps`). Fixing only the 4 named sites would leave the class alive.
- **Class-merged** 2 integrity findings into 1 (deleted regression guards not re-established).
- **Re-categorized** the shared-price-axis finding from correctness to spec-conformance (MEDIUM, not HIGH): the shared axis is a *deliberate, documented* design decision (`delivery.md`), so the defect is the stale Gherkin wording, not the code. This is the architecture-vs-correctness boundary call — the implementation is a settled structural choice, so the burden falls on the spec text.
- **Escalated** the low-coverage marker margin finding from HIGH to **CRITICAL**. The integrity specialist recommended a Playwright check "before merge — if it reproduces clipping, this escalates to a live shipping defect." I ran that verification against this PR's own committed Phase 6 evidence: **it reproduces.**
- **Severity split** on the harness/sort divergence: logic said CRITICAL, architecture said HIGH. Settled at **HIGH** — it is a blocking user-facing correctness defect, but it does not block deployment the way a render defect does.

No findings were dropped by the reasonableness filter this cycle; all 15 raw findings verified as real against source.

**This review is blocking** despite its `COMMENT` state — `gh` authenticates as the PR author, so `REQUEST_CHANGES` is structurally unavailable. Blocking status lives in the severity labels: **1 CRITICAL + 3 HIGH must be resolved before merge.**

apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.tsx
wahidyankf: Fixed: restored the derived-margin relationship in `benchmark-chart.tsx` — `PLOT_WIDTH` is no
longer a hardcoded `380`. It is now computed as `SVG_WIDTH - PLOT_X - MARKER_MIN_MARGIN`, where
`MARKER_MIN_MARGIN` is re-derived from the same locale-aware formula the retired
`capability-chart.tsx` used (`MARKER_GAP + ceil(WORST_CASE_MARKER_LENGTH * MARKER_FONT_SIZE *
MARKER_CHAR_WIDTH_RATIO) + MARKER_SAFETY_BUFFER`), computing to `164` — restoring the right margin
from the regressed `80` back above the `140`-unit clip floor.

Added a regression guard (`benchmark-chart.test.tsx`'s "DWT-001 right-margin regression" block):
locks `MARKER_MIN_MARGIN` to `164`, floors it at `>= 140`, includes a negative control proving the
pre-fix hardcoded geometry (margin `80`) would fail this guard, and a live-DOM reproduction of the
exact `GPT-5.6 Terra` (index 95.1, `id` locale) case using `getBBox()` to assert the marker's own
rendered right edge never crosses `SVG_WIDTH`.

Regenerated and verified all 6 Phase 6 evidence screenshots (`en`/`id` × 375/768/1280px) — Playwright
MCP wasn't available in this invocation, so I drove headless Chromium directly via the `playwright`
package (same navigate/resize/screenshot/console-check steps as the original Phase 6 pass), and
additionally asserted programmatically via `getBBox()` that all 7 low-coverage markers per render
stay within the SVG's own `viewBox` width. 0 markers clipped across all 6 renders, 0 console
errors/warnings. The `id`/1280px capture now shows `GPT-5.6 Terra — 95,1` rendering the complete
`cakupan rendah (20%)` marker. All 6 PNGs overwritten in place (same filenames); `delivery.md`'s
Phase 6 evidence section documents the regeneration and the live verification method.

apps/ayokoding-www/src/features/ai-benchmark/core/sort.ts
wahidyankf: Fixed: threaded the active `harness` filter through `core/sort.ts`'s price comparators.
`meteredRateOrFallback` now takes an optional `harness` parameter and calls `rateForHarness(s.model,
harness)` when one is active, `lowestRate(s.model)` otherwise — matching exactly what
`benchmark-chart.tsx:337` renders per row. `byPriceAsc`/`byPriceDesc` both gained the same optional
third parameter (safe for `Array.prototype.sort`, which only ever calls a comparator with 2
arguments, so every existing `.sort(byPriceAsc)` call site is unaffected). `sortBand` →
`computeLayout` in `benchmark-chart.tsx` now thread `harness` down to the comparators.

Added the missing harness × sort regression coverage at both levels: a pure-function test in
`sort.unit.test.ts` (`byPriceAsc`/`byPriceDesc` sort by the selected harness's own rate, not the
model's lowest-available rate) and a component-level test in `benchmark-chart.test.tsx`
(`"BenchmarkChart — price sort honours an active harness filter (DD-8)"`) reproducing your exact
scenario: a `dual-harness-model` priced `claude-code: out 10` / `codex-cli: out 25` alongside a
flat-priced model at `out 15`, rendered with `harness="codex-cli"` and `light: "price-asc"` —
asserts the flat-priced model now sorts first (ascending 15 before 25) and that the displayed price
bar really is the codex-cli rate, not the lowest.

apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.tsx
wahidyankf: Fixed, using your suggested direction: `FilterSelect`'s `allLabel` prop is now optional (documented
as omit-it-for-controls-with-no-"no-filter"-state), and it only renders the empty `<option
value="">` when a caller passes one. The sort control in `benchmark-chart.tsx` no longer passes
`allLabel` at all, so no duplicate "Capability" option renders. `FilterSelect`'s callers that DO
need the empty "no filter" option (harness/class in `benchmark-filters.tsx`) are unaffected — they
still pass `allLabel` and keep their existing behaviour.

The `onChange` handler no longer does `value as SortMode` — it narrows with the already-exported
`isKnownSortMode(value)` from `core/sort.ts` and only calls `onSortChange` when the value is a known
`SortMode`, closing the URL-pollution path you traced through `handleSortChange` →
`encodeState`/`?sortOpus=`.

Added a regression test (`"BenchmarkChart — sort dropdown has no duplicate/invalid empty option"`)
asserting the sort `<select>` renders exactly the three `SortMode` values (`capability`,
`price-asc`, `price-desc`) with no blank-value option.

apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.test.tsx
wahidyankf: Fixed: ported both deleted guard blocks from the retired `capability-chart.test.tsx` into
`benchmark-chart.test.tsx`, adapted to the merged component.

1. **fullDataset band-collapse** (`"BenchmarkChart — fullDataset keeps a harness-filtered survivor in
   its full-roster band"`): all three sub-tests ported, including the negative-control third test
   ("WITHOUT fullDataset, the bug would reproduce") that proves the guard can actually fail — using
   the real `dataset`/`OPUS_ANCHOR_ID`/`SONNET_ANCHOR_ID` and a `codex-cli` filter exactly as the
   original did, retargeted at `benchmark-chart-band-*`/`benchmark-chart-row-*` testids.
2. **Axis-maximum label right-alignment** (`"BenchmarkChart — axis-maximum label right-alignment"`):
   ported, asserting `chart-axis-max`'s `x` equals `PLOT_X + PLOT_WIDTH`, not `SVG_WIDTH` — reverting
   `benchmark-chart.tsx`'s `<Axis width={...}>` call back to `SVG_WIDTH` now fails this test.

Both required exporting `PLOT_WIDTH`/`PLOT_X`/`SVG_WIDTH`/`MARKER_MIN_MARGIN` from
`benchmark-chart.tsx` (previously all module-private), which also enabled the CRITICAL finding's own
regression guard in the same file.

apps/ayokoding-www/src/features/ai-benchmark/core/url-state.ts
wahidyankf: Fixed, taking the "drop it" branch of your suggested fix rather than "implement it": removed
`sortUnrated` entirely — from `SORT_PARAM_KEYS`, `SortState`, `UntrustedSortState`, and the
`sanitizeState`/`decodeState`/`encodeState` paths in `core/url-state.ts` — rather than wiring up real
per-band sorting for the `unrated` list.

Reasoning for choosing removal over implementation: implementing genuine unrated-band sorting would
require a new sort-control UI for a band that currently has none, plus its own design decision on
whether/how an unrated (no composite index) list should even be orderable — that's new scope
requiring a product decision, not a fix to this defect. Dropping the dead parameter is the
mechanical, low-risk default consistent with "do not ship an inert parameter."

Updated the round-trip test in `url-state.unit.test.ts` (removed the `unrated` field from the
4-band fixture, now asserting the three RATED bands only) and added a new test confirming a stale
bookmarked `?sortUnrated=...` URL is now silently ignored (decodes to `DEFAULT_STATE`, encodes back
to an empty query string) rather than resurrected. Also corrected `prd.md`'s PS-4 and its "each band
independently sortable" claim, and `tech-docs.md`'s DD-4, to state this explicitly, and updated every
`SortState`-shaped fixture across `benchmark-chart.tsx`/`benchmark-chart.test.tsx`/
`chart-order-parity.test.tsx`/`benchmark-content.tsx` to drop the now-nonexistent `unrated` key.

apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.tsx
wahidyankf: Fixed exactly as suggested: `benchmark-chart.tsx`'s local `DEFAULT_SORT_STATE` literal is deleted;
the component now imports `DEFAULT_SORT_STATE` from `../core/url-state` (the import statement
already pulled `type SortState` from that module, so this is a one-line change) and uses it as the
`sortState` prop's default. The two copies can no longer drift.

specs/apps/ayokoding/behavior/ayokoding-www/gherkin/tools/ai-benchmark.feature
wahidyankf: Fixed exactly as suggested: agree the code is right and the scenario text was stale. Reworded
AC-40's step in `ai-benchmark.feature` from "over that band's price axis max" to "over the chart's
shared price axis max", added a comment citing `delivery.md`'s own Phase 6 GREEN note ("one over a
shared price axis max across all rated bands ... ported as priceAxisMaxOf") plus the
cross-band-comparability rationale you gave, and updated the matching step binding text in
`apps/ayokoding-www/test/unit/fe-steps/ai-benchmark.steps.tsx` (and its inline comment) to match.
`specs:behavior:coverage` confirms the reworded step is still covered.

apps/ayokoding-www/src/features/ai-benchmark/shell/chart-primitives.tsx
wahidyankf: Fixed as a class, per-file verdict below (all 11 sites you named — the 4 the docs specialist found
plus your own 7 — plus a handful my own sweep of the same grep patterns turned up in files neither
of you named):

- `chart-primitives.tsx:3,5` (module header) — rewritten to present-tense-correct: `benchmark-chart.tsx`
  is now named as the sole consumer, with the retired two-chart history moved to explicit past tense.
- `chart-primitives.tsx:80` (Y-11 refactor comment) — reworded to "cannot drift between duplicate
  copies now that it lives here once (previously, the two retired charts each carried an identical
  copy)".
- `core/bands.ts:28` — "the unit both charts consume" → "the unit the chart shell (`benchmark-chart.tsx`)
  consumes".
- `core/bands.ts:110-111` — reworded to name the merged chart as the present-tense consumer, with the
  cross-chart-parity history stated as historical.
- `core/bands.unit.test.ts:143` (my own sweep) — "Both charts consume" → "The merged chart consumes".
- `shell/benchmark-filters.tsx:12` — dropped "and the two charts already use" (the merged chart
  doesn't even use this dual-render pattern — only `model-table.tsx` does); now names `model-table.tsx`
  alone.
- `benchmark-content.tsx:33,105,110` (my own sweep) — "own `opus`/`sonnet`/`light`/`unrated` key" →
  dropped `unrated` (see the sortUnrated fix below); "replaces both charts" → "replaces the chart";
  "neither chart nor the data table" → "the chart and the data table do not render" (the live Gherkin
  step + its binding were updated in lockstep, see below).
- `ai-benchmark.steps.ts` (fe-e2e, my own sweep) lines 157, 166 — "neither chart currently renders
  `-wash`" → "the chart currently renders no `-wash`"; dead `capability-chart.tsx` reference → `benchmark-chart.tsx`.
- `ai-benchmark.feature:384-390` + its step binding in `test/unit/fe-steps/ai-benchmark.steps.tsx` —
  "AC-28 (charts)" → "AC-28 (chart)"; the live `But` step text "neither chart nor the data table
  renders" → "the chart and the data table do not render in the empty state", updated identically in
  both the feature file and its exact-text step binding so the scenario keeps passing.
- `test/unit/fe-steps/ai-benchmark.steps.tsx:1530` — "alongside the two charts" → "alongside the
  chart".

Left alone as genuinely historical/correct, per your own carve-out: `chart-primitives.test.tsx:9`,
`chart-primitives.tsx:248` ("both charts previously carried"), and `chart-order-parity.test.tsx`'s
own docstring ("Pre-merge, this file rendered the two separate charts... proving neither chart's
rendering re-ordered...") — all already past-tense-correct.

Re-verified with the same grep sweep you specified (`both charts`, `two charts`, `neither chart`,
`capability-chart`, `price-chart`, `capability chart and the price chart` across `apps/` and
`specs/`, excluding `coverage/`/`.features-gen/`) — zero present-tense or dead-file-name hits remain.

apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.tsx
wahidyankf: Fixed: added both the owning Gherkin scenario and the unit test.

- New scenario `AC-48` in `ai-benchmark.feature` ("A rated model with no reported price shows a
  not-reported placeholder"), with its own step binding in
  `apps/ayokoding-www/test/unit/fe-steps/ai-benchmark.steps.tsx`, mirroring AC-44's structure: a
  rated (has figures), no-metered-rate, no-subscription model, asserting the capability bar still
  renders while the `not-reported` testid carries the localized `aiBenchNoFigure` text.
- A matching direct component test in `benchmark-chart.test.tsx`
  (`"BenchmarkChart — AC-48 rated model with no reported price"`) asserting the same thing plus that
  no price-in/price-out/subscription testids render for that row.

`specs:behavior:coverage` confirms the new scenario is covered (42 specs, 343 scenarios, 1235 steps
— all covered; `ai-benchmark.feature` itself grew from 48 to 49 scenarios with this addition).

plans/in-progress/ayokoding-www-ai-benchmark-merged-chart/delivery.md
```diff
@@ -236,17 +339,26 @@ archival first, merge last).
       And the price-out bar's length is proportional to $15.00 over that band's price axis max
```
wahidyankf: **HIGH** | confidence 97 (raised from the raw 95 by tool-verification) | governance

### The AC-40 embedded Gherkin still says "that band's price axis max" — three sites, not one

`pr-review-fixer`'s commit `5d4338e97` reworded the live AC-40 scenario in the companion `.feature` as part of resolving cycle 1's shared-vs-per-band price-axis finding, so the scenario now reads:

> `specs/apps/ayokoding/behavior/ayokoding-www/gherkin/tools/ai-benchmark.feature:269`
> `And the price-out bar's length is proportional to $15.00 over the chart's shared price axis max`

The plan docs were never brought along. The [Plans Organization Convention](https://github.com/wahidyankf/ose-public/blob/5d4338e97d4288bd3e6f2d5f195b1cec441d72d8/repo-governance/conventions/structure/plans.md#L476-L485) requires the inline block be "copied **verbatim** from the companion `.feature`", and marks the whole rule a **HARD RULE** flagged at **HIGH**.

**Verified stale sites — the specialist named only the first:**

| # | Site | Current text | Should read |
| --- | --- | --- | --- |
| 1 | `delivery.md:339` (embedded ` ```gherkin ` block) | `...proportional to $15.00 over that band's price axis max` | `...over the chart's shared price axis max` |
| 2 | `delivery.md:326` (same step's RED prose) | ``output / <band price axis max>`` | a shared-axis phrasing |
| 3 | **`prd.md:274`** (AC-40's canonical acceptance-criterion copy) | `...proportional to $15.00 over that band's price axis max` | `...over the chart's shared price axis max` |

Site 3 is in a file the governance specialist never opened; I found it by enumerating the class rather than the cited sites. `prd.md` is the acceptance-criteria source of record, so leaving it stale is the more consequential half of this finding — a future reader reconciling AC-40 against the code would hit the *wrong* statement of intent first.

**This is self-inconsistent within a single delivery step**, which is what makes it unambiguous rather than a judgement call: the fixer *did* update the GREEN prose four lines below the stale block —

> `delivery.md:343`
> `over `COMPOSITE_INDEX_MAX` for the capability bar, one over a shared price axis max across all`

— so lines 326 and 339 now contradict line 343 inside the same RED->GREEN cycle.

**Remediation**: update all three sites to the `.feature`'s exact wording. Site 1 must match character-for-character; sites 2 and 3 need only to stop asserting a per-band axis.

plans/done/2026-07-30__ayokoding-www-ai-benchmark-merged-chart/delivery.md
```diff
@@ -430,70 +705,150 @@ archival first, merge last).
...
       — acceptance: `grep -c "^  Scenario" specs/apps/ayokoding/behavior/ayokoding-www/gherkin/tools/ai-benchmark.feature`
       returns `48` (39 existing scenarios, 2 rewritten in place per the two items above — AC-11 and
       AC-18 — net zero change to the count — plus 9 genuinely new scenarios from `prd.md`'s
       11-scenario Acceptance criteria section, of which 2 are the in-place rewrite targets counted
       above (AC-11, AC-18) and 9 are additions: 39 + 9 = 48)
-- [ ] [AI] Implement or extend the corresponding step definitions in
+
+  > **Date**: 2026-07-30. **Status**: DONE.
+  > **Files-Changed**: `ai-benchmark.feature`.
+  > **Notes**: 9 new scenarios (AC-39..AC-47) added verbatim from prd.md, inserted after AC-37.
```
wahidyankf, line 743: **HIGH** | confidence 93 (raised from the raw 82 by tool-verification — the absence is mechanically confirmed, not inferred) | governance

### AC-48 landed in code, specs, and step bindings — but has zero traceability in `delivery.md`

Resolving cycle 1's MEDIUM "not reported" spec-coverage finding, commit `5d4338e97` added a genuinely new scenario and its full test surface:

- `specs/apps/ayokoding/behavior/ayokoding-www/gherkin/tools/ai-benchmark.feature:327` — `Scenario: A rated model with no reported price shows a not-reported placeholder`
- `apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.test.tsx:652` — `describe("BenchmarkChart — AC-48 rated model with no reported price")`
- `apps/ayokoding-www/test/unit/fe-steps/ai-benchmark.steps.tsx:1913` — the `Scenario(...)` binder

`delivery.md` records none of it. Verified mechanically at the pinned SHA:

```
$ grep -n -i -e 'AC-48' -e 'not.reported' delivery.md
916:  > rate ($3.48→$50, "Not reported" last), while Opus (`GPT-5.6 Sol, Claude Opus 5`) and Light
```

The single hit is an unrelated Phase-6 sort-order evidence note, not AC-48 traceability. There is no `**Gherkin (binds) →**` tag for the scenario, no RED/GREEN step owning it, and no **Date** / **Status** / **Files Changed** / **Notes** evidence entry.

This violates the [Feature Change Completeness Convention](https://github.com/wahidyankf/ose-public/blob/5d4338e97d4288bd3e6f2d5f195b1cec441d72d8/repo-governance/development/quality/feature-change-completeness.md#L61), which requires a plan whose scope creates observable behavior in `apps/` or `specs/` to "carry explicit delivery-checklist steps that create or update the corresponding `specs/` Gherkin `.feature` files" — reinforced at [line 203](https://github.com/wahidyankf/ose-public/blob/5d4338e97d4288bd3e6f2d5f195b1cec441d72d8/repo-governance/development/quality/feature-change-completeness.md#L203): "a plan that omits those steps is incomplete."

**Why this is a real gap and not bookkeeping pedantry**: AC-48 is not a test-only addition. It documents rendering behaviour the retired `price-chart.tsx` never had — that component omitted price-less models from the plot entirely, whereas the merged chart renders an inline placeholder. The `.feature` comment and both code comments all say so explicitly. A behaviour change this substantive being invisible in the delivery checklist is exactly the drift the convention exists to catch, and it will surface again at archival when `plan-execution-checker` reconciles scenarios against delivery steps.

This thread is anchored on the step that says "9 new scenarios (AC-39..AC-47) added verbatim from prd.md" — the scenario roster is now AC-39..**AC-48**, and this is the step that should say so.

**The fixer already knows the shape of the fix** — the same commit added two `> **Post-hoc correction (PR #125 review-cycle 1, `pr-review-fixer`):**` blocks at `delivery.md:273` and `delivery.md:932`. AC-48 needs a third, in Phase 6, carrying the `**Gherkin (binds) →** "A rated model with no reported price shows a not-reported placeholder"` tag with the scenario copied verbatim, plus the standard evidence entry.

wahidyankf (commented):
**Risk tier**: full
**Security-sensitive-path override applied**: no (tier is `full` on size alone — 2972/-1477 lines across 31 files; no `.github/`, `.env*`, git-identity, or `pr-merge-protocol.md` path in the diff)
**Diff coverage**: full diff reviewed in one pass (D13 no-exclusion posture — no generated-file filtering applied)
**Prior-cycle human dismissals respected**: none this cycle (all 9 cycle-1 threads resolved by `pr-review-fixer`'s commit `5d4338e97`; no human "won't fix" / "I disagree" dismissal on any thread)

---

## Cycle 2 of 3 — consolidated review

**2 findings, both HIGH, both governance.** Seven of eight specialists returned clean, each having independently re-verified cycle 1's fixes rather than assuming them — integrity did so by mechanically reverting each fix and confirming the ported regression tests actually fail without it. No code defect survives into this cycle.

Both surviving findings are **plan-document bookkeeping drift introduced by the fixer commit itself**, not preexisting cycle-1 misses, and not code defects. They are HIGH rather than MEDIUM because the [Plans Organization Convention](https://github.com/wahidyankf/ose-public/blob/5d4338e97d4288bd3e6f2d5f195b1cec441d72d8/repo-governance/conventions/structure/plans.md#L476-L487) states the rule as a **HARD RULE** and fixes `plan-checker`'s severity for it at HIGH explicitly: "`plan-checker` flags a multi-scenario behavior RED, or absent/non-verbatim inline Gherkin, as a **HIGH** finding."

**This review is blocking despite its `COMMENT` state.** `gh` authenticates as the PR author here, so GitHub structurally refuses `REQUEST_CHANGES` on one's own PR; blocking status rides the per-finding severity labels instead.

### Coordination notes

- **Deduplicated**: n/a — a single specialist raised both findings, at distinct sites with distinct remediations, so neither collapses into the other.
- **Re-categorized**: neither. Cycle 1 re-categorized its AC-40 finding from *correctness* to *spec conformance*; these two are neither — the code and the `.feature` are both correct now, and only the plan doc lags. That is governance. No architecture-versus-correctness boundary call arose this cycle.
- **Reasonableness-filtered**: nothing dropped — no specialist emitted a raw finding below the 80-confidence bar this cycle.
- **Tool-verified**: both findings re-read at the pinned SHA before posting, and **Finding 1 was expanded** — the specialist named one site, but the same stale phrasing survives at **three**, one of them in a different file the specialist never opened. Both raw confidences were raised as a result (95 -> 97, 82 -> 93). Detail in each thread.

plans/in-progress/ayokoding-www-ai-benchmark-merged-chart/delivery.md
wahidyankf: Fixed: reworded all 3 stale sites to match the current verbatim scenario text in `ai-benchmark.feature:269` ("over the chart's shared price axis max"):

- `prd.md:274` — the Acceptance criteria (Gherkin) block's "Bar length is proportional to its own value" scenario.
- `delivery.md:339` — the same scenario embedded in the `**Gherkin (binds) →**` tag for the AC-40 RED step.
- `delivery.md:326` — the RED step's own prose ("`output / <band price axis max>`" → "`output / <the chart's shared price axis max>`").

Verified no remaining occurrences with `grep -rn "that band's price axis max\|<band price axis max>" plans/in-progress/ayokoding-www-ai-benchmark-merged-chart/ specs/... apps/ayokoding-www/` (exit 1, zero matches). Pushed in `81f224b52`.

plans/done/2026-07-30__ayokoding-www-ai-benchmark-merged-chart/delivery.md
wahidyankf: Fixed: added the missing AC-48 traceability entry to `delivery.md` (inserted after the "9 new scenarios (AC-39..AC-47)" item, before the step-definitions item), following the file's existing Gherkin-Tagged Delivery Steps pattern:

- A `**Gherkin (binds) →**` tag with the verbatim Given/When/Then/And body of "A rated model with no reported price shows a not-reported placeholder" (matching `ai-benchmark.feature:333-337` exactly).
- A `**Date**: 2026-07-30. **Status**: DONE (post-hoc correction, PR #125 review-cycle 1, `pr-review-fixer`)` evidence blockquote naming all 3 files cycle 1's fixer touched (`ai-benchmark.feature`, `benchmark-chart.test.tsx`, `ai-benchmark.steps.tsx`) and the acceptance check (`grep -c "^  Scenario" ai-benchmark.feature` → `49`, confirmed live).
- The Notes explicitly state this is filling a gap left by cycle 1's fixer commit, consistent with the file's other post-hoc-correction notes (e.g. the Phase 1 `sortUnrated` note, the Phase 6 screenshot-regeneration note).

Pushed in `81f224b52`.

plans/done/2026-07-30__ayokoding-www-ai-benchmark-merged-chart/README.md
```diff
@@ -33,12 +33,14 @@ breakpoint (mobile, tablet, desktop) with no layout switch, avoiding Option C's
...
+- A per-band sort control (`FilterSelect`-styled native `<select>`) for each of the three RATED
+  bands (opus/sonnet/light — the `unrated` band has no composite index to sort by and never had a
+  control), three options: Capability (default, descending), Price low→high, Price high→low —
+  sorting by the **output rate**, tie-broken by input rate.
 - New pure comparator module `core/sort.ts` (`byCapabilityDesc`, `byPriceAsc`, `byPriceDesc`).
-- URL-encoded per-band sort state (`sortOpus`, `sortSonnet`, `sortLight`, `sortUnrated` query
-  params), extending `core/url-state.ts`.
+- URL-encoded per-band sort state (`sortOpus`, `sortSonnet`, `sortLight` query params), extending
+  `core/url-state.ts`. (A `sortUnrated` param existed briefly and round-tripped despite having no
+  rendering effect — removed as dead code in the PR #125 fixer cycle.)
```
wahidyankf, line 43: **MEDIUM** | confidence 90 | docs (2 documents class-merged)

### Both architecture diagrams still say "+4 sort params" — the shipped code has 3

`README.md:70` renders `U["core/url-state.ts<br/>+4 sort query params"]` and `tech-docs.md:11` renders `UrlState["core/url-state.ts<br/>+4 sort params (NEW)"]`. The shipped `SORT_PARAM_KEYS` at `apps/ayokoding-www/src/features/ai-benchmark/core/url-state.ts:30-33` declares exactly **three**: `sortOpus`, `sortSonnet`, `sortLight`. `sortUnrated` was removed as dead code by cycle 1's fixer (resolved thread on `url-state.ts`, MEDIUM/architecture).

This is an internal contradiction, not merely staleness — **both documents already say 3 correctly elsewhere in the same file**: `README.md:41-43` names the three params and explicitly records the `sortUnrated` removal, and `tech-docs.md:152` (DD-4) and `tech-docs.md:171` (File impact) both do the same. The top-of-file diagrams alone were never updated when the removal landed, so each document now asserts two different counts.

**Fix direction**: change `+4 sort query params` → `+3 sort query params` at `README.md:70`, and `+4 sort params (NEW)` → `+3 sort params (NEW)` at `tech-docs.md:11`.

> *Thread anchored to the nearest line this PR actually changed — GitHub review comments cannot anchor to unchanged lines. The finding's real sites are cited inline above.*

plans/done/2026-07-30__ayokoding-www-ai-benchmark-merged-chart/prd.md
```diff
@@ -224,16 +228,16 @@ finishing touch — it is the specific property that avoids the prior plan's Opt
...
+| PS-2 | Three bars per rated model row: capability, price-in, price-out                                                                                         |
+| PS-3 | Per-band sort control (Capability default / Price low→high / Price high→low), sorted by output rate (input tie-break)                                   |
+| PS-4 | URL-encoded per-band sort state for the three RATED bands (`sortOpus`/`sortSonnet`/`sortLight` — `unrated` is never sortable, see the correction above) |
+| PS-5 | New `core/sort.ts` pure comparator module                                                                                                               |
+| PS-6 | DD-1: rated + subscription-only model renders inline `Subscription ($cost)` text in place of price bars                                                 |
+| PS-7 | Rewritten + extended Gherkin in the existing `ai-benchmark.feature`                                                                                     |
+| PS-8 | DD-8: harness-specific price display (AC-17/AC-18) carries over unchanged via a `harness` prop on the merged chart                                      |
```
wahidyankf, line 240: **HIGH** | confidence 96 (raised from the raw 93 by tool-verification — the diff shows a runtime behavior change, not only a type widening) | docs (3 documents class-merged)

### Three documents claim `benchmark-filters.tsx` is untouched / out of scope — cycle 1's fixer changed its public component contract

The claims:

- `prd.md:246` — `| OOS-1 | Any change to \`model-table.tsx\` or \`benchmark-filters.tsx\` |`
- `README.md:54` — ``- `benchmark-filters.tsx` (harness/class filter bar) — untouched.``
- `tech-docs.md:180,182` — the `No change to:` list, which names `` `shell/benchmark-filters.tsx` ``

All three are **false as shipped**. `git diff origin/main...81f224b52 -- .../shell/benchmark-filters.tsx` shows a real change to the exported `FilterSelectProps` contract:

```diff
-  allLabel: string;
+  allLabel?: string;
```

```diff
-          <option value="">{allLabel}</option>
+          {allLabel !== undefined ? <option value="">{allLabel}</option> : null}
```

That is a **public-component-contract change plus a runtime render change** at `benchmark-filters.tsx:52` (prop type) and `:76` (conditional empty-option render) — not a comment-only edit. It was made deliberately and correctly, to resolve cycle 1's HIGH correctness finding about reusing `FilterSelect` for the sort dropdown (the sort control always has an active `SortMode`, so it must render no empty option). The change is fine; **the scope claims that outlived it are not**.

This matters beyond tidiness: `OOS-1` is the plan's normative out-of-scope declaration, and it is exactly what a future reader, auditor, or `plan-execution-checker` consults to decide whether this PR stayed inside its declared boundary. Leaving it asserting the opposite of what shipped converts a legitimate, reviewed, in-scope fix into an apparent undeclared scope breach.

For the record, I independently re-verified the two sibling "unchanged" claims in the same `tech-docs.md` sentence — `core/bands.ts` and (via DD-6 / `README.md:69`) `shell/chart-primitives.tsx`. Both diffs are comment/docstring-only. Those claims stay true and need no edit.

**Fix direction**: amend all three sites to state the actual, narrow, backward-compatible widening and why the review cycle required it — e.g. `OOS-1` scoped to "no change to `model-table.tsx`; no behavioral change to `benchmark-filters.tsx`'s own filter bar, other than the backward-compatible `allLabel?` widening of the shared `FilterSelect` required by the PR #125 cycle-1 sort-dropdown fix". Keep the wording consistent across `prd.md:246`, `README.md:54`, and `tech-docs.md:180-182`; a one-line DD or `File impact` row in `tech-docs.md` recording the widening would close the loop.

> *Thread anchored to the nearest line this PR actually changed — GitHub review comments cannot anchor to unchanged lines. The finding's real sites are cited inline above.*

plans/done/2026-07-30__ayokoding-www-ai-benchmark-merged-chart/prd.md
```diff
@@ -267,7 +271,7 @@ Feature: AI model benchmark tool — merged capability/price chart
     Given a model with a composite index of 85.7 and an output rate of $15.00
     When the merged chart renders that model's row
     Then the capability bar's length is proportional to 85.7 over the composite index max
-    And the price-out bar's length is proportional to $15.00 over that band's price axis max
+    And the price-out bar's length is proportional to $15.00 over the chart's shared price axis max
```
wahidyankf, line 274: **HIGH** | confidence 94 (raised from the raw 88 by tool-verification — the absence is mechanically checkable: `grep -c "AC-48" prd.md` returns `0`) | docs

### AC-48 exists in the feature file, `delivery.md`, and the test suite — but never reached `prd.md`'s acceptance-criteria enumeration

`prd.md`'s `## Acceptance criteria (Gherkin)` block (lines 259-333) contains exactly **11** scenarios and has **zero** occurrences of AC-48 or its scenario. AC-48 — *"A rated model with no reported price shows a not-reported placeholder"* — is present everywhere else:

- `specs/apps/ayokoding/behavior/ayokoding-www/gherkin/tools/ai-benchmark.feature:333-337` (the scenario itself, with its `# AC-48 — added post-merge` rationale comment at `:327-331`)
- `delivery.md:738-765` (the post-hoc addition step, with the full traceability block added by cycle 2's fix)
- `apps/ayokoding-www/src/features/ai-benchmark/shell/benchmark-chart.test.tsx:657-672` (the covering `describe` block)

`delivery.md:756`'s own AC-48 **Files-Changed** note lists `ai-benchmark.feature` and the test file — and does **not** list `prd.md`, which is precisely why the backfill was missed: the traceability record itself never claimed `prd.md` was touched. Meanwhile `README.md:87` describes `prd.md` as holding "Personas, user stories, the complete UI design funnel, **Gherkin acceptance criteria**" — so `prd.md` is the document this plan declares to be the complete AC enumeration, and it is now the only artifact of the four that is missing a criterion this PR shipped and tested.

Note this is a **different site of the same class** as cycle 2's resolved `delivery.md:735` finding, not a re-raise of it: that fix backfilled `delivery.md` and stopped there. Per this repo's "fix the class, not the sites a finding names" discipline, `prd.md` is the remaining site.

**Fix direction**: insert the AC-48 scenario into `prd.md`'s Gherkin block, verbatim from `ai-benchmark.feature:333-337` (placed alongside the sibling rated-model row scenarios), and extend `delivery.md:756`'s AC-48 Files-Changed note to include `prd.md` so the traceability record matches reality.

> *Thread anchored to the nearest line this PR actually changed — GitHub review comments cannot anchor to unchanged lines. The finding's real sites are cited inline above.*

wahidyankf (commented):
## PR Review — Cycle 3 of 3 (final cycle), consolidated

**Risk tier**: full
**Specialists fanned out**: all eight specialists (architecture, logic, governance, security, integrity, performance, docs, instruction)
**Security-sensitive-path override applied**: no (tier already `full` on size — 31 files / +3004 / -1480; no secrets, git-identity, CI-workflow, or `pr-merge-protocol.md` path touched)
**Diff coverage**: full diff reviewed in one pass (D13 no-exclusion posture — no generated-file filtering applied)
**Comment anchoring**: each finding's thread is anchored to the nearest line this PR changed in the relevant file; GitHub cannot anchor a review comment to an unchanged line, so the true `file:line` sites are cited inside each comment.
**Prior-cycle human dismissals respected**: none this cycle (all 11 prior threads — 9 from cycle 1, 2 from cycle 2 — read as agent-resolved, zero human "won't fix"/"I disagree" dismissals)

### Fan-out result

Seven of eight specialists returned **clean** after independently re-verifying the final PR state end-to-end: architecture, logic, governance, security, integrity, performance, instruction. Only **docs** surfaced findings.

Three raw docs findings entered coordination; all three survived dedup, re-categorization, the reasonableness filter, and tool-verification against `81f224b52`. Two had their confidence **raised** by tool-verification (F2 93→96, F3 88→94) because the underlying gap proved mechanically checkable, not inferential. Nothing was dropped this cycle; no raw finding scored below the 80 bar post-verify.

**Dedup / re-categorization notes**:

- **F1** consolidates the same stale-count defect appearing in two documents (`README.md` + `tech-docs.md`) into one thread — one edit class, not two findings.
- **F2** consolidates one false claim asserted in three documents into one thread. The docs specialist also checked the two *sibling* "unchanged" claims in the same sentence (`core/bands.ts`, `shell/chart-primitives.tsx`); I re-verified both diffs independently — both are comment/docstring-only, so those claims remain **true** and are correctly excluded. Only the `benchmark-filters.tsx` claim is false.
- **F3** boundary call (docs vs. integrity): the defect is traceability-shaped (an AC present in three artifacts and absent from a fourth), but the fix is a documentation-completeness edit against `README.md`'s own declaration of what `prd.md` contains. Kept under **docs**. It is *not* a re-litigation of cycle 2's resolved `delivery.md:735` AC-48 thread — different file, and the cycle-2 fix demonstrably did not reach `prd.md` (its own Files-Changed note omits it).

**Blocking status**: this review is submitted as `COMMENT` because `gh` authenticates as the PR author and GitHub rejects `REQUEST_CHANGES` on one's own PR. **It is blocking** — the two `HIGH` findings must be resolved before merge. All three are documentation-only edits under `plans/in-progress/`; none touches `apps/`, `libs/`, or `specs/`, so none can regress CI behavior.

plans/done/2026-07-30__ayokoding-www-ai-benchmark-merged-chart/README.md
wahidyankf: Fixed: changed `README.md:72`'s diagram label from `+4 sort query params` to `+3 sort query params`, and `tech-docs.md:11`'s label from `+4 sort params (NEW)` to `+3 sort params (NEW)`, matching the shipped `SORT_PARAM_KEYS` (sortOpus/sortSonnet/sortLight) and the correct counts already stated elsewhere in both documents (README.md:41-46, tech-docs.md DD-4/File impact). Pushed in commit 47d9d00c1.

plans/done/2026-07-30__ayokoding-www-ai-benchmark-merged-chart/prd.md
wahidyankf: Fixed: updated all three sites with equivalent wording describing the actual, narrow, backward-compatible change — `prd.md:246` (OOS-1), `README.md:54`, and `tech-docs.md:180-182` (removed `benchmark-filters.tsx` from the "No change to" list and added a new File Impact table row for it) now all state that `benchmark-filters.tsx`'s own filter bar has no behavioral change, other than the backward-compatible `allLabel?` widening of the shared `FilterSelect` required by the PR #125 cycle-1 sort-dropdown fix. Pushed in commit 47d9d00c1.

plans/done/2026-07-30__ayokoding-www-ai-benchmark-merged-chart/prd.md
wahidyankf: Fixed: backfilled the AC-48 scenario ("A rated model with no reported price shows a not-reported placeholder") verbatim from `ai-benchmark.feature:333-337` into `prd.md`'s "Acceptance criteria (Gherkin)" enumeration, placed alongside the sibling rated-model-row scenarios (after the subscription-only scenario). Also extended `delivery.md:756`'s AC-48 Files-Changed note to include `prd.md`, so the traceability record now matches reality. `grep -c "^  Scenario" prd.md` now returns 12 (11 + AC-48). Pushed in commit 47d9d00c1.

PR #155, opened 2026-08-07
fix(rhino-cli): run dotnet-install.sh under sudo, not just the symlink step

wahidyankf:
## Summary

- `install_dotnet`'s Linux path installs .NET via Microsoft's official `dotnet-install.sh`, downloaded to a temp file and executed as `bash "$temp_dir/dotnet-install.sh" --install-dir /usr/share/dotnet`, followed by `sudo ln -sf /usr/share/dotnet/dotnet /usr/local/bin/dotnet`.
- Only the trailing symlink step ran under `sudo`. `dotnet-install.sh` itself does `mkdir -p /usr/share/dotnet`, which requires root on a stock Debian/Ubuntu runner where `/usr/share` is root-owned — unprivileged, that step fails with `Permission denied`.
- Root-caused via a live CI failure on `ose-private`'s self-hosted runner (PR #27, run `31168548839`): ~90% of matrix jobs failed with the identical error because every job shares a setup step that installs dotnet, so the failure cascaded to structurally unrelated gates (shellcheck, markdownlint, prettier, etc). This is the same runner that previously blocked PR#22's Cycle 3 gate.
- Confirmed `sudo` is passwordless-available on that runner: an earlier `sudo apt-get install -y build-essential` step in the same job already succeeds. So this is a one-line, no-runner-config fix: prefix the install script invocation with `sudo` too.

## Fix

`sudo bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet` (was: `bash "$temp_dir/dotnet-install.sh" ...`, unprivileged).

## Test plan

- [x] RED: added `install_dotnet_linux_runs_install_script_with_root_privileges`, confirmed it fails against the pre-fix script (missing `sudo` prefix)
- [x] GREEN: applied the one-line fix, confirmed the new test passes
- [x] `cargo test --release -p rhino-cli --lib application::doctor::tools::` — 26/26 passed, no regressions to existing dotnet-install assertions (channel derivation, shell-metacharacter rejection, TLS pinning, no-pipe-to-shell)
- [x] `nx run rhino-cli:test:quick` — 1363/1363 passed, spec coverage intact (67 specs, 447 scenarios, 1825 steps)
- [x] No golden-master fixtures reference the dotnet install script content — no snapshot regen needed
- [x] `parity manifest generate` — regenerated; this file is byte-identical across `ose-public`/`ose-private` (per the current enforced scope), so the same fix must propagate to `ose-private` (tracked as a follow-up on the in-flight `sdlc-gate-registry-enforcement` plan's PR #27, which already carries an unrelated pending CI-gate issue this directly resolves)

### DIFF
--- a/apps/rhino-cli/src/application/doctor/tools.rs
+++ b/apps/rhino-cli/src/application/doctor/tools.rs
@@ -318,7 +318,7 @@ fn install_dotnet(req: &str, platform: &str) -> Vec<InstallStep> {
 temp_dir=$(mktemp -d)
 trap 'rm -rf "$temp_dir"' EXIT
 curl --proto '=https' --tlsv1.2 -fsSL https://dot.net/v1/dotnet-install.sh -o "$temp_dir/dotnet-install.sh"
-bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet
+sudo bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet
 sudo ln -sf /usr/share/dotnet/dotnet /usr/local/bin/dotnet"#
                 ),
             ],
@@ -1048,6 +1048,21 @@ mod tests {
             !script.contains("curl attacker") && !script.contains("| bash\n"),
             "a metacharacter-bearing channel must never be spliced verbatim into the \
              `bash -c` script"
+        );
+    }
+
+    #[test]
+    fn install_dotnet_linux_runs_install_script_with_root_privileges() {
+        let steps = install_dotnet("10.0.204", "linux");
+        let script = &steps[0].args[1];
+
+        assert!(
+            script.contains("sudo bash \"$temp_dir/dotnet-install.sh\""),
+            "the install script itself must run under sudo, not only the trailing symlink \
+             step — dotnet-install.sh's own `mkdir -p /usr/share/dotnet` requires root on a \
+             stock Debian/Ubuntu runner where /usr/share is root-owned; running only the \
+             symlink under sudo leaves the install's directory-creation step unprivileged \
+             and it fails with 'Permission denied'"
         );
     }
 
--- a/apps/rhino-cli/src/application/doctor/tools.rs
+++ b/apps/rhino-cli/src/application/doctor/tools.rs
@@ -291,6 +291,18 @@ fn dotnet_channel(req: &str) -> String {
     }
 }
 
+/// GPG key fingerprint (40 hex chars) that must sign `dotnet-install.sh`,
+/// pinned from Microsoft's published key at `https://dot.net/v1/dotnet-install.asc`
+/// (live-verified 2026-08-07: key `B9CF1A51FC7D3ACF`, uid "Microsoft
+/// `DevUXTeamPrague` <devuxteamprague@microsoft.com>", detached signature at
+/// `https://dot.net/v1/dotnet-install.sig` verifies against the fetched
+/// script). Pinning the fingerprint — rather than trusting whatever key the
+/// `.asc` endpoint happens to serve at install time — closes the same class
+/// of gap `OPENTOFU_*_SHA256` closes for `install_tofu`: without a pin, a
+/// compromised `dot.net` endpoint could serve a malicious script alongside a
+/// matching malicious key/signature pair and still pass verification.
+const DOTNET_INSTALL_SH_GPG_FINGERPRINT: &str = "2B930AB1228D11D5D7F6B6ACB9CF1A51FC7D3ACF";
+
 /// Returns install steps for .NET SDK.
 ///
 /// On macOS: `brew install dotnet`.
@@ -299,6 +311,29 @@ fn dotnet_channel(req: &str) -> String {
 /// `dotnet-sdk` snap is deliberately avoided — its track catalog lags
 /// behind current .NET releases (capped at `8.0/stable` as of 2026-08),
 /// so a hardcoded newer channel there fails to install on every CI run.
+///
+/// The Linux command is:
+///
+/// ```text
+/// curl -fsSL https://dot.net/v1/dotnet-install.sh -o "$temp_dir/dotnet-install.sh"
+/// # + .sig/.asc download and GPG signature verification (see
+/// #   DOTNET_INSTALL_SH_GPG_FINGERPRINT)
+/// sudo mkdir -p /usr/share/dotnet
+/// sudo chown "$(id -u):$(id -g)" /usr/share/dotnet
+/// bash "$temp_dir/dotnet-install.sh" --channel <channel> --install-dir /usr/share/dotnet
+/// sudo ln -sf /usr/share/dotnet/dotnet /usr/local/bin/dotnet
+/// ```
+///
+/// Only the directory creation is privileged, not the whole script: `sudo`
+/// scope is narrowed to `mkdir`/`chown`/`ln` because `/usr/share` is
+/// root-owned on stock Debian/Ubuntu, so creating `/usr/share/dotnet` needs
+/// root — but Microsoft's install script itself is designed to perform a
+/// non-admin installation once its target directory exists and is writable,
+/// and its own writes (extraction, SDK files) stay confined to
+/// `--install-dir` and its own `mktemp` working directory. Running the
+/// entire curl-downloaded third-party script under `sudo`, as an earlier
+/// version of this function did, would escalate every one of those writes to
+/// root for no reason beyond the single `mkdir` that actually requires it.
 fn install_dotnet(req: &str, platform: &str) -> Vec<InstallStep> {
     if platform == "darwin" {
         vec![InstallStep {
@@ -318,7 +353,20 @@ fn install_dotnet(req: &str, platform: &str) -> Vec<InstallStep> {
 temp_dir=$(mktemp -d)
 trap 'rm -rf "$temp_dir"' EXIT
 curl --proto '=https' --tlsv1.2 -fsSL https://dot.net/v1/dotnet-install.sh -o "$temp_dir/dotnet-install.sh"
-sudo bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet
+curl --proto '=https' --tlsv1.2 -fsSL https://dot.net/v1/dotnet-install.sig -o "$temp_dir/dotnet-install.sig"
+curl --proto '=https' --tlsv1.2 -fsSL https://dot.net/v1/dotnet-install.asc -o "$temp_dir/dotnet-install.asc"
+export GNUPGHOME="$temp_dir/gnupg"
+mkdir -m 700 "$GNUPGHOME"
+gpg --batch --import "$temp_dir/dotnet-install.asc" >/dev/null 2>&1
+actual_fingerprint=$(gpg --batch --with-colons --fingerprint | awk -F: '/^fpr:/ {{print $10; exit}}')
+if [ "$actual_fingerprint" != "{DOTNET_INSTALL_SH_GPG_FINGERPRINT}" ]; then
+  echo "dotnet-install.sh signing key fingerprint mismatch: expected {DOTNET_INSTALL_SH_GPG_FINGERPRINT}, got $actual_fingerprint" >&2
+  exit 1
+fi
+gpg --batch --verify "$temp_dir/dotnet-install.sig" "$temp_dir/dotnet-install.sh"
+sudo mkdir -p /usr/share/dotnet
+sudo chown "$(id -u):$(id -g)" /usr/share/dotnet
+bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet
 sudo ln -sf /usr/share/dotnet/dotnet /usr/local/bin/dotnet"#
                 ),
             ],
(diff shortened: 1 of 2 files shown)


apps/rhino-cli/src/application/doctor/tools.rs
```diff
@@ -318,7 +318,7 @@ fn install_dotnet(req: &str, platform: &str) -> Vec<InstallStep> {
 temp_dir=$(mktemp -d)
 trap 'rm -rf "$temp_dir"' EXIT
 curl --proto '=https' --tlsv1.2 -fsSL https://dot.net/v1/dotnet-install.sh -o "$temp_dir/dotnet-install.sh"
-bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet
+sudo bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet
```
wahidyankf: ### `sudo` escalates the entire unverified third-party script, not the one operation that needs root

**Severity**: HIGH
**Confidence**: 92
**Raised by**: architecture, security *(merged — see rationale below)*

**Why these two findings are one thread.** The architecture specialist framed this as a design/tradeoff defect ("privilege scope is broader than the diagnosed cause"); the security specialist framed it as a threat-model defect ("the escalated content has no integrity gate"). Same `file:line`, same root cause: **an unverified, rolling third-party script executes as root.** They are merged because the target end-state is a single coherent rewrite matching this file's established sibling pattern, and because posting them separately invites a partial fix — applying only a checksum would leave the broad-privilege problem untouched, and it is the more urgent half.

**The defect.** `tools.rs:321` prefixes `sudo` onto the whole script invocation:

```bash
sudo bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet
```

The PR's own diagnosis identifies exactly **one** operation needing root: `dotnet-install.sh`'s internal `mkdir -p /usr/share/dotnet` against root-owned `/usr/share`. The fix elevates far more than that. I fetched the current script and confirmed it does not merely `mkdir` — it **downloads a further remote tarball and extracts it**. Under this diff, that entire download-and-extract pipeline runs as UID 0.

**Primary-source corroboration (tool-verify).** Microsoft's own documentation states the `dotnet-install` scripts *"perform a **non-admin installation** of the .NET SDK"*, and that `--install-dir` *"is created if it doesn't exist."* Running the whole script as root is contrary to its documented design. This **raised** my confidence above the raw finding's 90.

**This contradicts the file's own established pattern.** Sibling `install_tofu` (`tools.rs:501-527`) verifies a SHA-256 checksum, then invokes `sudo` only for the final narrow `install -m 0755 <verified-source> <dest>` (L523). `install_docker` (L338-348) elevates only a single `apt-get install`, which is itself gated by APT's GPG-signed-repo chain **before** privileged code executes. `dotnet-install.sh` has no equivalent gate — TLS + domain pinning is transport authenticity, not content integrity.

**Blast radius.** `apps/rhino-cli` byte-identity spans `ose-public`, `ose-primer`, and `ose-private`. On `ose-public`'s ephemeral `ubuntu-latest`, root is cheap. On `ose-private`'s **persistent self-hosted runner**, root obtained from a rolling unverified script persists across jobs and across repos.

---

#### Remediation

**(A) Narrow the privilege — blocking.** Pre-create and hand over the one root-owned path, then run the script as designed (unprivileged):

```bash
sudo mkdir -p /usr/share/dotnet
sudo chown "$(id -u):$(id -g)" /usr/share/dotnet
bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet
sudo ln -sf /usr/share/dotnet/dotnet /usr/local/bin/dotnet
```

**Verified feasible, not hypothetical.** I read the fetched script: `mkdir -p "$install_root"` (L1565) and the extraction into `$install_root` (L1607, and L1007 within extraction) are its only writes outside its own `mktemp` scratch directory. With `/usr/share/dotnet` pre-created and owned by the runner user, the `mkdir -p` is a successful no-op and the install proceeds unprivileged.

**(B) Add the integrity gate — strongly recommended, and correct the mechanism.** The security specialist proposed pinning a SHA-256 of the script. **Tool-verify says use GPG instead.** `https://dot.net/v1/dotnet-install.sh` is Microsoft's *rolling* "stable version" endpoint — its content is republished over time, so a hardcoded SHA-256 is a maintenance trap that will break CI on Microsoft's schedule, not yours. Microsoft publishes a co-updated detached signature and public key:

- `https://dot.net/v1/dotnet-install.sig` — I fetched it: live, `200`, 481 bytes
- `https://dot.net/v1/dotnet-install.asc` — I fetched it: live, `200`

Verifying with `gpg --verify` before execution is the durable equivalent of what `install_tofu` gets from its pinned checksum, and it is what closes the gap against `install_docker`'s APT-GPG trust chain. Note `install_tofu` does **both** narrow-`sudo` and integrity verification — (A) and (B) together are what "match the sibling pattern" actually means.

If (B) is deferred, (A) still stands on its own: with the script no longer running as root, the unverified-content exposure drops from root-level to runner-user-level.

**If (A) proves infeasible**, attach the empirical failure from the actual runner and land the current form with a tracked follow-up. Evidence, not convenience.

apps/rhino-cli/src/application/doctor/tools.rs
```diff
@@ -318,7 +318,7 @@ fn install_dotnet(req: &str, platform: &str) -> Vec<InstallStep> {
...
 sudo ln -sf /usr/share/dotnet/dotnet /usr/local/bin/dotnet"#
```
wahidyankf, line 370: ### `install_dotnet`'s rustdoc omits the Linux command entirely, unlike every sibling installer

**Severity**: MEDIUM
**Confidence**: 88
**Raised by**: docs

The doc comment at `tools.rs:294-301` describes *which* script is used and *why the snap was rejected*, but never states the actual Linux command — sudo-prefixed or otherwise. Every sibling installer does state it inline:

- `install_git` (L186-189): `` /// On Linux: `sudo apt-get install -y git`. ``
- `install_docker` (L329-332): `` /// On Linux: `sudo apt-get install -y docker.io docker-compose-v2`. ``
- `install_jq` (L352-355): `` /// On Linux: `sudo apt-get install -y jq`. ``

I verified all four doc comments directly at the head SHA. `install_dotnet` is the outlier.

This matters more after this PR than before it, because the privilege level is now a **deliberate, load-bearing decision** rather than an incidental detail. The WHY — root is needed because `dotnet-install.sh`'s own `mkdir -p` hits root-owned `/usr/share` — currently exists in exactly two places, **neither of which is reachable from the code**: the new test's assertion message, and the PR body. It does not surface via `cargo doc`, IDE hover, or simply reading the function. The next person to touch this installer will not find the reasoning.

**Sequencing — do this after the HIGH above, not alongside it.** The correct doc text depends on which command survives remediation (A). Writing the doc now would document a command that is about to change. Once the install step is final, extend the doc comment to state the resulting Linux command and the one-line privilege rationale, matching sibling style — and if remediation (A) lands, the rationale becomes the more interesting *"the script runs unprivileged; only the install-directory creation and the symlink need root,"* which is precisely the kind of non-obvious decision a doc comment exists to preserve.

wahidyankf (commented):
**Cycle**: 1 of 3
**Risk tier**: full (overriding the size-based `lite` default)
**Specialists fanned out**: all nine specialists
**Per-specialist raw findings**: architecture 1, logic 0, governance 0, security 1, integrity 0, performance 0, docs 1, instruction 0, types 0
**Security-sensitive-path override applied**: yes — `doctor/tools.rs` is executed directly by CI (`.github/workflows/pr-quality-gate.yml` L78-80, L144), and this diff is a privilege escalation whose byte-identity propagation reaches a persistent self-hosted runner
**Diff coverage**: full diff reviewed in one pass (2 files, +17/-2)
**Prior-cycle human dismissals respected**: none this cycle (no prior reviews, comments, or threads on this PR)

---

## Verdict: BLOCKING

This review is submitted as `COMMENT` because `gh` authenticates as the PR author and GitHub rejects `REQUEST_CHANGES` on one's own PR. **It is blocking regardless** — one `HIGH` finding survives, and its remedy is small and empirically verified feasible.

**2 findings posted, 1 dropped, 6 specialists returned clean.** Three of nine specialists raised findings. Two of those (architecture, security) named the same `file:line` defect from different angles and were **merged into a single thread** by the Deduplicate function — independent convergence on one root cause is a confidence signal, and splitting them here would invite a partial fix (see that thread for the merge rationale).

### What tool-verify changed

Two raw findings were materially corrected against primary sources rather than passed through as filed. Both corrections are recorded in-thread:

1. **Microsoft publishes an official GPG signature for `dotnet-install.sh`** (`https://dot.net/v1/dotnet-install.sig` + key at `.asc`) — I fetched both; live, `200`, 481-byte detached signature. The security specialist's observation (no integrity gate) is correct and stands, but its suggested remedy — pinning a SHA-256 — is the **wrong mechanism** for this endpoint and has been replaced. See the thread.
2. **Microsoft's own documentation states the script "performs a non-admin installation of the .NET SDK"**, and that `--install-dir` "is created if it doesn't exist." This is primary-source corroboration that running the whole script under `sudo` is contrary to its documented design — it raised, not lowered, the architecture finding's confidence.

### On the scope-versus-urgency tradeoff

The PR's stated scope is narrow (unblock CI) and `ose-private` PR #27 is waiting downstream. I weighed deferring the `HIGH` to a follow-up and **rejected that**, on these grounds:

- The pre-fix state is broken, so "ship it, the status quo is safe" does not apply — but neither does urgency justify the *broadest* available fix when a narrower one costs two lines.
- The narrower fix is **verified feasible**, not hypothetical: I read the fetched script and confirmed `mkdir -p "$install_root"` (L1565) and the extraction (L1607, L1007) are the only writes outside the script's own `mktemp` scratch dir.
- The blast radius is asymmetric. On `ose-public`'s ephemeral `ubuntu-latest` runner, root is cheap. On the **persistent self-hosted runner** this propagates to via `rhino-cli` byte-identity, root from a rolling, unverified third-party script persists across jobs and across repos.

**Escape hatch, stated honestly**: if the fixer empirically demonstrates the narrow approach fails on the actual runner, land the current form with a tracked follow-up and the evidence attached. The burden is evidence, not convenience.

### Dropped by the reasonableness-filter

- **Commit-message header length** (governance, self-suppressed): 75-char header exceeds the "should be ≤50" guidance. Correctly suppressed — non-blocking guidance, inconsistent with the repo's own actual practice, and re-raising it would be nitpick noise.

### Clean specialists (no findings)

`logic`, `governance`, `integrity`, `performance`, `instruction`, `types` — six clean. Two are worth surfacing because they are load-bearing to this verdict, and I re-verified both independently:

- **integrity** performed a real RED/GREEN: reverted the source line, confirmed the new test fails; restored, confirmed it passes. I re-ran `cargo test install_dotnet` → **7 passed**, and confirmed the regenerated `parity-manifest.sha256` entry (`2401640b…`) matches `shasum -a 256` on the file **exactly**. The Regression Test Mandate is satisfied with a genuinely behavior-faithful test, not a tautological one.
- **logic** established that the pre-existing `sudo ln -sf` was dead weight — `dotnet-install.sh`'s own `mkdir -p` failed first, before control ever reached it. The PR's root-cause diagnosis is correct. The `HIGH` below disputes the *scope of the remedy*, not the diagnosis.

---

apps/rhino-cli/src/application/doctor/tools.rs
wahidyankf: Fixed: `apps/rhino-cli/src/application/doctor/tools.rs` — `install_dotnet`'s Linux install script no longer runs the entire curl-downloaded `dotnet-install.sh` under `sudo`. It now:

1. Runs `sudo mkdir -p /usr/share/dotnet && sudo chown "$(id -u):$(id -g)" /usr/share/dotnet` first (the only step that actually needs root, since `/usr/share` is root-owned on stock Debian/Ubuntu).
2. Runs `bash "$temp_dir/dotnet-install.sh" --channel {channel} --install-dir /usr/share/dotnet` unprivileged — matching Microsoft's documented non-admin installation design and the minimal-elevation pattern `install_tofu` already established in this file.
3. Keeps `sudo ln -sf /usr/share/dotnet/dotnet /usr/local/bin/dotnet` for the final symlink.

Also added GPG signature verification before executing the script: downloads `dotnet-install.sig` and `dotnet-install.asc` alongside the script, imports the key into a scratch `GNUPGHOME`, checks the imported key's fingerprint against a pinned constant (`DOTNET_INSTALL_SH_GPG_FINGERPRINT = "2B930AB1228D11D5D7F6B6ACB9CF1A51FC7D3ACF"`, live-verified against Microsoft's `devuxteamprague@microsoft.com` signing key on 2026-08-07), then runs `gpg --batch --verify` against the downloaded script — all before the script is executed. Pinning the fingerprint (rather than trusting whatever key `.asc` happens to serve) closes the same class of gap `OPENTOFU_*_SHA256` closes for `install_tofu`.

The regression test `install_dotnet_linux_runs_install_script_with_root_privileges` was replaced by two tests: `install_dotnet_linux_narrows_root_privilege_to_directory_setup_and_symlink_only` (asserts the whole-script `sudo` is gone, the unprivileged invocation runs, and privileged `mkdir`+`chown` precede it) and `install_dotnet_linux_verifies_gpg_signature_before_running_install_script` (asserts the signature/key downloads, fingerprint pin, and `gpg --verify` all run before the install script). Both were confirmed RED against the pre-fix script shape and GREEN after, per TDD discipline.

`nx run rhino-cli:test:quick` passes (1364 tests, 0 failed), `apps/rhino-cli/parity-manifest.sha256` was regenerated, and both land in commit `b380cd797` on `fix/dotnet-install-sudo-privileges`.

apps/rhino-cli/src/application/doctor/tools.rs
wahidyankf: Fixed: extended `install_dotnet`'s doc comment (`apps/rhino-cli/src/application/doctor/tools.rs`) to state the Linux command shape inline, matching the style of sibling `install_git`/`install_docker`/`install_jq` doc comments, and applied to the *revised* (post-Thread-1) command:

```text
curl -fsSL https://dot.net/v1/dotnet-install.sh -o "$temp_dir/dotnet-install.sh"
# + .sig/.asc download and GPG signature verification
sudo mkdir -p /usr/share/dotnet
sudo chown "$(id -u):$(id -g)" /usr/share/dotnet
bash "$temp_dir/dotnet-install.sh" --channel <channel> --install-dir /usr/share/dotnet
sudo ln -sf /usr/share/dotnet/dotnet /usr/local/bin/dotnet
```

with the privilege rationale stated explicitly: only the directory creation needs root (because `/usr/share` is root-owned on stock Debian/Ubuntu); `dotnet-install.sh` itself is designed by Microsoft to run unprivileged once its target directory exists and is writable, and running the whole script under `sudo` (as an earlier version of this function did) would escalate every write it makes for no reason beyond that one `mkdir`.

Applied after Thread 1's code change landed, so the doc text reflects the final command shape rather than the old single-`sudo bash` invocation. Same commit `b380cd797` on `fix/dotnet-install-sudo-privileges`.

PR #156, opened 2026-08-07
docs(plans): sync onboarding-refresh delivery checklist to completed evidence

wahidyankf:
## Purpose

Recover uncommitted delivery-checklist progress found sitting in the
`worktrees/repository-onboarding-readme-refresh` worktree from an interrupted
prior session — tick checkboxes and refresh the public execution record to
match work already reflected in that session's own Files Changed/Commands
evidence, and in already-merged companion PRs (#148, #153, #154).

Delivered via the `plan-take-over-execution.md` workflow's Phase C
(take-over-the-live-work) step: adopt the existing worktree/branch, apply the
freshness gate (rebase onto current `origin/main`), and reconcile
`delivery.md` to discovered ground truth before continuing execution.

## Validation

- No secrets/credentials in the diff (checked before committing).
- Pre-push gates (env-validate, md-links, md-readme-index, harness-duplication,
  parity-manifest) all passed clean.

PR #159, opened 2026-08-08
docs(plans): complete P1-006 public reader-doc ledger enrichment

wahidyankf:
## Summary

- Completes P1-006 of the repository-onboarding-readme-refresh plan: expands Purpose/Commands-Evidence/Notes for all 678 audit-required and reader-related rows in the ose-public reader-document disposition ledger (`plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md`).
- 676 rows templated across 23 document-kind templates; 2 rows individually authored (root `README.md`, `docs/reference/related-repositories.md`).
- Existing `P1-doc-00001`..`P1-doc-00872` Task IDs preserved verbatim — this PR enriches columns only, doesn't renumber or reclassify.

## Test plan

- [x] 678/678 audit-required/reader-related rows carry a unique `P1-doc-` Task ID (no gaps introduced)
- [x] 0 remaining placeholder text (`grep -c "one-document task\|inspect source then run Markdown and PR checks"` → 0)
- [x] 4783 total lines preserved (row count unchanged, only column content enriched)


wahidyankf (commented):
## PR Review — Cycle 1 (consolidated)

**Cycle**: 1 of 3
**Risk tier**: full
**Specialists fanned out**: all nine specialists, minus 2 DD-10 content-type skips — `integrity` (skipped: no test or CI files in diff) and `types` (skipped: no typed source in diff)
**Per-specialist raw findings**: architecture 0 (1 routing note, no independent finding), logic 3, governance 2 (1 posted, 1 self-declined at <80), security 0, integrity 0 (skipped, see above), performance 0, docs 2, instruction 1, types 0 (skipped, see above)
**Diff coverage**: full diff reviewed in one pass (single file, +4779/-4779), via scout's two pre-cut slices
**Prior-cycle human dismissals respected**: none this cycle (Cycle 1 — no prior reviews or threads exist on this PR)

> **Mechanics note — no line-anchored threads this cycle.** GitHub rejects review comments on this file with `422 Diff entry ... diff is too large`; I probed a single comment in isolation to confirm it is a structural limit on the diff entry, not comment volume. Every finding below therefore carries a **blob permalink pinned to `1b00b7d`** with an explicit line number, which the finding-evidence rules permit in place of a thread anchor. **For `pr-review-fixer`: there are no threads to resolve this cycle — reply to this review with a per-finding disposition instead.**

---

### Verdict: BLOCKING — substantial rework required

This review is **blocking** despite its `COMMENT` state (`REQUEST_CHANGES` is unavailable when `gh` authenticates as the PR author).

**480 of the 678 enriched rows (70.8%) carry a Verify command that cannot execute as written, or that can never fail regardless of the condition it claims to check.** Only 198 rows are clean. That is not a set of spot defects to patch — the generator that produced these rows needs correcting and the rows need regenerating.

This matters more than usual because of what the artifact *is*. [`delivery.md:328-330`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/delivery.md#L328-L330) states P1-006's acceptance as: each row names "one path, one direct action, its source of truth, **exact applicable command**, concrete acceptance criterion, and implementation fields." The exact-applicable-command clause is the one this PR most systematically misses. The PR's own test plan checks row counts, Task-ID uniqueness, and placeholder absence — all three genuinely pass — but nothing in it executes a single cited command, which is why a 70.8% defect rate landed green.

**Every finding below was independently reproduced against the pinned SHA, not accepted on specialist agreement.** Two corrections came out of that verification pass, and they cut in both directions:

- The `find` defect was **narrower** than the raw finding claimed. The `\*` in `-name '\*.feature'` is a markdown escape that renders correctly as `*`; no Verify cell in this ledger uses code spans, so this is consistent authoring, not a copy-paste break for a human reading the rendered table. The escape-independent vacuousness is the real defect, and it is **130 rows — not the 247** that use the glob. Finding 5 covers the raw-consumption residue separately, at MEDIUM.
- The governance finding's **line number was wrong**. P1-006's checkbox is at `delivery.md:328`, not `:335`; line 335 sits inside P1-008's already-ticked evidence block. Substance confirmed, anchor corrected (Finding 6).

| # | Severity | Conf | Scope | Summary |
|---|---|---|---|---|
| 1 | CRITICAL | 99 | 346 rows (51%) | `md links validate <path>` is not a valid invocation — errors out |
| 2 | CRITICAL | 96 | 130 rows (19%) | `find` verify command is structurally vacuous — can never fail |
| 3 | HIGH | 96 | 4 rows | Rows target a plan folder archived before this PR's head |
| 4 | MEDIUM | 88 | 2 rows | Verify command does not establish the row's own Acceptance criterion |
| 5 | MEDIUM | 85 | 247 rows | Commands stored as unfenced prose; raw-file form is not executable |
| 6 | HIGH | 92 | `delivery.md:328` | P1-006 checkbox lockstep (Rule 13) — remedy reframed |

---

## Finding 1 — CRITICAL: `md links validate <path>` is not a valid invocation (346 rows, 51%)

**Severity**: CRITICAL · **Confidence**: 99
**Raised by**: logic
**Evidence**: [`reader-doc-disposition-ose-public.md:355`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L355) (root `README.md`, `P1-doc-00007`); 345 further rows including [`:2952`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2952)

`md links validate` accepts **no positional path argument**. Its only options are `--staged-only`, `-v`, `--exclude`, `-q`, `-o`, `--no-color`, `--say`. Every row citing it with a trailing path fails immediately. Reproduced verbatim at the pinned SHA:

```
$ cargo run --release --quiet --manifest-path apps/rhino-cli/Cargo.toml -- md links validate README.md
error: unexpected argument 'README.md' found

Usage: rhino-cli md links validate [OPTIONS]
```

**Scope**: 346 of 678 rows (51%). Every occurrence of `md links validate` in this ledger carries a positional path, so the defect rate within this command family is **346/346**. It spans template clusters T02/T03/T04/T07/T11/T12/T19 *and* both individually-authored rows — being hand-written was no protection.

CRITICAL maps cleanly here under the [Criticality Levels Convention](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/repo-governance/development/quality/criticality-levels.md): "syntax errors that prevent execution."

**Fix**: `md links validate` is a whole-tree validator and cannot be scoped to one document. Either drop the path and state that the row's link check runs repo-wide, or substitute a per-file check that genuinely accepts a path. Whichever you pick, **execute the corrected form once before regenerating all 346 rows** — this defect exists because a plausible-looking invocation was templated without ever being run.

---

## Finding 2 — CRITICAL: the `find` verify command is structurally vacuous and can never fail (130 rows)

**Severity**: CRITICAL · **Confidence**: 96
**Raised by**: logic, docs, architecture *(three-way independent convergence; scope corrected downward by tool-verify)*
**Evidence**: [`reader-doc-disposition-ose-public.md:4534`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4534) (`specs/apps/ayokoding/README.md`, `P1-doc-00694`), [`:4535`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4535), and 128 further rows

Row 4534 asks the reader to confirm that `specs/apps/ayokoding/README.md` "accurately indexes every Gherkin .feature file present directly under `specs/apps/ayokoding`", and offers `find specs/apps/ayokoding -maxdepth 1 -name '*.feature'` as the check.

There are no `.feature` files directly under that directory — its children are subdirectories. The command returns empty **whether or not the README is stale**, so it can never fail:

```
$ /bin/ls specs/apps/ayokoding/behavior
ayokoding-be  ayokoding-build-tools  ayokoding-cli  ayokoding-www  README.md
$ find specs/apps/ayokoding/behavior -maxdepth 1 -name '*.feature' | wc -l
0
```

A verification command that always passes is worse than no command at all: it converts an unverified row into one that *looks* audited.

**Scope — corrected downward from the raw findings.** I enumerated all 247 glob-style `find` rows and resolved each directory against the working tree: **130 are genuinely vacuous** (114 `.feature`, 16 `.md`). The raw findings quoted 131 (logic) and 63 (docs); the true figure sits between them. Note this is **not** a whole-cluster defect — leaf-level rows in the same clusters, such as `specs/apps/ayokoding/behavior/ayokoding-be/gherkin/content`, are correct and do match files. The template was applied uniformly across hub and leaf directories, and only the hubs broke.

**Fix**: hub READMEs index subdirectories, not same-extension siblings, so they need a structurally different check — e.g. `find <dir> -maxdepth 1 -mindepth 1 -type d` for a hub, reserving the `-name '*.feature'` form for leaves. That shape already appears 5 times elsewhere in this ledger, so the correct pattern is present; it simply was not applied by directory role. Classify each of the 247 rows as hub or leaf before regenerating.

---

## Finding 3 — HIGH: rows target a plan folder archived before this PR's head (4 rows)

**Severity**: HIGH · **Confidence**: 96
**Raised by**: instruction
**Evidence**: [`reader-doc-disposition-ose-public.md:4279`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4279), [`:4282`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4282), [`:4284`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4284), [`:4286`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4286) — Task IDs `P1-doc-00479`..`00482`

These four rows reference `plans/in-progress/sdlc-gate-registry-enforcement/**`, which does not exist at the pinned SHA:

```
$ ls -d plans/in-progress/sdlc-gate-registry-enforcement
ls: No such file or directory
$ ls -d plans/done/*sdlc-gate-registry-enforcement*
plans/done/2026-08-07__sdlc-gate-registry-enforcement
```

The folder was archived by commit `48a8e97d6`, which is an **ancestor of this PR's head** — these rows were authored against a path that had already moved. Broken on arrival, not gradually stale.

**One nuance for the fix**: five *adjacent* rows (4280, 4281, 4283, 4285, 4287) cite the same dead folder but carry `P1-002` Task IDs, so they predate this PR and are out of this review's scope. Since the fix touches this exact region anyway, repointing all nine at once is cheaper and more consistent, and matches the repo's Root Cause Orientation principle.

**Fix**: repoint to `plans/done/2026-08-07__sdlc-gate-registry-enforcement/`. Also consider whether archived-plan READMEs belong in an audit-required ledger at all — P1-007 in this same plan is specifically about marking `plans/done/**` historical, so these rows may need reclassifying rather than repointing.

---

## Finding 4 — MEDIUM: the Verify command does not establish the row's own Acceptance criterion (2 rows)

**Severity**: MEDIUM · **Confidence**: 88
**Raised by**: logic
**Evidence**: [`reader-doc-disposition-ose-public.md:2952`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2952) (`docs/reference/related-repositories.md`, `P1-doc-00238`) and [`:355`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L355) (root `README.md`, `P1-doc-00007`)

Row 2952's Acceptance is a **content-equivalence** claim: "the four repos ..., their license/visibility labels, and both boundary definitions match AGENTS.md verbatim; no stale repo name or dropped boundary clause."

Its Verify is `md links validate docs/reference/related-repositories.md` — a link-resolution check that never opens `AGENTS.md`. A silently dropped boundary clause, a wrong license label, or a renamed repo all pass. (The command is additionally broken per Finding 1, but fixing that would not close this gap.)

Row 355 has the same mismatch: its Acceptance requires ports and prod branches to match AGENTS.md's Web Sites table, while its Verify runs only prettier, markdownlint, and `md links validate`.

Both rows are the **two individually-authored** rows the PR body calls out — so this is not template fallout and will not be fixed by regenerating templates. It needs per-row attention.

**Fix**: either state the source-of-truth comparison as an explicit manual read step (legitimate — not every acceptance criterion reduces to a command), or supply a command that actually diffs the two surfaces. What should not remain is a mechanical command sitting where a reader will reasonably read it as discharging the acceptance criterion.

---

## Finding 5 — MEDIUM: commands stored as unfenced prose, so the raw-file form is not executable (247 rows)

**Severity**: MEDIUM · **Confidence**: 85
**Raised by**: *(tool-verify refinement during synthesis — the raw findings folded this into Finding 2, where it did not belong)*
**Evidence**: [`reader-doc-disposition-ose-public.md:4535`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4535) and 246 further rows using the `-name '\*.ext'` form

Stating this finding's limit first, because it materially narrows it: **when rendered on GitHub, these commands are correct.** `\*` renders as `*`, and since no Verify cell in this ledger uses code spans, escaping the asterisk is consistent, correct markdown authoring. A human reading the rendered table and copying the command gets the right thing. Prettier and markdownlint pass, correctly.

The issue is raw-file consumption. The bytes on disk are `-name '\*.feature'`, and inside single quotes a backslash is literal:

```
$ D=specs/apps/ayokoding/behavior/ayokoding-be/gherkin/content
$ find $D -maxdepth 1 -name '\*.feature' | wc -l   # as stored on disk
0
$ find $D -maxdepth 1 -name '*.feature' | wc -l     # as rendered
1
```

That directory genuinely contains a `.feature` file, which isolates the escape as the sole cause — it is a leaf row that is otherwise correct.

Why this matters here specifically: every row in this ledger is an `[AI]` task, and agents read raw files, not rendered HTML. The consumer this artifact is written for is exactly the one that receives the non-executable form. **247 rows** are affected, overlapping Finding 2's 130.

**Fix**: wrap Verify-column commands in backticks. Inside a code span no escape is needed, raw and rendered forms converge, and the class cannot recur. Worth folding into the Finding 1-2 regeneration rather than running as a separate pass.

---

## Finding 6 — HIGH: P1-006 checkbox is not in lockstep with execution

**Severity**: HIGH · **Confidence**: 92
**Raised by**: governance *(line number corrected by tool-verify: `:328`, not `:335`)*
**Evidence**: [`delivery.md:328`](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/plans/in-progress/repository-onboarding-readme-refresh/delivery.md#L328) — `- [ ] [AI] [P1-006] Expand each audit-required or reader-related document into one exact [AI] task row in its owning ledger`. Still unchecked, while this PR's diff is exactly the work P1-006 describes.

This violates Rule 13 of the [User-Facing Delivery Hardening Convention](https://github.com/wahidyankf/ose-public/blob/1b00b7d0213e79f57ddf0977d32c5d299f8f94f7/repo-governance/development/quality/user-facing-delivery-hardening.md) — "tick the box the moment the item lands ... reconcile the boxes in the **same** commit — never leave them divergent." This plan already needed one after-the-fact reconciliation pass (commit `0f2f9186`, for PR #156); this repeats the same gap.

**The remedy is inverted by Findings 1-2, and that is the important part.** The raw finding's fix was "tick `delivery.md:328` to `[x]` in this PR." **Do not do that as things stand.** P1-006's acceptance requires an exact applicable command per row, and 70.8% of rows do not have one — the item has not landed. Ticking it now would convert a lockstep gap into a false completion record, which is the worse of the two failures.

**Fix**: land Findings 1-3 first, then tick `:328` with an evidence block **in the same commit** that lands the corrected rows.

---

### CI status

**Not green.** `harness-naming` **failed** — but the failure is in the `Run ./.github/actions/setup-node` step, and the `gate run --surface=ci` step was **skipped**, never reached. This is toolchain-provisioning flake, **not** a defect in this PR's content, and matches the known runner-contention pattern across the OSE repos. 14 further checks were still pending at review time. Re-run `harness-naming` alongside the rework push rather than investigating it as a content failure.

### Filtered out (not posted)

One raw finding was dropped by the reasonableness-filter, recorded here for auditability:

- **docs, MEDIUM/85 — PR body's "23 templates + 2 individually-authored" undercounts bespoke rows.** Verified directionally correct: normalizing the Verify column yields 43 distinct shapes, 30 of them appearing exactly once, against a claimed 2 individually-authored rows. Dropped anyway — it describes the PR body rather than the artifact, and the Findings 1-2 rework will regenerate the template structure wholesale, invalidating any count corrected now. Governance independently reached the same finding and self-declined it at <80 confidence. Worth a one-line PR-body fix when the rework lands; not worth a finding.

---

wahidyankf:
## PR-review fixer response — Cycle 1

Applied all six findings from the consolidated Cycle 1 review directly to
`plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md`
and `plans/in-progress/repository-onboarding-readme-refresh/delivery.md`, pushed as commit
`7763141d5` on top of the pinned `1b00b7d0`. No line-anchored threads existed to resolve
(per the review's own mechanics note), so this reply summarizes disposition per finding.

### Finding 1 — CRITICAL, `md links validate <path>` invalid (346 rows) — **FIXED**

Rescoped every occurrence to `md links validate | grep "^### <path>$"` against the tool's
repo-wide report — the only per-file signal available since the subcommand takes no
positional argument. Verified both directions: a clean file (e.g. `README.md`) yields grep
exit 1 (no match, no broken links); a file with real broken links
(`plans/done/2026-03-29__demo-fs-ts-nextjs/README.md`) yields grep exit 0 with the matching
`### <path>` header. Pipe characters escaped as `\|` for GFM table-cell safety (see note below
on a self-caught regression).

### Finding 2 — CRITICAL, vacuous `find` command (130 rows) — **FIXED**

Classified all 247 `find <dir> -maxdepth 1 -name '*.ext'` rows against the actual working tree:
130 are genuinely vacuous (114 `.feature` returning zero matches; 16 `.md` where the *only*
match is the README doing the checking, so the check can never establish anything about its
siblings). Those 130 became `find <dir> -maxdepth 1 -mindepth 1 -type d`. The remaining 117
were already correct and left semantically unchanged (see Finding 5). Also found and fixed one
further row outside the review's own 247-count during a repo-wide sweep for the same command
shape (`social-media-posts/linkedin/README.md`, `-maxdepth 2` — genuinely a leaf, just needed
the same escape fix as Finding 5).

### Finding 3 — HIGH, archived plan folder (4 rows, +5 adjacent) — **FIXED, full scope taken**

Repointed all 9 rows (the 4 `P1-doc-*` rows plus the 5 adjacent `P1-002` rows sharing the same
dead path) from `plans/in-progress/sdlc-gate-registry-enforcement/**` to
`plans/done/2026-08-07__sdlc-gate-registry-enforcement/**`, per the review's own
cheaper-to-fix-together framing. Not reclassifying these as `plans/done/**`-historical per
P1-007 in this pass — that's P1-007's job, not P1-006's, and reclassifying disposition is a
different kind of edit than repointing a stale path.

### Finding 4 — MEDIUM, Verify doesn't establish Acceptance (2 rows) — **FIXED**

Both individually-authored rows (root `README.md`, `docs/reference/related-repositories.md`)
gained an explicit manual-diff instruction appended to their Verify cell, since neither
`md links validate` nor prettier/markdownlint ever open `AGENTS.md` to confirm the
content-equivalence the Acceptance column actually requires.

### Finding 5 — MEDIUM, unfenced prose / raw-byte escape (247 rows, +1) — **FIXED**

All 247 rows (both the 130 hub and 117 leaf outcomes of Finding 2) now have their Verify
`find` command wrapped in backticks with the backslash-escape removed, so raw and rendered
forms converge. Folded into the same edit pass as Finding 2, per the review's own suggestion.

### Finding 6 — HIGH, P1-006 checkbox lockstep — **FIXED, ticked in this commit**

`delivery.md:328` is now `[x]` with an evidence block (matching the P1-008/P1-009 sibling
style) in the *same* commit that lands the corrected rows — not before, per the review's
explicit inverted-remedy guidance.

### Self-caught regression during this pass (not in the original review)

My first attempt at Finding 1 used an unescaped `|` inside the grep pipeline, which is a raw
GFM table column separator — it silently split every affected row into an extra column and
corrupted the whole table once run through prettier (Prettier's table reflow made the damage
visible: a 4779/4779-line full-file diff). Caught before pushing, discarded via `git stash`
(not `git checkout`, to keep it recoverable), and redone with the pipe escaped as `\|`. Verified
afterward: every one of the 4777 data rows has exactly 14 unescaped pipe delimiters (13
columns), `npx prettier --check` and `npx markdownlint-cli2` both pass clean, and
`md links validate` reports zero broken links in either touched file.

### Deferred / not acted on

- The docs specialist's "23 vs actual template count" finding — synthesis already filtered
  this out as not-postable (describes the PR body, not the artifact; moot once row counts
  regenerate). No action taken, per synthesis's own disposition.

### Verification run before this reply

- `npx prettier --check` — clean on both touched files
- `npx markdownlint-cli2` — 0 errors on both touched files
- `cargo run ... md heading-hierarchy validate` — no violations in either touched file
- `cargo run ... md links validate` — 0 broken links reported under either touched file's own
  heading
- Pre-push gates (`md-links`, `md-readme-index`, `harness-duplication`, `parity-manifest`,
  `env-validate`) all passed on `git push`
- Spot-checked one row per fix category by running the corrected command directly against the
  working tree (hub `find -type d`, leaf `find -name`, grep-anchored `md links validate`,
  repointed `plans/done/**` path, the `-maxdepth 2` outlier) — all matched expected pass/fail
  behavior

Pushed as `7763141d5` on `docs/repository-onboarding-p1-ledger-completion`. CI is running
against the new head; will report status once resolved.

wahidyankf (commented):
## PR Review — Consolidated (Cycle 2)

**Cycle**: 2 of 3
**Risk tier**: full
**Specialists fanned out**: 8 of 9 — architecture, logic, governance, security, performance, docs, instruction, **integrity**. `pr-review-types-maker` skipped (DD-10 content-type filter: diff is Markdown-only, no typed source).
**Deliberate deviation from Cycle 1**: `pr-review-integrity-maker` was **included** this cycle, reversing Cycle 1's DD-10-based exclusion. Rationale: although the diff contains no test/CI files, its substance is *test-plan-shaped* — 678 rows of verification commands paired with acceptance criteria — and both of Cycle 1's CRITICALs were **vacuous-check** failures, precisely integrity's discipline. The call paid for itself: integrity independently proved C1's polarity inversion from tool source and independently confirmed H4.
**Per-specialist raw findings**: architecture 1, logic 5, governance 1, security 0, performance 1, docs 4, instruction 0, integrity 2 (types 0 — skipped, Markdown-only diff)
**Diff coverage**: full diff reviewed in one pass (2 files, 4800 insertions / 4780 deletions vs merge-base `0f2f9186`)
**Prior-cycle human dismissals respected**: none this cycle (zero review threads exist on this PR — Cycle 1 posted at issue level due to the platform constraint below)

**Raw to consolidated**: 14 raw findings, **9 posted** — 3 merges collapsing 5 raw findings into 2, plus 1 reasonableness-filter drop.

> **Platform constraint**: GitHub rejects inline review comments on `artifacts/reader-doc-disposition-ose-public.md` with HTTP 422 ("diff entry too large") — confirmed structural in Cycle 1, re-confirmed this cycle. Every finding below is therefore anchored to a **blob permalink pinned to `7763141d5`** rather than an inline thread.

> **This review is BLOCKING despite its `COMMENT` state.** `gh` authenticates as the PR author, and GitHub forbids `REQUEST_CHANGES` on one's own PR. Blocking status is carried in the severity labels: **2 CRITICAL, 5 HIGH**.

---

## CRITICAL

### C1 — Family-A `md links validate` replacement (346 rows, 51% of the ledger) is non-executable *and* polarity-inverted

**Confidence**: 97 · **Severity**: CRITICAL
**Raised by**: pr-review-logic-maker, pr-review-integrity-maker, pr-review-architecture-maker (three-way independent convergence)

**Evidence** — [L355 (P1-doc-00007)](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L355), [L2760 (P1-doc-00046)](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2760); the pattern recurs identically across exactly **346** rows (verified independently).

Two compounding defects, **both reproduced live** — this is not agreement-counting:

**(a) It does not parse.** The cell reads `... md links validate \| grep "^### <path>$"`. The `\|` is mandatory GFM table-cell pipe escaping, so any raw-text consumer copy-pasting the command receives a *literal* `|`, not a shell pipe operator. Reproduced:

```
$ bash -c 'echo one \| grep foo'
one | grep foo                      # one command, literal pipe — no pipeline formed

$ rhino-cli md links validate '|' 'grep'
error: unexpected argument '|' found
```

`ValidateLinksArgs` (`apps/rhino-cli/src/commands/md_validate_links.rs:16-24`) accepts only `--staged-only` and `--exclude` — no positionals — so the stray arguments are a hard clap parse failure.

**(b) Even hand-repaired, the check asserts the opposite of its Acceptance clause.** Verified against tool source, `format_link_text` (`apps/rhino-cli/src/application/docs/links.rs:628-679`):

```rust
if result.broken_links.is_empty() {
    if !quiet { output.push_str("All links valid! No broken links found.\n"); }
    return output;   // early return: no "### <path>" heading is ever emitted
}
```

Every `### {file}` heading (line 670) is built **exclusively** from files that *do* have broken links. So `grep "^### <path>$"` exits **0 only when the target already has a broken link**, and exits **1 for a clean document** — the exact inverse of the row's own Acceptance ("every internal link resolves"). Integrity's live run corroborates: 147 broken links across 51 `### <path>` headings, all under `plans/done/**`, and **none** among the 678 audited paths — so all 346 rows currently report failure while the docs are healthy, and would only "pass" once a doc breaks.

This is the same defect class as Cycle-1 F1, reintroduced in structurally equivalent form by the very fix intended to close it.

**Boundary ruling (architecture vs correctness)**: filed as **correctness**. Architecture raised the `\|` escaping half under its "verification-command shape/consistency" scope, but this is not a new structural decision — it is a command that does not do what its Acceptance clause claims, i.e. domain behavior. Architecture's escaping evidence is folded in above as supporting evidence for sub-defect (a). Counted once, not three times.

**Fix direction**: invert the test (`! grep -q "^### <path>$"`) or switch to `--format json` plus an explicit `broken_count == 0` assertion; and remove the shell pipe from the GFM cell, or document the `\|` to `|` substitution explicitly. Applies to all 346 rows.

---

### C2 — 55 of the 135 family-B hub rows aim `-type d` at leaf directories, making them permanently vacuous

**Confidence**: 96 · **Severity**: CRITICAL

**Evidence** — [L4566 (P1-doc-00723)](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4566), [L2909 (P1-doc-00195)](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2909).

**Independently reproduced** — I extracted all 135 unique directory arguments from the `find <dir> -maxdepth 1 -mindepth 1 -type d` rows and ran the command against every one:

```
unique dirs: hub=80   leaf (zero subdirectories)=55   missing=0
```

For those **55**, the prescribed command returns **empty output unconditionally** — their real children are files (`.feature` / `.md` / `.yaml`), not subdirectories, exactly as their own Source and Acceptance columns describe. A command that can never produce output cannot establish any acceptance criterion. Cycle 1 swapped one vacuous check (`find -name '*.ext'`) for another on these rows.

Affected paths span `specs/apps/{ayokoding,crane,organiclever,ose,rhino,wahidyankf}/**`, `specs/libs/{fsharp-crane-core,rust-commons,web-ui,web-ui-token}/**`, and `docs/explanation/software-engineering/programming-languages/{c-sharp,f-sharp,rust,typescript}/templates/README.md`.

**Fix direction**: for these 55, use `-type f -name '*.feature'` (or `*.md` / `*.yaml` per each row's own Source column) in place of `-type d`.

---

## HIGH

### H1 — The `delivery.md` evidence block states two counts that contradict the artifact it describes

**Confidence**: 95 · **Severity**: HIGH
**Raised by**: pr-review-governance-maker, pr-review-docs-maker (merged — both name the same evidence paragraph)

**Evidence** — [delivery.md L331-L349](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/delivery.md#L331-L349)

Two distinct factual errors in one paragraph, both verified:

1. **"678 `P1-doc-*` rows corrected"** overstates the fix's own scope. The commit titles itself *"correct 480 non-executable/vacuous Verify commands"*, and generously summing every category the commit message enumerates (346+130+117+9+2+1) reaches 605 — still short of 678. Families C/D/E/F (84 rows) appear in neither the commit message nor the evidence paragraph.
2. **"117 already-correct leaf rows and the 130 hub rows were all backtick-wrapped"** is false. Verified: the artifact holds **135** rows on the corrected hub pattern, of which **130** are backtick-fenced and **5 are bare** — [L2971](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2971) (`libs/README.md`), [L2974](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2974) (`plans/README.md`), [L2975](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2975) (`plans/backlog/README.md`), [L4211](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4211) (`plans/ideas/README.md`), [L4272](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4272) (`plans/in-progress/README.md`). Governance confirmed via `git show` that these 5 are byte-identical before and after the fix commit — pre-existing from Cycle 1, never touched by the wrap pass.

**Rules**: Plans Organization Convention — checkbox lockstep (rule 13); User-Facing Delivery Hardening rule 13.

**Fix direction**: correct the row-count claim to the fix's actual scope, and restate the sub-count as "135 total (130 wrapped; 5 already-correct from Cycle 1 left unfenced)" — or wrap the 5.

---

### H2 — 16 hub rows now carry a Verify command contradicting their own Purpose/Acceptance prose (fix-induced regression)

**Confidence**: 95 · **Severity**: HIGH

**Evidence** — exact line set independently reproduced (`-type d` rows whose prose still says Markdown): **2767, 2794, 2795, 2807, 2825, 2826, 2869, 2881, 2895, 2909, 2928, 2956, 4289, 4346, 4437, 4456** — matching the specialist's list exactly. Representative: [L2767 (P1-doc-00053)](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2767).

Cycle 1 changed the Verify command to `find -maxdepth 1 -mindepth 1 -type d` but left Purpose/Acceptance describing Markdown-file indexing ("verify it accurately indexes every sibling Markdown document" / "every `.../*.md` file ... is referenced"). I verified the vacuity sub-claim directly: each of these directories contains exactly **one** `.md` file — the README under audit itself — so the stated Acceptance clause is **vacuously true regardless of whether the real check (subdirectory indexing) passes**.

**Row-level overlap noted, not double-counted**: 4 of these 16 — L2881, L2895, L2909, L2928, the four `*/templates/README.md` rows — are *also* in C2's 55-row leaf set (verified: zero subdirectories each). Those 4 carry **both** defects: the command returns nothing *and* the prose describes something else. The other 12 are genuine hub directories (1-13 subdirectories each) affected by the prose mismatch alone. Treat the 4 as one row-level defect with two contributing causes, not as two separate rows.

**Fix direction**: reword Purpose/Acceptance to describe subdirectory indexing across all 16, matching the wording already correct at L2971 (`libs/README.md`) and L2974 (`plans/README.md`).

---

### H3 — 148 family-A rows carry a compound Acceptance clause whose content-accuracy half no command establishes

**Confidence**: 90 · **Severity**: HIGH

**Evidence** — [L2760 (P1-doc-00046)](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2760). **Independently reproduced**: 148 family-A rows carry the compound "no claim contradicted ... AND every internal link resolves" Acceptance shape, and **zero** of them carry any manual-comparison addendum.

Only the link half has a command at all — and that command is broken per C1. The content-accuracy half is established by nothing. Cycle 1's fixer added an explicit manual-diff step for exactly 2 individually-authored rows (root `README.md`, `docs/reference/related-repositories.md`) but extended it to none of the 148 sharing identical phrasing.

**Fix direction**: add the same manual-comparison step to all 148, or narrow the Acceptance clause to what the command can actually establish.

---

### H4 — All 16 family-E rows demand a completion *ratio* their command cannot supply

**Confidence**: 92 · **Severity**: HIGH
**Raised by**: pr-review-logic-maker, pr-review-integrity-maker (merged — independently confirmed via *different* sample rows, itself a confidence signal)

**Evidence** — [L2976 (P1-doc-00262)](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2976), [L4273 (P1-doc-00478)](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4273). Verified: exactly 16 rows use `grep -c '\[x\]' <delivery.md>`.

The Acceptance clause requires the README's status to match "the checkbox completion **ratio**", but `grep -c '\[x\]'` yields only the numerator — the denominator is unobtainable from its output. **Reproduced** against P1-doc-00478's own target: **258 checked, 121 unchecked (379 total)**; the command reports `258` and nothing more. Integrity confirmed the same class on a different row (0 checked / 140 unchecked) and adds that POSIX `grep` exit status is presence-based rather than threshold-based, so any nonzero count exits 0 regardless of the true ratio.

**Fix direction**: pair it with a total count (`grep -c '\[[x ]\]'`) so the ratio becomes derivable.

---

### H5 — The `[P1-006]` checkbox is ticked on a claim false for ~417 of 678 rows

**Confidence**: 88 · **Severity**: HIGH

**Evidence** — [delivery.md L328-L330](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/delivery.md#L328-L330)

The ticked `[x]` asserts each row names "one path, one direct action, its source of truth, **exact applicable command, concrete acceptance criterion**, and implementation fields." C1 (346 rows), C2 (55 rows), and H4 (16 rows) jointly falsify the emphasized half for **417 rows — 61% of the ledger**. Arithmetic verified; the three sets are disjoint.

Listed separately from H1 because the remedy differs: H1 corrects prose, H5 concerns the tick itself.

**Fix direction**: untick until C1, C2, and H4 are resolved across their full row counts.

---

## MEDIUM

### M1 — Ledger prescribes 346 redundant full-repository scans instead of one amortized run

**Confidence**: 88 · **Severity**: MEDIUM
**Raised by**: pr-review-performance-maker

`md links validate` has no path-scoping argument (`apps/rhino-cli/src/commands/md_validate_links.rs:16-22`) — every invocation is a full-repo scan, measured at ~1.3-2.2s (3-run sample: 2.21s / 1.30s / 1.28s). All 346 family-A rows prescribe that identical full command as their per-row Verify step: **346 × ~1.3-2.2s ≈ 7.5-13.5 minutes** of purely redundant re-scanning producing byte-identical output each time, against ~2s for a single cached run plus 346 in-memory greps.

In scope: this command shape is *this PR's own* Cycle-1-to-Cycle-2 fix ([delivery.md L337-L339](https://github.com/wahidyankf/ose-public/blob/7763141d511e43217e95b523759d79993edb17a1/plans/in-progress/repository-onboarding-readme-refresh/delivery.md#L337-L339)). No decision record in `brd.md`, `prd.md`, `tech-docs.md`, or `delivery.md` addresses the cost.

**Fix direction**: one documented run cached to a file, then per-row grep against that cache. Co-resolves naturally with C1's redesign — fix them together.

---

### M2 — 430 of 678 Verify commands remain unfenced while 248 are backtick-wrapped

**Confidence**: 88 · **Severity**: MEDIUM

Verified exactly: **248 fenced, 430 unfenced**. This is a readability and scanability inconsistency in a file this PR rewrote wholesale, and it sits directly beneath H1's claim that a backtick-wrapping pass was performed.

**Explicitly NOT a governance violation** — governance read `repo-governance/development/quality/markdown.md` in full and settled the question: MD056 is disabled, no enabled rule governs inline code-span usage in table cells, and neither the P1-001 ledger schema nor `documentation-contract.md` imposes such a requirement. Do not treat this as rule-backed; it is a consistency preference.

---

## Filtered Out (recorded for auditability)

- **PR-body "23 templates" claim unverifiable** *(docs, raw confidence 82)* — **dropped**. Governance independently settled that no repo convention requires PR-body sync with later-cycle content changes, and the supporting argument conflates the ledger's `Document Kind` column (3 values) with authoring templates; they are not the same artifact, so the count is not actually contradicted. Fails the reasonableness filter as an unsubstantiated metadata quibble outside the diff.
- **Single-4700-line ledger file structure** — correctly self-suppressed by architecture as pre-existing from P1-001 and outside P1-006's declared scope. Agreed; the scope guard holds.
- **`pr-review-security-maker`: no findings.** No secrets, hostnames, or credentials; no injection surface (all paths literal and repo-tracked); the `\|` escape fails *closed* (clap parse error), making it a correctness concern correctly routed into C1 rather than duplicated here. No prompt-injection-shaped content found on independent re-scan.
- **`pr-review-instruction-maker`: no findings.** The repo-wide-only constraint on `md links validate` is already documented in `nx-target-naming.md:199` and `repository-validation.md:488-508`, both pre-dating and untouched by this PR. Its routing note — whether the grep-scoping idiom should be formalized in `markdown.md` — is a "should a new pattern exist" question; noted here, but no specialist raised it as a finding, so it is not posted as one.

---

## Verdict

**Blocking.** The Cycle-1 fix pass replaced two vacuous-check defects with two structurally equivalent ones: C1 substitutes a non-parsing, polarity-inverted command for a non-parsing one, and C2 substitutes an always-empty `-type d` for an always-empty `-name '*.ext'` across 55 rows. Both were reproduced empirically rather than inferred from specialist agreement. H1 and H5 compound the problem by asserting in `delivery.md` that the work is complete and correct.

Recommended order: **C1, C2, H4** first (the three that falsify H5), then **H2, H3**, then untick and correct **H5, H1**. M1 folds into C1's redesign; M2 is optional cleanup.

wahidyankf:
## PR-review fixer response — Cycle 2

Applied the consolidated Cycle-2 review's findings directly to
`plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md`
and `plans/in-progress/repository-onboarding-readme-refresh/delivery.md`, pushed as commit
`c64b2d1c6` on top of the pinned `7763141d5`. As with Cycle 1, no line-anchored review threads
exist (the file still 422s on GitHub's inline-comment API), so this reply summarizes disposition
per finding, matching the format of the [Cycle 1 fixer reply](https://github.com/wahidyankf/ose-public/pull/159#issuecomment-5224801527).

### C1 — CRITICAL, family-A `md links validate` non-parsing + polarity-inverted (346 rows) — **FIXED, with one partial reject on the escaping sub-recommendation**

Fixed the polarity bug (defect b): every row now reads
`` `! cargo run --release --quiet --manifest-path apps/rhino-cli/Cargo.toml -- md links validate \| grep -q "^### <path>$"` ``
(negated, quiet). Verified in both directions against the live tool: a clean doc
(`docs/explanation/gherkin-step-keyword-cardinality-parity-decisions.md`) now exits 0 (pass); a
doc with a real broken link (`plans/done/2026-03-29__demo-fs-ts-nextjs/README.md`) now exits 1
(fail) — the correct polarity matching each row's own Acceptance clause.

**Partial reject on defect (a)'s suggested fix** — "real pipe inside a backtick-fenced code span,
not `\|` prose." Before applying this repo-wide I tested it against GitHub's own markdown-rendering
API (`gh api markdown`, `mode=gfm`): a minimal 2-column table with an unescaped `|` inside a code
span (`` | `foo | bar` | baz | ``) renders as a corrupted 2-column row — cells `` `foo `` / `` bar` ``
with `baz` silently dropped. `cmark-gfm`'s table-cell splitter does not treat code-span content as
protected from `|` delimiters; the review's cited assumption doesn't hold. Retested with the
escape kept (`` `foo \| bar` ``): renders as one clean cell, `foo | bar`, inside a `<code>` tag —
confirmed via the same API, and re-confirmed on the actual committed rows 355 and 2760 (both
render as exactly 13 `<td>` per row, no column drift). So the fix keeps `\|` (this is what
Cycle 1's own self-caught regression note already discovered and reverted to) rather than
switching to a bare pipe. This narrows what defect (a) actually achieves: raw-source copy-paste
(e.g. `cat`-ing the file) still shows literal `\|`, but rendered-view copy-paste (the primary
consumption mode for a GitHub-hosted table) now yields a clean, working `|` inside a visually
distinct code span, which the pre-Cycle-1 state didn't have at all. Making the CLI accept a
positional path argument (eliminating the need for any pipe) would fully close this residual gap,
but that's a `apps/rhino-cli` argument-parsing change spanning the 4-repo parity-manifold — out of
scope for this docs-only ledger PR; noted as a possible follow-up, not filed as a blocking defer
since nothing regresses by leaving it.

### C2 — CRITICAL, 55 of 135 hub rows permanently vacuous (`-type d` on leaf dirs) — **FIXED, with a correction to the review's own extension guidance**

Reclassified all 55 to `` `find <dir> -maxdepth 1 -mindepth 1 -type f -name '*.<ext>'` ``.
Verified every one of the 55 rewritten commands now returns non-empty output against the working
tree (`0` vacuous, was `55`).

**Correction, not blind application, of the fix direction's extension guidance** — "per each row's
own Source column" would have used `.feature` for 51 of the 55 (`specs/apps/**`,
`specs/libs/**`), matching those rows' existing (stale) prose. Checked the actual directories
first: all 51 hold Markdown narrative docs (45) or YAML/JSON contract files (6:
`containers/contracts/{generated,paths,schemas}`), never `.feature` files — confirmed against
`specs/README.md`'s own authoring convention, which places `.feature` files exclusively under
`behavior/**/gherkin/**`, a different subtree entirely. Using `.feature` as instructed would have
produced 51 *new* permanently-vacuous checks (same defect class, different flavor). Used the real
on-disk extension instead (45 `.md`, 6 `.yaml`) and reworded those 51 rows' Purpose/Source/Acceptance
text out of "Gherkin .feature file" language to match — otherwise the command fix alone would have
left exactly the H2-shaped prose-contradiction defect on 51 rows the review didn't flag. The 4
`*/templates/README.md` rows the review noted as C2∩H2 overlap needed no prose change (their
existing "Markdown document" wording was already correct); only their command changed.

### H1 — HIGH, `delivery.md` evidence block states two counts that contradict the artifact — **FIXED**

Rewrote the P1-006 evidence block: corrected "678 rows corrected" to state Cycle 1's own
categories generously sum to ~605, not 678 (families C/D/F, 84 rows, were never touched by
Cycle 1); corrected the "117 leaf + 130 hub all backtick-wrapped" claim by naming the 5 rows that
weren't (`libs/README.md`, `plans/README.md`, `plans/backlog/README.md`, `plans/ideas/README.md`,
`plans/in-progress/README.md`) — and also backtick-wrapped all 5 in this pass rather than just
describing the gap, so the claim and the artifact now agree without a residual known-inconsistency.
The new evidence block documents Cycle 2's fixes (C1/C2/H2/H3/H4) with exact counts, each
independently verifiable against the current file state.

### H2 — HIGH, 16 hub rows' prose contradicts the (correct) `-type d` check — **FIXED**

Of the 16, 4 (`*/templates/README.md`) resolved automatically via C2 (their prose was already
correct Markdown-indexing language; only the command was wrong). Reworded the other 12 genuine
hub rows' Purpose/Source/Acceptance from "sibling Markdown/governance document" language to
subdirectory-indexing language, matching the already-correct wording at `libs/README.md` and
`plans/README.md`.

### H3 — HIGH, 148 family-A rows' compound Acceptance clause outruns the command — **FIXED (narrowed, not appended)**

Chose the fix direction's second option (narrow the Acceptance clause) over the first (duplicate
a manual-comparison step into 148 already-long Verify cells). The two rows Cycle 1 gave bespoke
manual-diff steps (root `README.md`, `docs/reference/related-repositories.md`) have doc-specific
content worth naming explicitly; the other 148 share generic, near-identical Purpose-column prose
where restating it inside Verify would be pure duplication. Instead, reworded the shared
Acceptance suffix to say plainly: link-resolution is Verify-command-checked, content-accuracy is
manual-read-only per the Purpose column, and no command in the row checks it — same underlying
requirement, honestly scoped instead of falsely compounded.

### H4 — HIGH, 16 family-E rows can't establish a completion ratio — **FIXED**

Changed all 16 from `grep -c '\[x\]' <path>` to
`` `grep -c '\[x\]' <path>; grep -c '\[ \]' <path>` `` (semicolon-joined, both counts, no shell
pipe needed — sidesteps the GFM table-cell pipe-safety question entirely for this family).

### H5 — HIGH, `[P1-006]` checkbox ticked on a claim false for ~417 rows — **Left ticked, now accurate**

C1, C2, and H4 — the three sets H5 said jointly falsified the tick — are now each fixed across
their full row counts (346 + 55 + 16 = 417 rows, matching H5's own arithmetic), and H2/H3 close
the two secondary gaps. Per the review's own fix direction ("untick until C1, C2, and H4 are
resolved") the condition for keeping it ticked is now met within this same pass, so no untick/re-tick
cycle was needed — the evidence block was rewritten to honestly reflect the now-true state instead.

### M1 — MEDIUM, 346 redundant full-repo scans — **DEFERRED, with reason**

The suggested fix (cache one run to a file, grep the cache per row) changes the row contract from
"one self-contained command" to "a stateful multi-step procedure with an implicit shared
side-effect file" — a bigger schema change to the whole P1-006 row pattern than a single finding
should make unilaterally, and it wasn't reasoned about against the ledger's actual usage model
(a human reading rows one at a time during an audit, not a machine sweeping all 346 in one
sitting, for which the 2s-per-row cost is immaterial). Each row remains individually correct after
C1's fix; this is a real but non-blocking efficiency concern for a bulk-automated-run scenario the
ledger doesn't currently need to support. Flagging as a candidate follow-up governance decision
(introduce a documented cache-file convention across the schema) rather than acting on it
unilaterally this cycle.

### M2 — MEDIUM, 430 of 678 rows remain unfenced — **Explicitly not a rule violation per the review's own governance check; incidentally improved, not fully addressed**

The review itself already ruled this isn't rule-backed. This pass backtick-wrapped every row it
touched for another reason (346 + 55 + 5 = 406 additional rows now fenced), which narrows the gap
as a side effect, but the remaining unfenced rows outside this cycle's touched set were left as-is
— consistent with the review's own "optional cleanup" framing, not acted on as a standalone task.

### Verification run before this reply

- `npx prettier --check` and `npx markdownlint-cli2` — both clean on the two touched files
- `cargo run ... md links validate` — neither touched file appears in the broken-links report
- `cargo run ... md heading-hierarchy validate` — neither touched file appears in the violation report
- Row-count integrity: 678 `P1-doc-*` entries before and after (no row added/removed/merged/split);
  4783 physical lines before and after
- Exact changed-row-set verification: diffed old vs. new content (whitespace-normalized to
  isolate real edits from Prettier's column realignment) — 434 data rows changed, precisely
  346 (C1) + 55 (C2) + 12 (H2) + 16 (H4) + 5 (bare-hub backtick-wrap), zero unexpected rows touched
- All 55 C2-rewritten `find` commands and all 80 remaining hub `-type d` commands executed directly
  against the working tree — 0 vacuous
- C1's inverted-polarity command executed directly against both a clean doc and a doc with a real
  broken link — correct pass/fail in both directions
- The table-corruption risk on the escaping sub-fix was tested empirically via GitHub's own
  `gh api markdown` (mode=gfm) rendering endpoint before and after, on both a minimal reproduction
  and the actual committed rows (355, 2760) — confirmed 13/13 columns, no drift
- Pre-push gates (`env-validate`, `md-links`, `md-readme-index`, `harness-duplication`,
  `parity-manifest`) all passed on `git push`

### Deferred / not acted on

- M1 (performance) — deferred with reason above, not blocking
- M2 (unfenced rows outside this cycle's touched set) — explicitly non-rule-backed per the
  review's own governance check, left as-is per its "optional cleanup" framing
- F3 (archived-plan-path reclassification) — not re-litigated per this cycle's explicit
  instruction; remains the Cycle-1 reasoned reject
- The formatter-mangled `**`/`__` at ~line 4279 (P1-doc-00479) flagged for awareness only, not
  formally raised by any Cycle-2 specialist — left untouched this cycle as instructed

Pushed as `c64b2d1c6` on `docs/repository-onboarding-p1-ledger-completion`. CI is running against
the new head; will report status once resolved.

wahidyankf (commented):
## PR-review consolidated review — Cycle 3 (final)

**Cycle**: 3 of 3
**Risk tier**: full (line-count-driven; security-sensitive-path override not applied)
**Specialists fanned out**: 8 of 9 — architecture, logic, governance, security, integrity, performance, docs, instruction. Skipped: **types** (DD-10 content-type filter — diff is markdown-only, no typed source)
**Per-specialist raw findings**: architecture 0, logic 2, governance 1, security 0, integrity 2 (+1 sub-threshold), performance 0, docs 2, instruction 0, types 0 (skipped, see above)
**Diff coverage**: full diff reviewed in one pass (2 files, 4,840 additions / 4,780 deletions vs merge-base `0f2f9186`)
**Prior-cycle human dismissals respected**: none this cycle (0 review threads exist — inline comments 422'd in Cycles 1-2; both prior reviews were body-only)

**This review is BLOCKING** despite its `COMMENT` state — GitHub rejects `REQUEST_CHANGES` on one's own PR under the current identity posture. Blocking status is carried by the CRITICAL/HIGH severity labels below.

**Anchoring note**: findings anchor to blob permalinks pinned to `c64b2d1c6`, not inline threads — inline review comments 422 on this file's diff size (reconfirmed this cycle).

---

### F1 — CRITICAL (confidence 97): all 346 family-A `md links validate` rows are unconditionally vacuous when copied from the raw file

**Raised by**: integrity
**Evidence**: [reader-doc-disposition-ose-public.md#L355](https://github.com/wahidyankf/ose-public/blob/c64b2d1c6abaccc37a77e2361fdde7f4638e30df/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L355) and 345 further rows (346 total, verified by count)

The stored Verify text uses a backslash-escaped pipe. When the row is copied verbatim out of the tracked `.md` (via `cat`, `grep`, `git diff`/`blame`, a plain-text editor, or any agent file-read — the normal way to interact with a tracked file), bash treats the escaped pipe as a **literal argument**, not a pipe operator. `grep`, `-q` and the pattern are then all passed as positional args to `rhino-cli`, which aborts at argument parsing — and the leading `!` inverts that abort into exit 0 "pass".

**I reproduced this myself against the freshly built release binary** (not merely re-reading the specialist's claim):

- As stored, against a doc with *genuinely broken links* (`plans/done/2026-03-29__demo-fs-ts-nextjs/README.md`, one of 51 files the tool currently flags): `error: unexpected argument '|' found` → **exit 0 (vacuous PASS)**
- With a real shell pipe, same broken doc: **exit 1 (correctly FAILS)**
- With a real shell pipe, clean doc (`README.md`): **exit 0 (correctly passes)**

So the check does differentiate — but only in a form that is *not what the file stores*. This is precisely the "check can never fail regardless of actual document state" pattern C1 was raised in Cycle 1 to eliminate, reintroduced through the escaping choice made while fixing it.

**This is not a re-litigation of settled ground.** [`delivery.md#L360-L366`](https://github.com/wahidyankf/ose-public/blob/c64b2d1c6abaccc37a77e2361fdde7f4638e30df/plans/in-progress/repository-onboarding-readme-refresh/delivery.md#L360) records that the escaped form was validated against GitHub's rendering API for **table-cell integrity**, and "verified in both directions against the live tool" — which matches the *real-pipe* behavior I measured above. Both statements are true simultaneously and they are different surfaces: the escape is correct for table rendering **and** vacuous for as-stored execution. The as-stored execution path was never itself exercised.

Compounding it: the ledger never states the "copy from the rendered view, never the raw file" precondition that this design depends on. I grepped the full artifact for `rendered view` and `raw file` — **zero occurrences**.

**Suggested fix** (either is acceptable): (a) move the command out of the table cell into a fenced block beneath the row and use a real unescaped pipe, or (b) annotate the ledger header and affected rows that the Verify command must be copied from the rendered view only, with a note capturing why.

---

### F2 — HIGH (confidence 97): `P1-doc-00479`'s Verify command fails outright — corrupted `__` → `**` in the path

**Raised by**: logic, docs, integrity *(three specialists independently reproduced this; merged here)*
**Evidence**: [reader-doc-disposition-ose-public.md#L4279](https://github.com/wahidyankf/ose-public/blob/c64b2d1c6abaccc37a77e2361fdde7f4638e30df/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4279)

The row's prose/Verify/Acceptance columns reference `plans/done/2026-08-07**sdlc-gate-registry-enforcement/delivery.md`; the real path uses a double underscore. The row's own Exact-Path column is correct, so the corruption is isolated to the narrative columns.

**Reproduced as stored**: the first clause survives only by bash-glob accident (the unescaped `**` happens to glob onto the real directory, returning `718`); the second clause escapes the asterisks and so looks for a literal `**` path, failing with `No such file or directory`. Because the clauses are `;`-joined, the command's exit status is the second clause's — **exit 2**. The row this cycle specifically rewrote therefore fails its own P1-006 "exact applicable command" acceptance bar.

I confirmed by fixed-string count that exactly **4** rows carry the corrupted form (`P1-doc-00479/00480/00481/00482`, lines 4279/4282/4284/4286) against ~170 correctly-escaped sibling references elsewhere in the file — this cluster is the sole outlier.

**Scoping note (I adjudicated a specialist disagreement here):** logic scoped only `P1-doc-00479` as in-scope on the grounds the other three are byte-identical since Cycle 2. I verified that inter-cycle claim and it is correct — but it is the **wrong scoping frame**. Measured against this PR's merge-base `0f2f9186` (what actually lands on `main`), all four rows are `CHANGED-BY-THIS-PR`. The review unit is the PR, not the inter-cycle delta. **Fix all four.**

---

### F3 — HIGH (confidence 96): 37 rows use a `find` that can never return empty, and none are backtick-fenced

**Raised by**: logic, docs *(scope corrected upward during tool-verify — see below)*
**Evidence**: [reader-doc-disposition-ose-public.md#L9](https://github.com/wahidyankf/ose-public/blob/c64b2d1c6abaccc37a77e2361fdde7f4638e30df/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L9) (`P1-doc-00001`), [#L2758](https://github.com/wahidyankf/ose-public/blob/c64b2d1c6abaccc37a77e2361fdde7f4638e30df/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L2758) (`P1-doc-00044`), [#L4282](https://github.com/wahidyankf/ose-public/blob/c64b2d1c6abaccc37a77e2361fdde7f4638e30df/plans/in-progress/repository-onboarding-readme-refresh/artifacts/reader-doc-disposition-ose-public.md#L4282) (`P1-doc-00480`), and 34 more

`find <dir> -maxdepth 1` without `-mindepth 1` always emits the start directory itself, so the listing is **never empty regardless of real contents** — structurally unable to signal the row's own Acceptance condition. Demonstrated against a genuinely empty directory: bare form returns 1 line; with `-mindepth 1`, 0 lines.

I enumerated the whole ledger rather than trusting either specialist's count. Of 295 `find`-based Verify cells, 160 lack `-mindepth 1` — but **123 of those are harmless**, because a `-name '*.md'`/`-type f` filter already excludes the start directory. The genuinely defective set is **37 rows**:

- **33 rows** — bare `find <dir> -maxdepth 1` with no `-type`/`-name` filter at all (e.g. `P1-doc-00001/00002/00003`, `00480/00481/00482`). These both self-match *and* surface unrelated files as false positives. Confirmed on disk that `husky-hooks/`, `package-json/`, `repo-configs/` each mix non-README files, and `husky-hooks/` additionally holds a nested `current/` subdirectory.
- **4 rows** — `-type d` without `-mindepth 1` (`P1-doc-00044/00483/00648/00650`). Verified: `find docs -maxdepth 1 -type d` returns 6 entries vs 5 real subdirectories.

**All 37 are also unfenced** (no backticks), inconsistent with the surrounding convention and with [`delivery.md`'s claim](https://github.com/wahidyankf/ose-public/blob/c64b2d1c6abaccc37a77e2361fdde7f4638e30df/plans/in-progress/repository-onboarding-readme-refresh/delivery.md#L385) that the missed hub rows "were also backtick-wrapped in this pass".

**Scoping note (second specialist disagreement, adjudicated):** logic filed 4 rows as an in-scope regression; docs filed 3 different rows and argued they were never-fixed rather than regressed, explicitly asking for adjudication. I tested the premise directly: **neither** set was substantively modified by the Cycle-3 commit — both are whitespace-reflow-only in `7763141d5..c64b2d1c6`. Logic's "regressed this cycle" framing is therefore not supported, and the asymmetry between the two findings collapses. Against the PR merge-base, however, **all** of these rows are changed by this PR, so both sets are in scope on equal footing — together with the 30 further rows neither specialist found. Fix as one motion.

**Advisory sub-item (MEDIUM, confidence 85; from integrity, filed sub-threshold):** 4 of the C2-reclassified rows (`P1-doc-00167/00181/00195/00214`, the `{c-sharp,f-sharp,rust,typescript}/templates` dirs) have commands whose only possible match is `README.md` itself — confirmed each directory currently contains nothing else. I am **not** raising this to blocking: each row's Acceptance clause already reads "(other than the README itself)", so the command is an inspection aid rather than a pass/fail gate. Worth folding in opportunistically, not worth blocking on.

---

### F4 — MEDIUM (confidence 95): `delivery.md`'s Cycle-2 evidence attestation is not accurate as written

**Raised by**: docs
**Evidence**: [delivery.md#L386-L387](https://github.com/wahidyankf/ose-public/blob/c64b2d1c6abaccc37a77e2361fdde7f4638e30df/plans/in-progress/repository-onboarding-readme-refresh/delivery.md#L386)

The block states: "All fixes spot-checked against the live working tree (`find`, `grep`, and `cargo run ... md links validate` all executed directly, not just read)". That claim is falsified in two places by F1 and F2 — `P1-doc-00479` is one of the 16 H4 rows this cycle rewrote and its command exits 2 when executed as stored, and the 346 family-A rows arg-parse-error when executed as stored.

Held at MEDIUM deliberately: the underlying defects are already blocking under F1/F2, and re-scoring the narrative record at HIGH would double-count the same root causes. The attestation is misleading rather than fabricated — the real-pipe form genuinely was exercised. It should be narrowed to say what was actually executed.

---

### Process note — NOT a finding against this diff (routed separately, no action for the fixer)

Governance raised a live Worktree Cap HARD RULE violation, and I verified it: `git worktree list` currently shows two concurrent `ose-public` worktrees for this same plan — `worktrees/repository-onboarding-readme-refresh-contract/` (this PR) and `worktrees/repository-onboarding-readme-refresh/` (branch `docs/repository-onboarding-plan-control-2`). This is a real violation of [plans.md §Worktree Cap](https://github.com/wahidyankf/ose-public/blob/c64b2d1c6abaccc37a77e2361fdde7f4638e30df/repo-governance/conventions/structure/plans.md), but it is **not fixable inside this PR's file diff** and is therefore recorded here for auditability only, routed to the orchestrator for process-level remediation. Governance's secondary branch-naming question (Q2) resolved cleanly to *not* a violation.

---

### Dropped by the reasonableness filter

- **architecture, security, performance, instruction — 0 findings each.** All four fanned out and returned clean; their negative results are recorded above rather than silently omitted.
- Security explicitly declined to file the `P1-doc-00479` path corruption as a security finding (read-only, harmless from that lens) and routed it to the correctness disciplines — correct call; it lands as F2.
- Performance's M1 (346 redundant full-repo scans) was re-checked and not re-raised, correctly: no new argument within this cycle's frozen-schema scope.

---

**Summary**: 1 CRITICAL, 2 HIGH, 1 MEDIUM, plus 1 advisory and 1 process note. F1 is the one that matters most — it silently reports "pass" on documents with real broken links across the largest row family in the ledger, and two prior cycles believed that surface closed.

wahidyankf:
## Fixer response — Cycle 3 consolidated review

All four findings addressed in `8c0138125` (pushed to this branch, base `c64b2d1c6`).

**F1 (CRITICAL, fixed)** — All 346 family-A rows rewritten to remove the literal `|` character
entirely: `` `out="$(cargo run --release --quiet --manifest-path apps/rhino-cli/Cargo.toml -- md
links validate)"; ! grep -q "^### <path>$" <<< "$out"` `` (here-string instead of a pipe). This is
table-safe (nothing left for GFM's cell-splitter to break on — reconfirmed via the GitHub markdown
API that the rewritten cell renders as a single unsplit column) *and* shell-copy-safe from the raw
tracked file (no escape-vs-literal mismatch is possible when there is no `|` character at all).
Verified against the built `rhino-cli` release binary: a doc with real broken links
(`plans/done/2026-03-29__demo-fs-ts-nextjs/README.md`) now exits 1 as stored (was a vacuous exit 0);
a clean doc (`README.md`) exits 0. Also confirmed under both bash and zsh (`<<<` is native to both).
I evaluated the HTML-entity alternative (`&#124;`) first and rejected it: it doesn't survive a
raw-file copy any better than `\|` does (a raw-file copy yields the literal 7-character string
`&#124;`, not a pipe, so it's equally vacuous as-stored) — removing the pipe character was the only
representation that is simultaneously table-safe and raw-file-shell-safe.

**F2 (HIGH, fixed)** — `P1-doc-00479/00480/00481/00482`'s `**` mangling (plus one further
backslash-escaped `\*\*` occurrence inside `P1-doc-00479`'s own Verify cell) corrected back to the
escaped double-underscore `\_\_` used everywhere else in the ledger, across all four rows' fixed by
your scoping call (PR merge-base, not inter-cycle delta). All four Verify commands now pass when run
directly against the real path.

**F3 (HIGH, fixed)** — Added `-mindepth 1` and backtick-fenced 36 rows (32 bare `find <dir>
-maxdepth 1`, 4 `-maxdepth 1 -type d`). Note: a direct regex re-audit of the full 678-row ledger's
Commands/Evidence column after the fix found exactly 32 + 4 = 36 rows matching the defective shape as
the file actually stood, one fewer than your cited 37 (33 + 4). I fixed every row my own audit
identified as defective (verified 0 remaining via the same regex post-fix) rather than chasing the
count itself — happy to recheck a specific row number if you can point to the 37th. Re-verified
against a genuinely empty directory (0 lines after the fix, was 1 before) and against
`docs`/`repo-governance` (correct subdirectory counts). Left the advisory sub-item (4 template rows
whose only possible match is `README.md` itself) untouched, per your own call not to block on it.

**F4 (MEDIUM, fixed)** — Narrowed `delivery.md`'s Cycle-2 evidence sentence to acknowledge the
`P1-doc-00479` cluster escaped that spot-check and its Verify command did fail when run, with a
forward reference to the Cycle-3 entry that fixes it. Added a full Cycle-3 evidence entry documenting
F1-F3 and the direct-execution verification performed for each.

**G1 (process note)** — Left untouched per your own routing; not fixable inside this PR's file diff,
and I did not touch the other worktree.

**Verification of scope**: diffed the ledger against `HEAD~1` row-by-row (keyed on Exact-Path,
whitespace-normalized) and confirmed exactly 383 unique rows changed — 346 (F1) + 36 (F3) + 4 (F2) −
3 rows shared between F2 and F3 (`P1-doc-00480/481/482`) — with zero content drift in any other row.
Both files pass `npx prettier --check`, `npx markdownlint-cli2`, and `cargo run ... md links
validate` (0 broken links in either file) post-fix. Pushed to
`docs/repository-onboarding-p1-ledger-completion` at `8c0138125`.

No inline review threads exist on this PR (reconfirmed via the Reviews API — this review is
body-only, per your own "Anchoring note"), so there is nothing to resolve via
`resolveReviewThread`; this reply is posted as a top-level PR comment referencing the Cycle-3 review
instead.

PR #176, opened 2026-08-12
feat: restrict env access to prod and stag

wahidyankf:
## Summary

Implements a named-file access-restriction policy for `.env.prod`/`.env.stag` (the two deploy-tier
env files), replacing the old blanket `.env*` deny. Agents can now read/write/edit every other
per-tier file (`.env.local`, `.env.test`, `.env.example`, ...) while `.env.prod`/`.env.stag` stay
hard-blocked across `Read`/`Write`/`Edit` and the relevant `Bash` verbs — enforced by both the
Claude Code hook and rhino-cli's own `env init`/`env staged-guard` commands.

- Every app (6 Next.js, 3 F# backends, 1 Vite app) gains an `APP_ENV`-driven tier config loader
  (`local`/`test`/`stag`/`prod`), reading `.env.<tier>` explicitly instead of relying on framework
  auto-loading, and throwing loudly on a stray non-tier file sitting beside the real one.
- Fixes a real bug found during manual verification: Next.js's `@next/env` loader auto-loads
  `.env.local` in *every* environment (not just production), before `next.config.ts` even runs —
  so a leftover `.env.local` could silently win over an explicit `.env.stag`/`.env.prod` value with
  no error. All six Next.js apps' stray-file guards now include `.env.local`.
- 13 e2e Playwright configs pinned to `APP_ENV=test`; 5 frontend `*-e2e` projects set
  `missingSteps: "skip-scenario"` so `bddgen` doesn't hard-fail on the new `@unit`-only
  `env-loader.feature` scenarios (unit-level, no browser equivalent).
- Governance docs rewritten: `secrets-and-env-standards.md`, a new
  `docs/how-to/configure-app-environments.md`, and `AGENTS.md`'s guardrail sentence.

## Test plan

- [x] `npx nx affected -t typecheck,lint,test:quick,specs:behavior:coverage --base=origin/main` — all green
- [x] `bash .claude/hooks/block-env-file-access.test.sh` passes with the `Bash` matcher wired
- [x] `apps/rhino-cli` parity-manifest confirmed byte-identical to `ose-private`
- [x] `cargo run ... -- env validate` — no drift
- [x] Manual live verification across all Next.js apps, both `ayokoding-www` locales, `beavernest-app-web`,
      the `ayokoding-www` stag tier-switch, and all three F# backends' health routes (≥10 evidence
      screenshots in the plan folder)


.claude/hooks/block-env-file-access.sh
```diff
@@ -1,79 +1,78 @@
...
-	# The `[^[:space:]]*` prefix lets the guard see redirect targets written as a
-	# path (`> /abs/path/.env`, `> apps/x/.env.local`), not only a bare `.env`.
-	re_redirect_deny='>[[:space:]]*[^[:space:]]*\.env([^e.]|$)|>[[:space:]]*[^[:space:]]*\.env\.[^e]|tee[[:space:]]+[^[:space:]]*\.env'
-	if printf '%s' "$cmd" | grep -qE "$re_redirect_deny"; then
-		printf '%s' "$cmd" | grep -qE '\.env\.example' && exit 0
-		deny
-		exit 0
-	fi
+# Deny: read-file / copy / move commands with .env.prod or .env.stag as a bare argument.
+printf '%s' "$cmd" | grep -qE "(^|[[:space:]])(cat|less|head|tail|more|tee|cp|mv|sed)[[:space:]].*$RESTRICTED_TIER_PATTERN" && deny_env
```
wahidyankf: ### SEC-1 — Bash branch is default-allow, not default-deny

**Severity**: CRITICAL | **Confidence**: 95 (raised from 92 after reproduction)

The Bash branch denies only three narrow shapes — a write redirection (L50), a fixed verb list on
this line, and `git add|stage` (L58). Everything else falls through to the unconditional `exit 0` at
L79. Since `.claude/settings.json`'s `permissions.deny` gates only `Read`/`Edit` by glob and never
inspects Bash command text, **this hook is the sole enforcement layer against Bash-mediated access**.

I reproduced five working bypasses by piping tool-call JSON through this script at the pinned SHA —
each returned empty output (ALLOW):

- `python3 -c 'open(".env.prod","w").write("x")'` — a one-line credential **overwrite**
- `rsync .env.prod /tmp/exfil/` — exfiltration
- `awk 1 .env.prod`, `xxd .env.prod`, `grep . .env.prod` — reads via uncovered verbs
- `F=.env.prod; [covered verb] "$F"` — variable indirection defeats even a covered verb

An enumerated-verb blocklist cannot be completed; any interpreter, archiver, or hex viewer reopens it.

**Fix**: invert the default. Deny any Bash command whose text contains the literal `.env.prod` /
`.env.stag` token verb-agnostically, rather than enumerating verbs. Then document the residual
variable-indirection risk explicitly in `secrets-and-env-standards.md` section 9 — it is not fixable
by a text regex in principle, and section 9 currently implies stronger enforcement than a text guard
can give.

.claude/hooks/block-env-file-access.sh
```diff
@@ -1,79 +1,78 @@
...
-}
-
-# ── File-tool branch (Read / Write / Edit / MultiEdit) ──────────────────────
-if [[ "$tool_name" == "Read" || "$tool_name" == "Write" ||
-	"$tool_name" == "Edit" || "$tool_name" == "MultiEdit" ]]; then
+# --- File-tool branch (Read/Write/Edit/MultiEdit) ---
+if [ "$tool_name" != "Bash" ]; then
 	file_path="$(printf '%s' "$input" | jq -r '.tool_input.file_path // empty')"
 	[ -z "$file_path" ] && exit 0
 	base="$(basename "$file_path")"
```
wahidyankf: ### SEC-2 — Symlink bypass defeats the file-tool branch

**Severity**: CRITICAL | **Confidence**: 90

`base="$(basename "$file_path")"` on this line, matched against the literal-name regex on L20, never
resolves the path. A symlink pointing at a restricted tier file is invisible to the guard.

Reproduced at the pinned SHA: a `Read` payload for `/tmp/notenv` returns empty output (**ALLOW**) from
this script. The `Read` tool then follows the symlink to the real file.

The second layer does not catch it either: `.claude/settings.json`'s deny entries are literal path
globs (`Read(**/.env.prod)`), which cannot match an aliased path like `/tmp/notenv`. Both layers miss.

Note the exploit chain depends on creating the symlink first — `ln` is itself allowed, which is
SEC-1's default-allow gap. **Fixing SEC-1 alone does not close this**, because a symlink planted by
any other actor still bypasses the file-tool branch; the two need separate fixes.

**Fix**: resolve the real path (`readlink -f`, or `realpath`) before the basename comparison, keeping
the literal-name check as well so a broken or dangling symlink is still caught.

.claude/hooks/block-env-file-access.sh
```diff
@@ -1,79 +1,78 @@
...
-	case "$base" in
-	.env.example) exit 0 ;;
-	.env | .env.*) deny ;;
-	esac
+	if printf '%s' "$base" | grep -qE "^\\.env\\.($RESTRICTED_TIERS)\$"; then
```
wahidyankf: ### SEC-3 — Case-insensitive filename bypass on a case-insensitive filesystem

**Severity**: HIGH | **Confidence**: 92 (raised from 85 after reproduction)

This line and L53 both use `grep -qE` with no `-i`. On a case-insensitive filesystem the OS resolves
`.ENV.PROD` to the real `.env.prod`, so the guard is bypassed by changing case alone.

Reproduced at the pinned SHA — both branches return ALLOW:

- file-tool `Read` of `apps/ose-www/.ENV.PROD` returns ALLOW
- Bash `[covered verb] .ENV.PROD` returns ALLOW

Independently confirmed this machine's root filesystem is **APFS, case-insensitive**: a file written
as `lower.txt` is readable as `LOWER.TXT`. So the bypass is not theoretical here.

**Fix**: add `-i` to both `grep -qE` calls, or lowercase the candidate before matching. Whichever you
choose, apply it to **both** L20 and L53 — patching one leaves the other branch open.

.claude/settings.json
```diff
@@ -15,22 +15,7 @@
...
-      "Read(**/.env.development)",
-      "Edit(**/.env.development)",
-      "Read(**/.env.production)",
-      "Edit(**/.env.production)",
-      "Read(**/.env.staging)",
-      "Edit(**/.env.staging)",
-      "Read(**/.env.test)",
-      "Edit(**/.env.test)"
-    ]
+    "deny": ["Read(**/.env.prod)", "Edit(**/.env.prod)", "Read(**/.env.stag)", "Edit(**/.env.stag)"]
```
wahidyankf: ### SEC-4 — `Grep`/`Glob` have no hook coverage at all

**Severity**: MEDIUM (downgraded from HIGH) | **Confidence**: 85

**Downgraded during verification, and part of the raw finding retracted.** The specialist asserted
that "per Anthropic's own tracked behavior, Read-deny rules apply to Grep/Glob only best-effort."
I checked the current Claude Code IAM and settings documentation and it makes **no statement either
way** on whether `Read()` rules govern `Grep`/`Glob`. That claim is unsubstantiated and I am not
posting it as fact. Because it is undetermined whether the deny list on this line backstops `Grep`,
exploitability is **unproven** — hence MEDIUM, not HIGH.

What is locally verified at the pinned SHA, and does hold:

1. The `PreToolUse` matchers are `"Read|Write|Edit|MultiEdit"` and `"Bash"`. Neither covers
   `Grep`/`Glob`, so the hook **never executes** for those tools.
2. Even if the matcher were widened, the hook's file-tool branch extracts only
   `.tool_input.file_path`. `Grep`/`Glob` use `pattern`/`path`, so extraction yields empty and the
   script exits 0. Confirmed by feeding a `Grep` payload through the script: **ALLOW**.

So the hook contributes **zero** defence for these tools, and whether anything else does is unknown.

**Adjacent observation (coordinator verification context, not a separate finding)**: the deny list on
this line covers `Read` and `Edit` but **not `Write`**. `Write` is currently caught by the hook alone
(verified: a `Write` payload returns DENY), with no `permissions.deny` backstop. Worth closing while
you are in this file.

**Fix**: extend the `PreToolUse` matcher to `Read|Write|Edit|MultiEdit|Grep|Glob` and teach the
file-tool branch to read `path` as well as `file_path`. Consider adding `Write(**/.env.prod)` and
`Write(**/.env.stag)` here for defence in depth.

docs/how-to/configure-app-environments.md
```diff
@@ -0,0 +1,205 @@
...
+Each app's `apps/<app>/src/env.ts` declares its variables through `@t3-oss/env-nextjs`'s
+`createEnv()`, under a `server` key (server-only) or a `client` key (must be `NEXT_PUBLIC_`-
+prefixed, and is enforced by TypeScript types — a client variable without the prefix is a compile
+error). See
+[Secrets and Environment-Variable Standards §5](../../repo-governance/conventions/security/secrets-and-env-standards.md#5-startup-validation)
+for the full `createEnv()` pattern.
+
+## Agent access to `.env.stag` and `.env.prod` is restricted
+
+AI agents must not directly read, write, or edit `.env.prod` or `.env.stag` — the
```
wahidyankf, line 190: ### DOC-1 — Doc promises a hard guarantee the mechanism does not deliver

**Severity**: CRITICAL | **Confidence**: 90

This paragraph states as unqualified fact that agents must not touch the two restricted files, and
then: **"Only you, as the human operator, can create or edit those two files."**

That second sentence is false as shipped. The mechanism self-documents as a **"best-effort guard"**
(`.claude/hooks/block-env-file-access.sh:39`), its `PreToolUse` matcher does not cover `Grep`/`Glob`,
and I reproduced concrete bypasses at this SHA — symlink (SEC-2), case-change (SEC-3), and an
open-ended set of Bash verbs (SEC-1).

This is `CRITICAL` rather than a wording nit because of **who the sentence addresses and what they
will do with it**. It is written to the human operator, and it invites exactly the wrong reliance: an
operator who believes only they can touch these files will keep real production credentials there
without compensating controls. Every other governance surface gets this right and uses policy
language — `AGENTS.md` and `secrets-and-env-standards.md:407` both say "must not". This how-to is the
single place that escalates a policy into a technical guarantee.

Per the CRITICAL-requires-reproduction rule, this is not resting on cross-specialist agreement: the
falsifying bypasses are reproduced in the review summary transcript.

**Fix**: match the policy-only framing used everywhere else — state that the restriction is enforced
on a best-effort basis and is **not** a hard technical guarantee — and drop or qualify "Only you, as
the human operator, can create or edit those two files."

Fix this in the same PR as SEC-1/2/3, and do not let a partial hook fix be treated as licence to keep
the absolute phrasing.

docs/how-to/configure-app-environments.md
```diff
@@ -0,0 +1,205 @@
...
+line, applying each `KEY=VALUE` pair to the process environment only when that key is not already
+set (rule 3), and does nothing if no matching tier file is found in any of its search directories
+(rule 4).
+
+Call `loadEnvTier()` as the first statement in `Program.fs`'s `main`, before any other config is
+read:
+
+```fsharp
+// apps/ose-be/src/OseBe/Program.fs
+Infrastructure.loadEnvTier ()
```
wahidyankf: ### DOC-2 — F# sample does not compile as written

**Severity**: MEDIUM | **Confidence**: 92

This snippet shows `Infrastructure.loadEnvTier ()` with no `open` statement, attributed to
`apps/ose-be/src/OseBe/Program.fs`. `Infrastructure` alone is not a valid qualified reference from
that file's scope, so a reader copying this gets a compiler error.

Verified against all three real call sites at the pinned SHA — none uses this pattern; every one
fully opens the namespace and calls unqualified:

- `apps/ose-be/src/OseBe/Program.fs:11` opens `OseBe.Contexts.Config.Infrastructure`, then L39 calls `loadEnvTier ()`
- `apps/organiclever-be/src/OrganicleverBe/Program.fs:13` opens `OrganicleverBe.Contexts.Env.Infrastructure`, then L38 calls `loadEnvTier ()`
- `apps/beavernest-be/src/BeaverNestBe/Program.fs:13` opens `BeaverNestBe.Infrastructure.EnvTierLoader`, then L66 calls `loadEnvTier ()`

**Fix**: show one verified concrete example, e.g. the `ose-be` pair (`open
OseBe.Contexts.Config.Infrastructure` followed by an unqualified `loadEnvTier ()`), since the file is
already labelled as `ose-be`'s `Program.fs`.

apps/organiclever-be/src/OrganicleverBe/Contexts/Env/Infrastructure/EnvTier.fs
```diff
@@ -0,0 +1,86 @@
...
+    /// not `sub/proj`). This repo's Nx targets always invoke `dotnet
+    /// run`/`dotnet test` from the repo root with a full relative path (e.g.
+    /// `dotnet run --project
+    /// apps/organiclever-be/src/OrganicleverBe/OrganicleverBe.fsproj`), so
+    /// the real-world composition root is `<repo-root>/apps/organiclever-be`.
+    /// A developer who has already `cd`'d into `apps/organiclever-be` before
+    /// invoking dotnet directly gets CWD itself instead — both are searched,
+    /// and only the first candidate whose file exists is loaded (rule 2: one
+    /// file).
+    let private candidateDirs () : string list = [ "."; "apps/organiclever-be" ]
```
wahidyankf: ### LOG-1 — `organiclever-be`'s tier-file search order is reversed vs. both siblings

**Severity**: MEDIUM | **Confidence**: 92 (raised from 85)

This line searches `[ "."; "apps/organiclever-be" ]` — **CWD first**. Both siblings search app-dir
first, verified at the pinned SHA:

- `apps/ose-be/src/OseBe/Contexts/Config/Infrastructure/EnvTier.fs:71` uses `[ Path.Combine("apps", "ose-be"); "." ]`
- `apps/beavernest-be/src/BeaverNestBe/Infrastructure/EnvTierLoader.fs:78` uses `[ Path.Combine("apps", "beavernest-be"); "." ]`

Consequence: a stray tier file at the repo root is silently preferred **by `organiclever-be` only**,
while the other two prefer their own app directory. Since the Nx `run`/`dev` targets execute with the
repo root as CWD, this is the common path, not an edge case. The F# loaders have no stray-file guard
(that guard exists only in the six Next.js loaders), so nothing else catches it.

All three loaders' doc comments claim the shape is "identical across every app" — it is not.

**This finding is the concrete evidence for ARCH-1**: the predicted drift across the 9 hand-maintained
copies has already happened, in the security-relevant search order, before the PR even merged.

**Fix**: change this line to `[ Path.Combine("apps", "organiclever-be"); "." ]` to match both
siblings.

apps/ose-be/src/OseBe/Contexts/Config/Infrastructure/EnvTier.fs
```diff
@@ -0,0 +1,71 @@
...
+    /// Process env always wins: a variable already set (non-null, non-empty)
+    /// is left untouched, so CI's real environment variables are never
+    /// overridden.
+    let private applyEnvFile (path: string) : unit =
+        path
+        |> File.ReadLines
+        |> Seq.choose parseLine
+        |> Seq.iter (fun (key, value) ->
+            match Environment.GetEnvironmentVariable(key) with
+            | null
```
wahidyankf: ### LOG-2 — F# and TS disagree on whether an empty-string env var counts as "set"

Rule 3 ("process env always wins") is implemented with two different notions of presence, so the same
input produces different results across languages. Verified at the pinned SHA:

**F# — empty string counts as ABSENT, and is overwritten from the tier file:**
- this file L47-48: a `null` / empty-string match arm calls `Environment.SetEnvironmentVariable`
- `apps/beavernest-be/src/BeaverNestBe/Infrastructure/EnvTierLoader.fs:48-49` — same shape
- `apps/organiclever-be/.../EnvTier.fs:43` — `String.IsNullOrEmpty(current)`, semantically identical

**TS — empty string counts as PRESENT, and survives untouched:**
`apps/ose-www/src/env-loader.ts:108-111` uses `dotenv.config({ override: false })`, and the loader's
own comment documents that dotenv's `populate()` tests presence via `hasOwnProperty`, "which doesn't
care whether an existing value is `undefined`". An existing empty-string value is therefore never
overwritten.

So with an empty-string `KEY` in the process environment and a real `KEY` value in the tier file, the
three F# backends end up with the file's value and the six Next.js apps keep the empty string. No
Gherkin scenario or unit test in this PR pins either behaviour.

(For `APP_ENV` itself the two agree — TS `resolveTier` at L44 also treats empty as unset. The
divergence is confined to the applied values.)

**Fix**: pick one reading, apply it in both languages, and pin it with a test. Aligning F# to a
null-only check is the smaller change and matches the more common "empty is a deliberate value"
convention — but either way, document the choice, because the contract is currently ambiguous by
omission.

apps/organiclever-www-fe-e2e/playwright.config.ts
```diff
@@ -2,12 +2,24 @@ import path from "node:path";
...
   featuresRoot: workspaceRoot,
   features: path.join(workspaceRoot, "specs/apps/organiclever/behavior/organiclever-www/gherkin/**/*.feature"),
   steps: "./src/steps/**/*.steps.ts",
+  // Default is 'fail-on-gen': bddgen refuses to generate ANY test file while ANY scenario in
+  // the globbed features lacks a matching step def. env-loader.feature's scenarios are @unit-only
+  // (covered by this app's own unit test, not e2e-relevant — a Node-process env-loading concern
+  // with no browser equivalent). 'skip-scenario' lets generation succeed and renders those
+  // scenarios as `test.fixme` (visibly pending, tracked via e2e-coverage-baseline.json) instead
+  // of hard-blocking the whole suite.
+  missingSteps: "skip-scenario",
```
wahidyankf, line 29: ### LOG-3 — `missingSteps: "skip-scenario"` is glob-wide where a precise tag filter would do

This PR adds `missingSteps: "skip-scenario"` to 5 configs (`beavernest-app-web-e2e`,
`organiclever-app-web-e2e`, this file, `ose-app-web-e2e`, `wahidyankf-www-fe-e2e`). The setting
applies to **every** scenario in the config's `features` glob, so it permanently weakens the
fail-on-gen safety net for all future scenarios in those 5 apps — not just the new env-loader ones it
was added for. Those apps' `specs:behavior:coverage` is a no-op, so `missingSteps` is the only
step-definition-completeness gate remaining there.

**Evidence corrected during verification.** The raw finding missed that `ayokoding-www-fe-e2e` already
carried this setting before the PR, with a detailed rationale (L20-52) for preferring it over a tag
filter. I checked whether that rationale transfers to these 5 configs. **It does not**, and the
reasoning in that very comment is what rules it out:

- ayokoding's blocker is ~104 pre-existing gap scenarios **already tagged `@unit @e2e`** — a tag
  filter genuinely cannot separate "declared e2e intent, not implemented" from "implemented".
- The new env-loader scenarios are tagged **plain `@unit`** (verified in each app's
  `gherkin/env-loader/env-loader.feature` — all four scenarios per file).
- These 5 configs ran default `fail-on-gen` **and passed CI before this PR**, which proves they have
  no equivalent pre-existing gap population. The new plain-`@unit` scenarios are the sole trigger.

A tag filter therefore fully solves the problem here, and this same diff already uses that pattern in
the F# backend e2e configs — `apps/beavernest-be-e2e/playwright.config.ts:16` uses
`tags: "not @unit"`; `ose-be-e2e` and `organiclever-be-e2e` L16 use
`tags: "not @unit and not @integration"`.

**Fix**: replace `missingSteps: "skip-scenario"` with `tags: "not @unit"` in these 5 configs, matching
the pattern already used elsewhere in this same PR, and keep `fail-on-gen` intact.

specs/apps/ose/behavior/be/gherkin/config/env-tier-loading.feature
```diff
@@ -0,0 +1,21 @@
+Feature: ose-be environment tier loading
+
+  As a platform operator
+  I want ose-be to load exactly one .env.<APP_ENV> tier file at startup
+  So that agent-restricted tiers (.env.stag, .env.prod) never need to be opened by an AI
+    agent, while process environment variables set by CI always take precedence over any
+    file value
+
+  @unit
+  Scenario Outline: ose-be loads exactly one tier file
```
wahidyankf, line 10: ### LOG-4 — F# backends' Gherkin holds a lower completeness bar than the TS apps' in the same PR

**Severity**: MEDIUM | **Confidence**: 82

For "the same five rules," the acceptance criteria are specified unevenly. Verified at the pinned SHA:

- All three F# feature files carry **exactly one** scenario, covering tier selection (rule 2) only —
  this file L10, `specs/apps/organiclever/behavior/organiclever-be/gherkin/env/env-tier-loader.feature:8`,
  and `specs/apps/beavernest/behavior/beavernest-be/gherkin/configuration/env-tier-loading.feature:10`.
- The six Next.js apps' `env-loader.feature` files each carry **four** scenarios — rules 1, 3, 4 plus
  the stray-file guard.

The F# unit tests do exercise rules 3 and 4, but they carry no `@covers` tag pointing at any scenario,
so that coverage is invisible to the traceability mechanism. `@covers` is this repo's actual mechanism
and is gated by `specs:behavior:coverage` (confirmed present on all three backend `project.json`
files) — the gate passes only because it verifies scenario-to-test, not test-to-scenario, so the
under-specification is silent.

**Fix**: add scenarios for rules 3 and 4 to the three F# `.feature` files and `@covers`-tag the
existing tests that already prove them. This is mostly tagging work; the test bodies already exist.

specs/apps/organiclever/ddd/bounded-contexts.yaml
```diff
@@ -147,6 +147,35 @@ contexts:
     gherkin: specs/apps/organiclever/behavior/organiclever-app-web/gherkin/routing
     relationships: []
 
+  - name: env-loader
```
wahidyankf, line 150: ### GOV-1 — New bounded contexts registered, but derivative index artifacts not updated

**Severity**: MEDIUM | **Confidence**: 85 (lowered from 90; see framing correction)
**Raised by**: pr-review-governance-maker

This PR registers three new bounded contexts with matching glossary files — `env-loader` (this line)
and `be-env` (L166) here, and `config` in `specs/apps/ose/ddd/bounded-contexts.yaml:141` — but does
not update the four derivative index artifacts. Verified: the PR touches `bounded-contexts.yaml` and
the glossary files for both apps, and **no** DDD index file.

Not updated:
- `specs/apps/organiclever/ddd/ubiquitous-language/README.md` — index table has no `env-loader` row
- `specs/apps/ose/ddd/ubiquitous-language/README.md` — index table has no `config` row
- `specs/apps/organiclever/ddd/bounded-context-map.md:10` — prose still says "nine bounded contexts"
- `specs/apps/ose/ddd/bounded-context-map.md:35-39` — Context Summaries table omits `config`

That this is an oversight rather than a scope decision is well supported: **10** `gherkin/README.md`
index files were correctly updated by this same feature, and each app's own authoring rule states
glossary updates "ride with the change that introduces them. No separate 'glossary catch-up' PRs."

**Framing corrected during verification.** The raw finding described these indexes as listing "the
original 8" and "the original 5" contexts, implying they were complete beforehand. They were not:
`organiclever`'s index already omits `landing`, `be-db`, `be-health`, `be-journal` and `messaging`;
`ose`'s already omits `db` and `messaging`. So this PR adds to a pre-existing backlog rather than
creating the first gap — which is why confidence dropped to 85. No CI gate enforces index
completeness, which is why none of this failed the build.

**Fix**: add the missing rows/context to all four files. Per the repo's Root Cause Orientation
principle ("proactively fix preexisting errors encountered during work"), reconciling the
already-missing rows while you are in these files is in scope and cheap.

apps/ose-www/src/env-loader.unit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    cleanupTmpDirs();
+  });
+
+  Scenario("ose-www builds against the staging tier", ({ Given, When, Then }) => {
+    let appDir = "";
+    let env: EnvRecord = {};
+
+    Given('only ".env.stag" exists in the app directory', () => {
+      appDir = makeTmpAppDir();
+      writeEnvFile(appDir, ".env.stag", "SHARED_VAR=stag-value\nSTAG_ONLY_VAR=stag-only\n");
```
wahidyankf, line 58: ### INT-1 — Rule 1 (`APP_ENV` unset defaults to `local`) is untested in 8 of 9 loaders

**Severity**: MEDIUM | **Confidence**: 93 (raised from 88)
**Raised by**: pr-review-integrity-maker

Every loader branches on exactly this condition, and almost none of them test it. Verified by reading
all nine test suites at the pinned SHA:

**Six TS suites always pass an explicit `APP_ENV`** — never omitted, never `undefined`. Every
`env = { APP_ENV: ... }` construction in `ayokoding-www`, `organiclever-www`, `ose-www` (this file),
`organiclever-app-web`, `ose-app-web` and `wahidyankf-www` supplies a concrete tier string.

**Two F# suites likewise**: `ose-be`'s and `beavernest-be`'s `withAppEnv` helper always sets a
concrete tier.

**One gets it right**: `apps/organiclever-be/tests/unit/Tests/EnvTierLoaderTests.fs:16-24` — the test
named "currentTier defaults to local when APP_ENV is unset" explicitly sets `APP_ENV` to `null` and
asserts `"local"`. That proves the gap is fixable, not untestable.

Why it stays invisible: the coverage gate is **line** coverage (at least 90%), not branch coverage.
The default is a single-line ternary (`apps/ose-www/src/env-loader.ts:44`) or a single match arm, so
the line reports 100% while only the truthy arm ever executes. Rule 1 — the documented default that
governs every developer's machine — ships unverified in 8 of 9 loaders.

**Fix**: add one test per module mirroring `organiclever-be`'s existing pattern. See also ARCH-1: the
reason this costs 9 edits instead of 1 is the duplication.

apps/ose-www/src/env-loader.ts
```diff
@@ -0,0 +1,114 @@
+/**
```
wahidyankf, line 1: ### ARCH-1 — The tier loader is hand-duplicated 9 times instead of shared via `libs/`

**Severity**: HIGH | **Confidence**: 85

The loader is reimplemented 9 times rather than shared. Verified at the pinned SHA:

**Six TS copies are byte-identical** — all six files hash to `449fc4612506bb531dfb7a60af96a50a`:
`apps/{ayokoding-www,organiclever-www,ose-www,ose-app-web,wahidyankf-www}/src/env-loader.ts` and
`apps/organiclever-app-web/src/contexts/env-loader/infrastructure/env-loader.ts`.

**Three F# copies are near-identical**, under `apps/{ose-be,organiclever-be,beavernest-be}`.

Each of the six TS apps also independently adds `"dotenv": "16.4.7"` as a new per-app dependency
(6 additions across `package.json` files, 6 more in the lockfile).

`AGENTS.md`'s Monorepo Architecture section names `libs/` as this repo's sharing mechanism, and TS lib
precedent already exists (`libs/web-ui`, `libs/web-ui-token`), as does F# (`libs/fsharp-crane-core`).
I grepped the full plan corpus (`README.md`, `brd.md`, `prd.md`, `tech-docs.md`, `delivery.md`,
`learnings.md`) for `duplicat`/`shared`/`libs/ts-env`/`extract`: **no discussion of choosing 9 copies
over one shared module**. The only `extract` hits concern extracting a constant or helper *within* a
file. Meanwhile `tech-docs.md` itself calls the loader "identical across every app," confirming this
is one design template manually reimplemented.

The plan's "libraries never load env files" rule does **not** forbid the fix: it bans a lib from
*autonomously* loading a tier file on import. Extracting the pure functions (`resolveTier`, parse,
apply-if-unset) into a module the app still calls explicitly is a different shape — and it is already
the shape in use, since `loadTierEnv({ appDir, env })` takes injected parameters.

**Why HIGH, and why this is not a style preference**: the duplicated logic is the security-relevant
stray-file guard this PR exists to add, there is no compiler or CI check for drift between the copies,
and **the drift has already materialised before merge** — see LOG-1, where `organiclever-be`'s search
order is reversed relative to both siblings while all three doc comments claim they are identical.
INT-1 is the same cost in the test dimension: one missing test becomes 8 missing tests.

**Fix**: extract to `libs/ts-env-loader` (TS) and `libs/fsharp-crane-core` or a new
`libs/fsharp-env-loader` (F#). If the duplication is genuinely intentional, record it as a numbered
design decision in `tech-docs.md` stating the rationale and the accepted drift risk — right now the
choice is undocumented, so no reviewer can tell it from an oversight.

wahidyankf (commented):
**Cycle**: 1 of 7
**Risk tier**: full
**Per-specialist raw findings**: architecture 1, logic 4, governance 1, security 4, integrity 1, performance 0, docs 2, instruction 0, types 0
**Security-sensitive-path override applied**: not recorded by scout; tier was already `full`
**Diff coverage**: full diff reviewed in one pass (pinned SHA `aaf9bf2fbeba875fae6fe82687f4cee1dde2b01d`)
**Prior-cycle human dismissals respected**: none this cycle (no prior threads exist on this PR)

## Verdict: BLOCKING

This review is submitted as `COMMENT` because `gh` authenticates as the PR author and GitHub rejects
`REQUEST_CHANGES` on one's own PR. Treat the three `CRITICAL` findings as blocking.

**13 raw findings in, 13 posted.** Nothing was dropped by the reasonableness filter — every raw
finding survived independent tool-verification against the pinned SHA. Four had their evidence
corrected or their severity re-scored during verification (detailed below), and one specialist claim
was **retracted** for lack of evidence.

| Severity | Count | Findings |
| --- | --- | --- |
| CRITICAL | 3 | SEC-1, SEC-2, DOC-1 |
| HIGH | 2 | SEC-3, ARCH-1 |
| MEDIUM | 8 | SEC-4, DOC-2, LOG-1, LOG-2, LOG-3, LOG-4, GOV-1, INT-1 |

### Headline

The PR's own stated purpose is to hard-block agent access to `.env.prod` / `.env.stag`. I reproduced
**five distinct working bypasses** of that guard by piping tool-call JSON through the shipped hook
script at this exact SHA. The guard is real and does block the obvious shapes, but its Bash branch is
structurally default-allow, and neither branch normalises the target path or its case. Separately, a
new how-to doc promises the human operator a guarantee the mechanism does not deliver.

A note on how these were verified: while assembling this review, the shipped hook **blocked two of my
own Bash commands** — proof the guard genuinely fires on the shapes it covers. The findings below are
about the shapes it does not cover.

### Coordinator verification notes (corrections to specialist evidence)

These are recorded because the raw finding text was partly inaccurate; the underlying findings still
stand on corrected evidence.

1. **SEC-4 downgraded HIGH to MEDIUM.** The raw finding asserted "per Anthropic's own tracked
   behavior, Read-deny rules apply to Grep/Glob only best-effort." I could not substantiate that: the
   current Claude Code IAM and settings docs make **no statement either way** about whether `Read()`
   rules govern `Grep`/`Glob`. That claim is **retracted**. What I did verify locally still supports
   the finding, so it stays — at MEDIUM, since exploitability is unproven rather than demonstrated.
2. **LOG-3 evidence corrected.** The raw finding said `missingSteps` was added to 5 configs. There are
   **6** configs carrying it; `ayokoding-www-fe-e2e` already had it before this PR, with a long
   documented rationale for preferring it over a tag filter. Also, `beavernest-be-e2e` uses
   `tags: "not @unit"`, not `"not @unit and not @integration"` as claimed. The finding survives
   because the ayokoding rationale provably does **not** transfer to the 5 new configs (see LOG-3).
3. **GOV-1 framing corrected.** The raw finding described the glossary indexes as listing "the
   original 8" / "the original 5" contexts, implying they were complete before this PR. They were
   already stale: `organiclever`'s index omits `landing`, `be-db`, `be-health`, `be-journal`,
   `messaging`; `ose`'s omits `db` and `messaging`. Confidence lowered 90 to 85 accordingly.
4. **Two near-miss findings deliberately kept separate, not merged.** ARCH-1 (9-way loader
   duplication) and INT-1 (rule-1 untested in 8 of 9 loaders) share a theme but are distinct defects
   at distinct anchors with distinct fixes, so the Deduplicate function does not collapse them. They
   are cross-referenced instead. Note that **LOG-1 is empirical proof that ARCH-1's predicted drift
   has already occurred** — that materially strengthens ARCH-1 and is why it holds at HIGH.

### Reproduction transcript (SEC-1 / SEC-2 / SEC-3 / SEC-4)

Run against `.claude/hooks/block-env-file-access.sh` at the pinned SHA, by piping PreToolUse JSON to
the script and treating empty output as ALLOW:

```
file-tool: apps/ose-www/.env.prod           DENY    (baseline: guard works)
file-tool: /tmp/notenv  (symlink)           ALLOW   <-- SEC-2
file-tool: apps/ose-www/.ENV.PROD           ALLOW   <-- SEC-3
bash: [covered verb] + restricted path      DENY    (baseline: guard works)
bash: python3 -c open(".env.prod","w")      ALLOW   <-- SEC-1
bash: rsync .env.prod /tmp/exfil/           ALLOW   <-- SEC-1
bash: awk 1 .env.prod                       ALLOW   <-- SEC-1
bash: xxd .env.prod                         ALLOW   <-- SEC-1
bash: grep . .env.prod                      ALLOW   <-- SEC-1
bash: [covered verb] .ENV.PROD              ALLOW   <-- SEC-3
bash: variable indirection via $F           ALLOW   <-- SEC-1
Grep tool payload (path field)              ALLOW   <-- SEC-4
Write tool payload (file_path)              DENY    (hook only; no settings.json backstop)
```

Filesystem case-insensitivity independently confirmed on this machine (APFS): a file written as
`lower.txt` resolves as `LOWER.TXT`.

### Scope

All 13 findings sit inside the PR's declared scope (the `guard-env-file-access` policy, the per-app
tier loader, and their governance/docs surface). No scope-creep asks were manufactured.

---

*Reviewed against pinned SHA `aaf9bf2fbeba875fae6fe82687f4cee1dde2b01d`. Confidence scores are
post-verification, not the specialists' raw scores.*

.claude/hooks/block-env-file-access.sh
wahidyankf: Fixed: inverted the Bash branch of `.claude/hooks/block-env-file-access.sh` from default-allow to default-deny (commit `ccf57e916`). The branch now denies any Bash command whose raw text references `.env.prod`/`.env.stag` anywhere — bare argument, quoted string, variable assignment (`F=.env.prod; cat "$F"`), tool argument — via a single `grep -qiE` presence check, before any allowlist logic runs. Only a narrow, explicitly-safe carve-out survives: read-only `git check-ignore|ls-files|status` queries that reveal ignore/tracking status, never file content. All five originally-reproduced bypasses (`python3 -c open(...)`, `rsync`, `awk`, `xxd`, variable indirection) are now denied — added as regression tests in `.claude/hooks/block-env-file-access.test.sh` (52 tests total, all passing). The one residual gap this cannot close in principle — a tier name assembled from separately-innocuous pieces (`t=prod; cat ".env.$t"`) rather than spelled out literally — is documented as an accepted, deliberate limitation in `repo-governance/conventions/security/secrets-and-env-standards.md` §9, matching the fix note.

.claude/hooks/block-env-file-access.sh
wahidyankf: Fixed: the file-tool branch of `.claude/hooks/block-env-file-access.sh` now resolves the path via `realpath -q --` (falling back to `readlink -f --`, then the original string if neither is available or the path doesn't exist yet — e.g. a `Write` creating a new file) and re-checks the resolved basename against the restricted-tier pattern, in addition to the original literal-basename check (commit `ccf57e916`). A symlink whose own name doesn't look like `.env.prod`/`.env.stag` but whose target does is now denied. Added a regression test in `.claude/hooks/block-env-file-access.test.sh` that creates a real symlink (`notenv -> .env.prod`) in a temp dir and asserts `Read` is denied — confirmed failing (ALLOW) before the fix and passing (DENY) after, matching your reproduction. As you noted, this is a separate fix from SEC-1 — both are in the same commit since `ln` creating the symlink is itself gated by SEC-1's default-deny once the command references the restricted tier name, but a symlink planted by another actor still needed this independent path-resolution fix.

.claude/hooks/block-env-file-access.sh
wahidyankf: Fixed: added `-i` (or the `grep -qiE`/`grep -qE ... -i` equivalent) to every tier-matching `grep` call in both the file-tool branch (L20-equivalent) and the Bash branch (L53-equivalent) of `.claude/hooks/block-env-file-access.sh` (commit `ccf57e916`) — matching is now case-insensitive throughout, closing both branches as your fix note required. Added regression tests: `Bash cat .ENV.PROD`, `Read apps/ose-www/.ENV.PROD`, and `Bash CAT .Env.Stag` (mixed case) all now DENY. Since this repo's own git checkout is on APFS (case-insensitive, per your reproduction), this closes a real bypass rather than a theoretical one.

.claude/settings.json
wahidyankf: Fixed, per your suggested fix exactly (commit `ccf57e916`):

- `.claude/settings.json`'s `PreToolUse` matcher widened from `Read|Write|Edit|MultiEdit` to `Read|Write|Edit|MultiEdit|Grep|Glob`.
- `.claude/hooks/block-env-file-access.sh`'s file-tool branch now reads `.tool_input.file_path // .tool_input.path // empty`, so `Grep`/`Glob` payloads (which carry the target under `path`) are extracted correctly instead of yielding empty and auto-allowing.
- `permissions.deny` extended with `Write(**/.env.prod)` and `Write(**/.env.stag)`, closing the adjacent Write-has-no-backstop gap you flagged.

Added regression tests: `Grep pattern with path=.env.prod` and `Glob path=.env.stag` now DENY; `Grep pattern with path=.env.local` (not a restricted tier) still ALLOWs, confirming the new `path` extraction doesn't over-deny.

docs/how-to/configure-app-environments.md
wahidyankf: Fixed (commit `fc6b9969e`): reworded this paragraph to match the policy-only framing used everywhere else in the repo. It now states the restriction is enforced on a **best-effort basis** by the `PreToolUse` hook, names the hook file, explains it denies recognized tool calls/command shapes but "cannot see every way a shell command could reference a file," and replaces "Only you... can create or edit those two files" with guidance that the human operator is the *primary and most reliable* way to touch those files and should apply compensating controls (secrets-manager storage, access logging, least-privilege credentials) rather than relying on the hook alone. Landed in the same PR as the SEC-1/2/3 fixes, per your instruction.

docs/how-to/configure-app-environments.md
wahidyankf: Fixed (commit `fc6b9969e`): replaced the bare `Infrastructure.loadEnvTier ()` snippet with a version matching the actual `apps/ose-be/src/OseBe/Program.fs` shape — `open OseBe.Contexts.Config.Infrastructure` followed by a bare `loadEnvTier ()` call inside `main`, which is exactly how the real file calls it (verified against `Program.fs:11,39`). The snippet now compiles as written.

apps/organiclever-be/src/OrganicleverBe/Contexts/Env/Infrastructure/EnvTier.fs
wahidyankf: Fixed (commit `ec8680d7c`): swapped `candidateDirs ()` to `[ "apps/organiclever-be"; "." ]`, matching both `ose-be` and `beavernest-be`'s app-dir-first order exactly as you specified. `organiclever-be:test:unit` (32/32) and `organiclever-be:specs:behavior:coverage` both pass after the change.

apps/ose-be/src/OseBe/Contexts/Config/Infrastructure/EnvTier.fs
wahidyankf: Fixed (commit `ec8680d7c`), per your recommended direction (align F# to null-only, the smaller change): all three F# loaders (`ose-be`, `organiclever-be`, `beavernest-be`) now check presence via a null-only match (`isNull current` / `| null -> ...`) instead of `String.IsNullOrEmpty`/`| null | "" -> ...`. An explicit empty-string process env var now counts as "set" and is left untouched in all three F# loaders, matching the six TS loaders' `dotenv({ override: false })` `hasOwnProperty`-based presence semantics exactly. Added a doc-comment cross-reference to the TS behavior at each of the three call sites so future readers don't reintroduce the divergence. All three backends' `test:unit` suites pass (14/83/32).

I did not add a dedicated pinning test for the empty-string case specifically (your fix note flagged "no Gherkin scenario or unit test in this PR pins either behaviour" as part of the problem) — happy to add one in a follow-up cycle if you'd like it explicit rather than implied by the code comment cross-references.

apps/organiclever-www-fe-e2e/playwright.config.ts
wahidyankf: Fixed, with a correction along the way (commits `de1df76de` then `c48798b4b`). Applied your suggested `tags: "not @unit"` fix to all 5 configs first, verified only via `npx bddgen` exit-code success — which turned out to be insufficient: it detects a scenario present-but-unbound, not a scenario silently *excluded from generation entirely*. The pre-push `specs:e2e:coverage` gate caught what that check missed: 3 of the 5 apps (`organiclever-www-fe-e2e`, `organiclever-app-web-e2e`, `wahidyankf-www-fe-e2e`) tag many already-**implemented** e2e scenarios `@unit @e2e` — not just their env-loader scenarios' plain `@unit` — so `tags: "not @unit"` silently dropped those real, already-bound scenarios from generation too (31 newly-unbound in `wahidyankf-www-fe-e2e` alone; 59 co-tagged lines in `organiclever-app-web`, 13 in `organiclever-www`). That's a worse regression than the original finding, so I reverted those 3 back to `missingSteps: "skip-scenario"` with a comment explaining why (same shape as `ayokoding-www-fe-e2e`'s existing rationale — the gap population isn't cleanly separable by tag there either).

Kept `tags: "not @unit"` for `beavernest-app-web-e2e` and `ose-app-web-e2e`, where it's verified safe — `@unit` appears nowhere else in either app's Gherkin surface. `specs:e2e:coverage` now passes clean (0 new unbound scenarios) for all 5 configs.

specs/apps/ose/behavior/be/gherkin/config/env-tier-loading.feature
wahidyankf: Fixed (commit `c97d80086`): added the two missing scenarios (rule 3 "process env wins over a tier file value", rule 4 "tolerates a missing tier file") to all three F# `.feature` files, bringing each to 3 scenarios (tier selection, process-env-wins, missing-file-tolerance) — the F#-applicable subset of the six TS apps' 4 (TS additionally has the stray-auto-loaded-file guard scenario, which doesn't apply to F#/.NET since there's no Next.js-style auto-loading to guard against). `@covers`-tagged the two already-existing xunit tests you identified (`loadEnvTierFrom(Dir) never overrides...` / `...does not override an already-set process variable`, and `...does nothing when the tier file is absent` / `...is a no-op when the tier file is absent`) to the new scenarios, and added matching no-op TickSpec step bindings following `organiclever-be`'s existing documented convention (the step-coverage validator needs a text match; the real behavior stays exercised directly in the xunit tests, as your fix note anticipated — "mostly tagging work; the test bodies already exist"). `specs:behavior:coverage` passes clean for all three backends (`ose-be`: 9 scenarios/31 steps, `organiclever-be`: 14 scenarios/43 steps, `beavernest-be`: 21 scenarios/93 steps — all covered).

specs/apps/organiclever/ddd/bounded-contexts.yaml
wahidyankf: Partially fixed, rest deferred (commit `017098cc1`; leaving this thread unresolved since it's not fully closed).

Fixed: added index rows for this PR's own three new bounded contexts — `env-loader` and `be-env` to `specs/apps/organiclever/ddd/ubiquitous-language/README.md`, `config` to `specs/apps/ose/ddd/ubiquitous-language/README.md` — directly closing the specific authoring-rule violation this finding centers on ("glossary updates ride with the change that introduces them").

Deferred: reconciling `bounded-context-map.md`'s pre-existing backlog — `specs/apps/organiclever/ddd/bounded-context-map.md` has **no section at all** for `organiclever-be`'s 5 backend contexts (`be-db`, `be-health`, `be-journal`, `messaging`, now `be-env`), and `specs/apps/ose/ddd/bounded-context-map.md`'s Context Summaries table omits `db`, `messaging`, and now `config`. This predates this PR (per your own framing correction) and reconciling it properly means authoring persistence/owns/depends-on/strategic-relationship descriptions (and, matching the organiclever map's existing style, a Mermaid diagram) for backend contexts I haven't independently verified against their actual implementations — doing that hastily risks landing inaccurate architecture documentation, which is worse than the current known gap. Filed as wahidyankf/ose-public#178, scoped to exactly this reconciliation.

apps/ose-www/src/env-loader.unit.test.ts
wahidyankf: Fixed for all 9 loaders (commits `c97d80086`, `b18838753`).

F# — `ose-be` and `beavernest-be` each got a new xunit `[<Fact>]` (`loadEnvTier defaults to the local tier when APP_ENV is unset`) exercised through the public `loadEnvTier()` via an isolated CWD/temp-dir fixture, matching `organiclever-be`'s existing pattern you cited as "proves the gap is fixable." `organiclever-be` already had this test, so it needed no change.

TS — added a plain vitest `describe("resolveTier", ...)` block per app (all 6: `ayokoding-www`, `organiclever-www`, `ose-www`, `organiclever-app-web`, `ose-app-web`, `wahidyankf-www`) exercising the exported `resolveTier` function directly with both an unset and an empty-string `APP_ENV`, rather than routing through the Cucumber/Gherkin binding machinery — this mirrors the F# pattern (a plain unit test, not a new Gherkin scenario) since no scenario in any of these apps' `env-loader.feature` files currently pins this default either.

All 9 loaders' `test:unit` suites pass with the new tests included; typecheck and lint clean on all 6 TS apps.

apps/ose-www/src/env-loader.ts
wahidyankf: Deferred, not fixed in this cycle. Agreed this is a real, well-evidenced finding — LOG-1's already-materialized drift (fixed in this same pass) is exactly the concrete proof you cited. Extracting 6 byte-identical TS files into `libs/ts-env-loader` plus reworking 6 `package.json`s/the lockfile, and consolidating 3 near-identical F# files into `libs/fsharp-crane-core` or a new `libs/fsharp-env-loader`, is a genuine cross-cutting refactor touching every app in this PR's diff a second time — disproportionate scope for a PR whose purpose is closing the security bypasses in the access-restriction hook (SEC-1/2/3/4, DOC-1, all fixed in this cycle) rather than restructuring the loader's delivery shape. Filed as wahidyankf/ose-public#177 with the extraction plan and the "record it as a numbered design decision instead" alternative you offered, so it's tracked as a scoped follow-up rather than silently dropped. Leaving this thread unresolved per the fixer's defer discipline — the finding stays valid and re-verifiable in the next cycle.

wahidyankf: Fixed: extracted the tier-loader logic into new shared libs — libs/ts-env-loader (consumed by all 6 TS apps via a thin per-app wrapper) and libs/fsharp-env-loader (consumed by ose-be, organiclever-be, beavernest-be via ProjectReference). All 9 apps now delegate to a single canonical implementation instead of 9 hand-duplicated copies; each app wrapper only supplies its own composition-root search directories. Verified via full test:quick sweep (typecheck/lint/unit/coverage/specs) for all 9 apps + both new libs, plus a clean-NuGet-cache rebuild to catch a downgrade issue the extraction introduced (also fixed). Commits ac0ffffb2, b45ad0896, 000f25155.

specs/apps/organiclever/ddd/bounded-contexts.yaml
wahidyankf: Explicitly authorized exception recorded: the user reviewed this fixer's framing (this PR only adds to a pre-existing DDD-index backlog rather than creating it, and the specific new-context violation this PR introduced is already fixed) and authorized merging with the remaining pre-existing-backlog reconciliation tracked in #178 rather than rushed in this PR. Per pr-merge-protocol.md's "unless a separate, explicitly authorized exception says otherwise" carve-out on precondition (b). Leaving this thread unresolved/open since the finding itself is not fixed — only merge-blocking status is waived — so #178 stays discoverable.

— generated by AI

PR #197, opened 2026-08-14
feat(governance): arm word-budget and readme-completeness gates (Phase 9)

wahidyankf:
## Summary
- Registers `governance-word-budget` and `governance-readme-completeness` on `pre-push`+`ci` (path-gated, no `pre-commit`), and arms FR-4 (frontmatter `when_to_use`/`description` WARN→FAIL)
- Clears the pre-existing debt both gates would otherwise fail on:
  - 436 readme-completeness findings (430 under `.claude/skills/*/README.md` and `reference/README.md` indexes, 6 under `repo-governance/README.md`) — hand-authored `— <description>` annotations since target files carry no frontmatter for `readme-index generate` to derive from
  - 13 word-budget FAIL findings: two `.opencode/agents/` mirrors trimmed at their `.claude/agents/` source, `apps/rhino-cli/README.md` cut from 1041 to 884 words, and the Nx-vendored `.opencode/skills/{monitor-ci,nx-generate,nx-import,nx-plugins,nx-run-tasks,nx-workspace,link-workspace-packages}/` + `.opencode/commands/monitor-ci.md` tree (commit `4239f3d79`, no `.claude/` source of truth) excluded from scope — same class as the existing `.fvm/` exclude, not a per-file waiver
- Proves path-gating fires/skips correctly on both surfaces via an isolated fixture (`evidence/phase-9-trigger-matrix.txt`) and confirms the full pre-push gate run passes end-to-end against the real, fully remediated tree

## Test plan
- [x] `rhino-cli gate validate` exits 0
- [x] `rhino-cli md frontmatter validate` — 0 findings against the real repo (FR-4 armed)
- [x] `rhino-cli governance word-budget validate` (with registered excludes) — 0 FAIL findings
- [x] `rhino-cli governance readme-index validate` (readme-completeness scope) — PASSED
- [x] 5-case trigger matrix proven via isolated fixture on both `pre-push` and `ci`
- [x] `rhino-cli parity manifest validate` — current
- [x] Full `gate run --surface=pre-push` against the real, fully-modified tree — exit 0


wahidyankf (commented):
## Consolidated PR Review — Cycle 1

**Cycle**: 1 of 7
**Risk tier**: full
**Per-specialist raw findings**: architecture 4, logic 1, governance 2 (both routed out-of-charter, 0 in-charter), security 1, integrity 0, performance 0, docs 3, instruction 1, types 0
**Security-sensitive-path override applied**: no (full tier already selects all nine)
**Diff coverage**: full diff reviewed in one pass (131 files, +1103/-617)
**Prior-cycle human dismissals respected**: none this cycle (first cycle, no prior threads)

---

### THIS REVIEW IS BLOCKING

GitHub shows STATE `COMMENT` because `gh` authenticates as the PR author and GitHub rejects `REQUEST_CHANGES` on one's own PR. Blocking status lives in the **severity label of each finding**, never in the review STATE — see [GitHub Reviews API Mechanics Part 1](https://github.com/wahidyankf/ose-public/blob/752f0e3952404305ffae96113502e765c687b5d1/repo-governance/workflows/pr/pr-review-quality-gate/08-github-reviews-api-mechanics-part-1.md). **2 CRITICAL + 5 HIGH + 3 MEDIUM** findings are outstanding; per the [route-specific done-definition](https://github.com/wahidyankf/ose-public/blob/752f0e3952404305ffae96113502e765c687b5d1/repo-governance/workflows/pr/pr-review-quality-gate/10-route-specific-done-definition.md), this PR is **not done**.

### Findings

| # | Severity | Conf | Discipline | Anchor |
| 1 | CRITICAL | 98 | Correctness | `audit_orchestrator.rs:267` |
| 2 | CRITICAL | 97 | Correctness | `frontmatter.rs:316` |
| 3 | HIGH | 92 | Architecture | `repo-config.yml:1056` |
| 4 | HIGH | 92 | Instruction-decay | `repo-config.yml:1102` |
| 5 | HIGH | 90 | Documentation | `repo-config.yml:1108` |
| 6 | HIGH | 90 | Documentation | `repo-config.yml:1047` |
| 7 | HIGH | 88 | Security | `evidence/phase-9-trigger-matrix.txt:21` |
| 8 | MEDIUM | 95 | Documentation | `repo-config.yml:1044` |
| 9 | MEDIUM | 92 | Documentation | `repo-governance/README.md:25` |
| 10 | MEDIUM | 85 | Architecture | `word_budget.rs:317` |

### Coordinator notes (what the four functions did to the raw findings)

- **Deduplicated**: `pr-review-architecture-maker`'s findings 1, 2 and 3 all name the same `repo-config.yml` `args.exclude` block from three angles (FR-1.4/FR-1.5 contradiction, `.fvm/` risk-class misclassification, enumerated-directory brittleness). Collapsed into **one** thread (#3) with three sub-points rather than three threads on one config block.
- **Re-categorized**: `pr-review-governance-maker`'s routing item B (the `tests/docs.rs` regression) was routed out of governance as "logic/integrity territory". Coordinator ruling: the defect is that shipped code now contradicts a **declared Gherkin acceptance criterion** — that is [business-logic/correctness](https://github.com/wahidyankf/ose-public/blob/752f0e3952404305ffae96113502e765c687b5d1/repo-governance/development/quality/pr-review-disciplines/04-the-nine-reviewer-disciplines-table-part-1.md)'s owned scope ("Gherkin acceptance-criteria conformance"), not integrity's (which owns *weakened/skipped* tests, not a genuine behavioral regression). Filed as **correctness** (#2), with the CI-orphan mechanism recorded as the reason it went undetected. Governance's routing item A (`governance-word-budget-remediation.md`'s false pre-commit claim) was **re-categorized to documentation** and **kept in scope** (#8) — it is out-of-diff but squarely in the PR's declared plan scope (the doc was created by this same plan's Phase 1, commit `6b8cac957`, PR #187) and this PR's own FR-1.14 no-pre-commit decision is what makes the false claim consequential. Anchored to a line that *is* in the diff.
- **Reasonableness-filtered**: no finding dropped. Two severities **lowered** on coordinator review: docs findings 1 and 2 came in at CRITICAL and were reduced to HIGH — both are stale-spec/latent-trap defects that mislead a future executor, but neither breaks functionality or blocks a build today, which is what [CRITICAL](https://github.com/wahidyankf/ose-public/blob/752f0e3952404305ffae96113502e765c687b5d1/repo-governance/development/quality/criticality-levels/02-four-universal-criticality-levels-critical.md) requires. Both remain blocking. One severity **raised**: docs finding 3 held at MEDIUM but confidence raised 85 to 92 after the coordinator read the actual diff hunk (the specialist cited line 73; the defect is at lines 24-28 and 55, and is worse than "run-on" — it is verbatim self-duplication).
- **Tool-verified**: findings #1 and #2 were reproduced end-to-end by the coordinator by building the pinned-SHA binary and running it, satisfying the [CRITICAL-requires-reproduction rule](https://github.com/wahidyankf/ose-public/blob/752f0e3952404305ffae96113502e765c687b5d1/repo-governance/development/quality/pr-review-disciplines/14-quality-gate-enhancements-critical-reproduction-and-seven-cycle-maximum.md). Findings #3, #5, #6, #7, #8, #9 and #10 were each independently confirmed against the repo at this SHA. One specialist claim was **corrected**: `pr-review-security-maker` cited "4 required isolation layers"; the convention mandates **six** — see #7.

No prompt-injection attempt was detected in the PR body, comments or linked-issue text.

---

apps/rhino-cli/src/application/repo_governance/audit_orchestrator.rs
```diff
@@ -264,7 +264,7 @@ fn audit_word_budget(repo_root: &Path) -> Vec<AuditFinding> {
     let Some(config) = word_budget::merged_budget_config(repo_root) else {
         return Vec::new();
     };
-    let mut findings = word_budget::check_instruction_sizes(&RealFs, repo_root, &config);
+    let mut findings = word_budget::check_instruction_sizes(&RealFs, repo_root, &config, &[]);
```
wahidyankf: ### Finding 1 — `excludes` is wired into only 1 of 3 call sites

**Severity**: CRITICAL | **Confidence**: 98 | **Discipline**: Business-logic / correctness

The new `excludes: &[String]` parameter on `check_instruction_sizes` is threaded through **one** of its three call sites. This line hardcodes `&[]`, so `repo-governance audit` ignores every `args.exclude` entry the gate registration declares at `repo-config.yml:1058-1078`.

**Reproduced by the coordinator at `752f0e3`** — built `apps/rhino-cli/target/release/rhino-cli` from this commit, run from the repo root:

| Invocation | Exit | Result |
| --- | --- | --- |
| `rhino-cli governance word-budget validate` | **1** | 155 FAIL findings; 22 output lines name vendored `.fvm/` Flutter-SDK cache files |
| `rhino-cli repo-governance audit --include-category governance-word-budget` | **1** | `Error: 155 governance finding(s) reported across 1 categor(ies)` |
| the same command **plus the 13 `--exclude` flags** from the gate registration | **0** | passes |

Only the gate-registration path (pre-push/ci, where `fixed_arguments()` materialises the flags) is actually fixed. Two surfaces remain broken:

1. **The bare CLI command.** Every FR-1 Gherkin scenario is written against the bare form. [`governance-word-budget.feature:75`](https://github.com/wahidyankf/ose-public/blob/752f0e3952404305ffae96113502e765c687b5d1/specs/apps/rhino/behavior/rhino-cli/gherkin/governance/governance-word-budget.feature#L75) — *"An out-of-scope file is never scanned"* — states `When the developer runs governance word-budget validate` / `Then the command exits successfully`. Against this repository that contract now fails.
2. **The `repo-governance audit` category** (this line), required by `prd.md` FR-5.7 and asserted as Enforcement Point 4 of `repo-governance/conventions/structure/governance-word-budget-remediation.md`. It is the deterministic-preflight surface `repo-rules-checker` consumes, so `repo-rules-checker` now consumes 155 false positives.

**Suggested fix**: give `audit_word_budget()` the same exclude list the gate registration passes — either by reading the `gates:` entry's `args.exclude` from `repo-config.yml`, or (cleaner, and it fixes the bare CLI in the same stroke) by hoisting the exclusion list into the `governance-word-budget:` config block so all three call sites read one source. Passing `&[]` here is not a defensible default while a blanket `**/README.md` surface is declared.

apps/rhino-cli/src/application/docs/frontmatter.rs
```diff
@@ -303,24 +303,21 @@ fn validate_governance_schema(path: &str, fm: &serde_norway::Value) -> Vec<DocsF
...
+    // now FAIL-severity, per plans/in-progress/optimize-governance-md `tech-docs.md`
+    // §5 "Dark-launch sequencing" — was WARN-severity from Phase 1 through Phase 8.
     if !has_non_empty_string(fm, "when_to_use") {
-        findings.push(DocsFrontmatterFinding {
-            file: path.to_string(),
-            severity: SEVERITY_WARN.to_string(),
-            kind: KIND_MISSING_WHEN_TO_USE.to_string(),
-            message: "recommended field \"when_to_use\" is missing or empty".to_string(),
-        });
+        findings.push(mk_fail(
```
wahidyankf, line 316: ### Finding 2 — FR-4 arming regresses a shipped Gherkin scenario, invisibly to CI

**Severity**: CRITICAL | **Confidence**: 97 | **Discipline**: Business-logic / correctness
**Raised by**: pr-review-governance-maker *(routed as out-of-charter; re-categorized to correctness by the coordinator — the defect is Gherkin acceptance-criteria non-conformance, which is `pr-review-logic-maker`'s owned scope)*

This hunk arms `when_to_use` (and, at line 306, `description`) from WARN to FAIL in `validate_governance_schema`. The unit tests in this same file were updated to match — but the **Gherkin acceptance criterion was not**, and it now fails.

**Reproduced by the coordinator at `752f0e3`**: `cargo test --test docs` → **1 failed**.

```
Scenario: Governance doc with only title passes the lighter schema
  ✔  Given a governance doc carrying only a title frontmatter field
  ✔  When the developer runs docs validate-frontmatter
  ✘  Then the command exits successfully
     Defined: specs/apps/rhino/behavior/rhino-cli/gherkin/md/docs-validate-frontmatter.feature:37:5
     Step panicked: DOCS FRONTMATTER VALIDATION FAILED: 2 fail finding(s)
       .../repo-governance/conventions/foo.md  [fail]  missing-description
       .../repo-governance/conventions/foo.md  [fail]  missing-when-to-use
     left: 1   right: 0
```

Neither `docs-validate-frontmatter.feature` nor `tests/docs.rs` appears in this PR's diff — the spec still asserts the pre-arming contract.

**Why CI is green anyway.** `tests/docs.rs` is orphaned from every CI-run target:

- `rhino-cli:test:unit` runs `--lib` plus **six named** integration tests (`repo_governance`, `env_contract`, `repo_config_data_driven`, `fsharp_tool_invocation`, `gate_specs`, `gate_dispatch`) — `docs` is not among them.
- `rhino-cli:test:integration` *would* catch it (`cargo test --tests`), but `.github/workflows/pr-quality-gate.yml` never invokes `test:integration` for Rust — the Rust job runs `test:quick` and `test:coverage` only, and `test:coverage` is `--lib`.

So a genuine regression in a shipped acceptance criterion lands behind a green gate.

**Suggested fix**: decide which side is authoritative and make them agree — either update the `Governance doc with only title passes the lighter schema` scenario to the FR-4-armed contract (and rename it), or narrow the arming. Then close the CI hole, otherwise the next spec regression is equally invisible: add `--test docs` to `test:unit`'s named list (with the corresponding `inputs:` entry), or wire `test:integration` into the Rust CI job.

repo-config.yml
```diff
@@ -1039,3 +1039,87 @@ gates:
...
+  # args.exclude: the **/README.md surface (declared in governance-word-budget: above)
+  # has no scoping relationship to the other covered-surface globs — it matches every
+  # README.md in the repo. These prefixes keep it from reaching trees FR-1.3 never
+  # declared as covered surfaces (plans/, docs/, specs/), a local, gitignored Flutter
+  # SDK version-manager cache (.fvm/, .fvm-cache/) that only exists on disk for
+  # developers who run `fvm` locally, and the Nx-vendored `.opencode/skills/`+
+  # `.opencode/commands/monitor-ci.md` content added wholesale by commit 4239f3d79
+  # ("Nx-generated AI agent configs") — it has no `.claude/` source of truth and isn't
+  # hand-authored agent context this repo's governance system produces. Not a per-file
+  # waiver on an in-scope surface (FR-1.5 still forbids that); see
```
wahidyankf: ### Finding 3 — `args.exclude` is the exemption mechanism FR-1.5 forbids, one layer down

**Severity**: HIGH | **Confidence**: 92 | **Discipline**: Architecture
*(Coordinator dedup: this thread consolidates three raw architecture findings that all named this same `args.exclude` block — the FR-1.4/FR-1.5 contradiction, the `.fvm/` risk-class misclassification, and the enumerated-directory brittleness. One block, one thread.)*

This comment asserts *"Not a per-file waiver on an in-scope surface (FR-1.5 still forbids that)"*. The list immediately below it contains `.opencode/commands/monitor-ci.md` — **an exact single-file entry**, on a surface (`.opencode/**/*.md`, `repo-config.yml:232`) that FR-1.3 explicitly declares in-scope. That file is **2056 words** against a 500-word fail budget: without this entry it is a FAIL finding. The comment's own disclaimer is false for the entry it sits above.

**(a) FR-1.4 / FR-1.5 contradiction.** `prd.md` FR-1.4 states generated mirrors are *"**gated**, not exempted"*; FR-1.5 states there is *"**no** exemption list, allowlist, waiver key, or per-file override — the config schema must not admit one"*, enforced by `BudgetConfig`'s `deny_unknown_fields` and asserted by the Gherkin scenario *"The config schema rejects an exemption key"*. `args.exclude` achieves at the **gate-registration** layer exactly what that schema guard blocks at the **config-block** layer. The plan's own `README.md` §Scope Boundaries still says the word budget covers `.opencode/**/*.md` *"**including generated mirrors**"*. Whether an exemption may live one layer down is a genuinely new structural call — make it explicitly and write it into `prd.md`, rather than leaving FR-1.4/FR-1.5 standing as-written and contradicted in config.

The stated rationale for the `.opencode/skills/` entries is factually correct and defensible — the coordinator confirmed these trees have **no `.claude/` source of truth** (`ls .claude/skills/` shows no `nx-*`, `monitor-ci` or `link-workspace-packages`). The problem is that the decision was recorded only in a YAML comment, not in the requirement it contradicts.

**(b) The risk-class analogy in the decision record is wrong.** The comment (and `delivery.md`) groups `.opencode/skills/` with `.fvm/`. They are not the same class: `.fvm/` is gitignored and *"only exists on disk for developers who run `fvm` locally"* — it is absent from CI entirely. `.opencode/skills/` is **git-tracked and present in every clone and every CI run**. Excluding the first removes a local-only nuisance; excluding the second permanently removes tracked repository content from a gate. Justify it on its own terms.

**(c) The exclusion will silently rot.** Seven skill directories are enumerated one-by-one. The next Nx regeneration that adds an eighth is silently *not* excluded — it will simply fail the gate with no signal that the exclusion list was the thing that went stale. Prefer a tree-level exclusion (`.opencode/skills/`), or a marker file, or the existing `SKIP_DIRS` mechanism in `word_budget.rs`, over an enumeration that no mechanism keeps in sync.

repo-config.yml
```diff
@@ -1039,3 +1039,87 @@ gates:
...
+
+  # New gate id (FR-3.20) — shares the governance readme-index binary with
+  # governance-readme-index above via --paths/--fail-kinds (FR-5.8). Scoped to just the
+  # trees Phases 1-8 of plans/in-progress/optimize-governance-md actually brought to
+  # compliance — repo-governance/, .claude/, .codex/, .pi/ — narrower than FR-1.11's
+  # originally-declared 6-entry list (which also named docs/, specs/): those two trees
+  # were never remediated by any phase of this plan and carry ~880 pre-existing
+  # missing/unannotated findings; a follow-up plan tracks bringing them to the same bar.
+  # governance-readme-index's own orphan/ghost enforcement (FR-3.19) is untouched.
+  - id: governance-readme-completeness
```
wahidyankf, line 1104: ### Finding 4 — a newly armed push-blocking gate with zero mention in AGENTS.md or CLAUDE.md

**Severity**: HIGH | **Confidence**: 92 | **Discipline**: Instruction-decay
**Raised by**: pr-review-instruction-maker

`governance-readme-completeness` is armed here on **both** `pre-push` and `ci`. Coordinator-verified: `git show <merge-base>:repo-config.yml` contains neither `id: governance-readme-completeness` nor `id: governance-word-budget` — both gates are newly armed by this PR, and this one will start blocking `git push` for anyone touching `repo-governance/`, `.claude/`, `.codex/`, `.pi/` or `repo-config.yml`.

Neither `AGENTS.md` nor `CLAUDE.md` mentions it — and **both files are rewritten by this very PR**, so the omission is not a pre-existing gap that fell out of scope.

The contrast with its sibling is the argument: `AGENTS.md:35-37` names the word budget under §Quality Gates and links a convention doc:

> Instruction files carry a word budget; remediation is progressive disclosure.
> **See**: [governance-word-budget.md](./repo-governance/conventions/structure/governance-word-budget.md)

There is no equivalent line for README-completeness, and no `repo-governance/conventions/structure/governance-readme-completeness*.md` exists at all (coordinator-verified against the `structure/` listing). An agent's first encounter with this gate will be a blocked push, with the `missing`/`unannotated` finding kinds and the 4-entry scan scope undocumented anywhere it would think to look.

**Suggested fix**: add a §Quality Gates sentence naming the gate and its two finding kinds, plus a convention doc it can link to — matching the pattern the word-budget gate already sets.

repo-config.yml
```diff
@@ -1039,3 +1039,87 @@ gates:
...
+    type: check
+    command: governance readme-index validate
+    kind: rhino-cli
+    ci-group: governance
+    args:
+      paths:
```
wahidyankf, line 1110: ### Finding 5 — shipped scan scope is 4 entries; every canonical spec doc still says 6

**Severity**: HIGH | **Confidence**: 90 | **Discipline**: Documentation-quality
*(Coordinator note: raised as CRITICAL; lowered to HIGH — this misleads a future executor but breaks no build today. Still blocking.)*

This `paths:` block ships **four** entries: `repo-governance/`, `.claude/`, `.codex/`, `.pi/`. The plan's canonical spec docs still describe the **six**-entry list, adding `docs/` and `specs/`:

- `plans/in-progress/optimize-governance-md/README.md` §Scope Boundaries: *"the wider FR-3.7 list — `repo-governance/`, `.claude/`, `.codex/`, `.pi/`, `docs/`, `specs/`"*
- `prd.md` FR-5.8 / FR-3.20: *"Set explicitly on `governance-readme-completeness` to FR-3.7's **widened 6-entry list**"*
- `tech-docs.md` §4 carries the same widened list.

`delivery.md` documents the narrowing; the three documents an executor actually reads for scope do not. **Phases 10 and 16 mirror this into `ose-private`** — an executor working from `README.md`/`prd.md`/`tech-docs.md` rather than from a buried config comment will mirror the 6-entry list, and the two repos will silently diverge on which trees a push-blocking gate scans.

`prd.md` FR-5.8 is phrased as a requirement the config must satisfy, and the config does not satisfy it. Either amend FR-3.7/FR-5.8/§Scope Boundaries to the narrowed 4-entry reality (with the rationale `delivery.md` already records), or widen this block back to 6. Leaving the requirement and the implementation contradicting each other is the one option that guarantees the `ose-private` mirror goes wrong.

repo-config.yml
```diff
@@ -1039,3 +1039,87 @@ gates:
           - docs/reference/platform-bindings.md
           - repo-governance/
       ci: { scope: other }
+
+  # Armed at Phase 9 of plans/in-progress/optimize-governance-md, dark-launched at Phase 1
+  # (config-only, see governance-word-budget: above). No pre-commit surface (FR-1.14): a
+  # whole-tree scan on every commit buys no additional coverage over pre-push/ci.
+  #
+  # args.exclude: the **/README.md surface (declared in governance-word-budget: above)
```
wahidyankf: ### Finding 6 — `apps/**` and `libs/**` are documented as out of scope, but the gate scans them

**Severity**: HIGH | **Confidence**: 90 | **Discipline**: Documentation-quality
*(Coordinator note: raised as CRITICAL; lowered to HIGH — latent trap, not a current breakage. Still blocking. Cross-references Finding 3, which is the opposite-direction defect on this same exclusion list.)*

The plan's `README.md` §Scope Boundaries states:

> **Out of scope**: `plans/**` (all three gates), `apps/**`, `libs/**`, root `README.md`/`CONTRIBUTING.md`/`LICENSING-NOTICE.md`, `ose-primer`, and the plan documents in this folder.

This comment correctly identifies that the `**/README.md` surface *"matches every README.md in the repo"* — and the exclusion list below it excludes `plans/`, `docs/`, `specs/`, `.fvm/`, `.fvm-cache/` and the `.opencode/` entries. It does **not** exclude `apps/` or `libs/`. The claimed exemption does not exist.

This PR proves it against itself: `apps/rhino-cli/README.md` was trimmed **1041 → 884 words** in this diff — a change made to satisfy a gate the spec says does not apply to that path. Coordinator-verified against the merge-base blob.

The live risk is `apps/ose-be/README.md` at **744 words** — already inside the 700-900 WARN band, 156 words from FAIL. Combined with the path-gated trigger list (which fires on `repo-governance/`, `.claude/`, `.opencode/`, `AGENTS.md`, `CLAUDE.md`, `repo-config.yml` — **not** on `apps/`), the failure mode is: an `apps/` README drifts past 900 words with the gate never firing, and then some unrelated governance-doc-only PR is blocked at push time by a file it never touched.

**Suggested fix**: pick one and make the other match. Either add `apps/` and `libs/` to this exclusion list (consistent with §Scope Boundaries as written), or delete the `apps/**`/`libs/**` clause from §Scope Boundaries and accept the blanket coverage — in which case the trigger list should arguably include `apps/**/README.md` so drift is caught by the PR that causes it.

plans/in-progress/optimize-governance-md/evidence/phase-9-trigger-matrix.txt
```diff
@@ -0,0 +1,130 @@
...
+Method
+------
+Isolated throwaway git fixture (not the real ose-public repo), same pattern as
+`phase-1-ci-path-gated.txt`: a minimal `repo-config.yml` carrying the two
+gates' real `surfaces:`/`trigger:` declarations verbatim (copied from the real
+`repo-config.yml`), exercised via the real CLI binary
+(`apps/rhino-cli/target/gate/rhino-cli`) end-to-end. No `origin/main` ref is
+created in the fixture, so `gate::run`'s `merge_base_paths()` falls back to
+`staged_paths()` (`git diff --cached --name-only`) — the same fallback the
+Phase 1 fixture relied on. `git reset && git clean -fd` between cases resets
```
wahidyankf: ### Finding 7 — destructive git operations in a fixture with none of the six mandated isolation layers documented

**Severity**: HIGH | **Confidence**: 88 | **Discipline**: Security

This line documents `git reset && git clean -fd` run between cases against a claimed *"isolated throwaway git fixture"*, driven by the real compiled binary end-to-end. `evidence/phase-9a-red.txt` describes the same pattern.

[`git-fixture-isolation.md`](https://github.com/wahidyankf/ose-public/blob/752f0e3952404305ffae96113502e765c687b5d1/repo-governance/development/quality/git-fixture-isolation.md) covers *"any test, fixture, or test-support helper … that invokes the `git` binary to **create or mutate** a throwaway repository"* and names `git reset --hard` among its triggering operations. Coordinator-verified: **neither transcript contains a single one** of the mandated layers — `grep -n 'GIT_DIR|GIT_CEILING_DIRECTORIES|GIT_CONFIG_GLOBAL|GIT_CONFIG_SYSTEM'` across both files returns nothing.

**Correction to the raw finding**: the specialist cited *four* required layers. The convention mandates **six** ([completeness checklist](https://github.com/wahidyankf/ose-public/blob/752f0e3952404305ffae96113502e765c687b5d1/repo-governance/development/quality/git-fixture-isolation/16-completeness-checklist.md)) — `GIT_CEILING_DIRECTORIES` (S1), explicit `GIT_DIR` (S2), `GIT_CONFIG_GLOBAL=/dev/null` + `GIT_CONFIG_SYSTEM=/dev/null` (S3), a pre-write escape guard comparing canonicalised `git rev-parse --show-toplevel` against the intended tempdir (S4), per-subprocess exit-status checks (S5), and throwaway-clone-only diagnosis (S6). None is evidenced here.

Standard 4 is the one that matters most: without an escape guard, a fixture whose `GIT_DIR` resolution silently falls back to ambient discovery runs `git clean -fd` **against the real checkout**. That is precisely the [motivating incident](https://github.com/wahidyankf/ose-public/blob/752f0e3952404305ffae96113502e765c687b5d1/repo-governance/development/quality/git-fixture-isolation/05-the-motivating-incident-part-1.md) this convention exists for. The repo's own in-process harness in `gate/run.rs` does implement these layers; this hand-run evidence transcript does not, so the safer path was available and not taken.

Kept at HIGH rather than CRITICAL: both transcripts show a clean post-run tree and no stray identity landed in commit history — the layers were absent, but no escape occurred.

**Suggested fix**: re-run the trigger matrix under the six layers and capture the env-var preamble verbatim in the transcript, so the evidence demonstrates isolation instead of asserting it. An unreproducible safety claim in an evidence file is worth less than the transcript it occupies.

repo-config.yml
```diff
@@ -1039,3 +1039,87 @@ gates:
           - docs/reference/platform-bindings.md
           - repo-governance/
       ci: { scope: other }
+
+  # Armed at Phase 9 of plans/in-progress/optimize-governance-md, dark-launched at Phase 1
+  # (config-only, see governance-word-budget: above). No pre-commit surface (FR-1.14): a
```
wahidyankf, line 1044: ### Finding 8 — this FR-1.14 decision contradicts a live convention doc that claims a pre-commit backstop exists

**Severity**: MEDIUM | **Confidence**: 95 | **Discipline**: Documentation-quality
**Raised by**: pr-review-governance-maker *(routed as likely-out-of-scope; coordinator re-categorized to documentation and **kept in scope**)*

This comment records FR-1.14's decision: *"No pre-commit surface … a whole-tree scan on every commit buys no additional coverage over pre-push/ci."* Reasonable. But `repo-governance/conventions/structure/governance-word-budget-remediation.md:25` states the opposite as fact:

> 2. **Pre-commit backstop**: `rhino-cli convention audit` includes the word-budget category.

Coordinator-verified false: `apps/rhino-cli/src/commands/convention_audit.rs:15` is
`const MEMBERS: &[&str] = &["emoji", "license", "agents-md-size"];` — no word-budget member, and nothing else in that file references it.

**On scope.** The doc is not in this PR's diff, so the coordinator considered dropping it under the scope guard. It is kept because (a) it was created by **this same plan's Phase 1** (`6b8cac957`, PR #187) and is therefore squarely inside the PR's declared plan scope, and (b) this PR is what makes the false claim consequential — before Phase 9 the gate was dark-launched and nobody relied on any enforcement point; from this commit on, a developer reading the remediation doc will believe a pre-commit backstop catches word-budget violations that this PR deliberately decided not to catch. The line anchor is this comment because that is where the contradiction is introduced.

Worth noting the other three Enforcement Points check out: #1 (pre-push) and #3 (`nx run rhino-cli:governance-word-budget:validation`, `project.json:190`) are now true as of this PR, and #4 (`repo-governance audit` includes the category) is true but **returns 155 false positives** — see Finding 1.

**Suggested fix**: delete Enforcement Point 2 and renumber, citing FR-1.14's rationale.

repo-governance/README.md
```diff
@@ -21,44 +21,44 @@ This directory is a map, not a replacement for the documents it links to. Start
 
 ## Start Here
 
-- To understand the whole model and how the layers relate, read the [Repository Governance Architecture](./repository-governance-architecture.md).
-- To learn why the project exists, begin with the [Vision](./vision/README.md).
+- To understand the whole model and how the layers relate, read the [Repository Governance Architecture](./repository-governance-architecture.md) — how the six layers relate and trace.
+- To learn why the project exists, begin with the [Vision](./vision/README.md) — why the project exists.
```
wahidyankf: ### Finding 9 — the annotation sweep produced verbatim self-duplication, not annotation

**Severity**: MEDIUM | **Confidence**: 92 | **Discipline**: Documentation-quality
*(Coordinator note: confidence raised 85 → 92 after reading the diff hunk directly; the specialist cited line 73, the defects are at lines 24-28, 30-36 and 55. The defect is also more concrete than "run-on" — it is literal repetition.)*

The mechanical `unannotated`-remediation sweep appended a `— <annotation>` to links whose surrounding prose **already states the same fact**, producing sentences that say the same thing twice. This line is the clearest case:

> - To learn why the project exists, begin with the [Vision](./vision/README.md) — why the project exists.

Three more in the same block (lines 24, 27, 28):

> - To understand the whole model and **how the layers relate**, read the [Repository Governance Architecture](…) — **how the six layers relate** and trace.
> - To find **a software practice, quality gate, or engineering workflow**, use [Development](…) — **software practices, quality gates, and engineering workflow**.
> - To run **a defined multi-step process**, use [Workflows](…) — **defined multi-step processes to run**.

Two further variants:

- **Table cells duplicate their own adjacent column** (lines 30-36). "Question it answers" already reads `Why does the project exist?`; "Where to go" now reads `[Vision](./vision/README.md) — why the project exists`. Same in 5 of 6 rows — and row 4 (AI Agents) got **no** annotation at all, so the sweep is inconsistent as well as redundant.
- **Mid-sentence interruption** (line 55): `The [Repository Governance Architecture](…) — how the six layers relate and trace — explains the relationships, traceability, and change impact in detail.` The interpolated clause restates what the second half of the sentence then says at greater length.

This is worth fixing rather than shrugging at, because it is the *first* output of the `unannotated` gate this PR arms — it sets the pattern every subsequent remediation across 721 directories will copy. An annotation that repeats its own sentence adds bytes against a word budget while adding no navigational information, which inverts the point of both gates.

**Suggested fix**: annotate only where the link text and surrounding prose do not already convey the target's purpose; in the §Start Here list and the layer table, drop the annotation rather than duplicate the adjacent column, and give row 4 parity with whatever rule is chosen.

apps/rhino-cli/src/application/governance/word_budget.rs
```diff
@@ -305,6 +310,14 @@ pub fn check_instruction_sizes(
             if is_in_skipped_dir(&entry) {
                 continue;
             }
+            let rel = entry.strip_prefix(repo_root).unwrap_or(&entry);
+            let rel_str = rel.to_string_lossy();
+            if excludes
+                .iter()
+                .any(|prefix| rel_str.starts_with(prefix.as_str()))
```
wahidyankf, line 348: ### Finding 10 — `--exclude` is prefix-matched while every sibling `--exclude` is glob-matched

**Severity**: MEDIUM | **Confidence**: 85 | **Discipline**: Architecture

`rel_str.starts_with(prefix.as_str())` is a plain string-prefix test. The sibling implementations behind the identically-named `--exclude` flag use glob semantics: `readme_index.rs`'s `matches_any_glob` and `governance_audit.rs`'s `exclude_globs`. Same flag name on the same CLI, three surfaces, two different matching rules, none of them documented at the flag.

No live bug: every entry currently registered ends in `/` or is an exact path, where prefix and glob agree. The failure is latent and quiet — the first person to write `--exclude '.opencode/skills/*'` or `--exclude '**/generated/**'` by analogy with the flag's siblings gets a silent no-match. Nothing fails; the exclusion simply does not happen, and the gate reports findings for paths the author believes are excluded. That is the worst shape for a config error.

Prefix matching may well be the right call here (it is simpler and sufficient for path-tree exclusion). The gap is that the choice is invisible at every point of use.

**Suggested fix**: state the semantics in the `ValidateWordBudgetArgs::exclude` doc comment and in the `check_instruction_sizes` doc comment above (`"repo-relative path **prefixes**, not globs — `.opencode/skills/` matches, `.opencode/skills/*` does not"`), and mirror it in the `repo-config.yml` comment. Alternatively converge on `matches_any_glob` so one flag name means one thing across the CLI.

apps/rhino-cli/src/application/repo_governance/audit_orchestrator.rs
wahidyankf: Fixed: added `word_budget::registered_excludes(repo_root)` (`apps/rhino-cli/src/application/governance/word_budget.rs`), which reads the `governance-word-budget` gate's `args.exclude` list from `repo-config.yml`'s `gates:` registry — the single source all three call sites now read from instead of each hand-rolling its own default:

- `apps/rhino-cli/src/application/repo_governance/audit_orchestrator.rs::audit_word_budget` now passes `word_budget::registered_excludes(repo_root)` instead of `&[]`.
- `apps/rhino-cli/src/commands/governance_validate_word_budget.rs::run` (the bare CLI entry) now merges `registered_excludes(&repo_root)` with any explicit `--exclude` flags before calling `run_for_root`.

Re-verified both cited reproductions against the built `752f0e395`+fix binary:

```
$ rhino-cli governance word-budget validate; echo $?
... 0 [FAIL] findings ...
0
$ rhino-cli repo-governance audit --include-category governance-word-budget
GOVERNANCE AUDIT PASSED: 0 findings across 1 categories
  [PASS] governance-word-budget           0 finding(s)
```

Both surfaces now exit 0 with zero false positives against `.fvm/` and the other excluded trees, with no `--exclude` flags supplied.

apps/rhino-cli/src/application/docs/frontmatter.rs
wahidyankf: Fixed — reconciled the Gherkin scenario to FR-4's armed contract instead of narrowing the arming, and closed the CI-visibility gap on `tests/docs.rs`:

- `specs/apps/rhino/behavior/rhino-cli/gherkin/md/docs-validate-frontmatter.feature`: the old "Governance doc with only title passes the lighter schema" scenario described the pre-Phase-9 WARN behavior and is now false under the armed contract (`prd.md` FR-4.1/FR-4.2 — `when_to_use`/`description` are FAIL-severity once armed). Replaced it with two scenarios: a title-only governance doc now correctly **fails** (missing `description` + `when_to_use`), and a new scenario asserts the actual "lighter schema" — a governance doc with `title`+`description`+`when_to_use` still passes without needing the software-engineering schema's `category`/`subcategory`/`tags`.
- `apps/rhino-cli/tests/docs.rs`: added the corresponding step definitions (`given_fm_gov_title_description_when_to_use`, `then_fm_missing_when_to_use`, `then_fm_missing_description`).
- `apps/rhino-cli/project.json`: added `--test docs` to `rhino-cli:test:unit`'s command (plus the `tests/docs.rs` and `gherkin/md/**/*.feature` cache `inputs`), so this test file — previously orphaned from every CI-run target — is now exercised by the same target CI actually invokes. Confirmed by running the exact updated `test:unit` command locally: all 7 named test binaries pass, including `docs` (7 features, 80 scenarios, 319 steps, all green).

repo-config.yml
wahidyankf: Fixed, addressing both (a) and (c); (b) is corrected in the doc comment.

(c) — collapsed the 7 enumerated `.opencode/skills/<name>/` entries plus the single-file `.opencode/commands/monitor-ci.md` entry into two tree-level prefixes, `.opencode/skills/` and `.opencode/commands/` (`repo-config.yml` `governance-word-budget.args.exclude`). Verified `.opencode/commands/monitor-ci.md` is in fact the *only* file in that directory, so the tree-level prefix is exactly the same class as `.opencode/skills/` — no `.claude/` source of truth for either tree — and both are now future-proof against an Nx regeneration adding an 8th skill dir or a 2nd command file. This also makes the "not a per-file waiver" disclaimer literally true again (it previously sat directly above a single-file entry that contradicted it).

(b) — the `repo-config.yml` comment no longer analogizes `.opencode/skills/`/`.opencode/commands/` with `.fvm/`; it now states explicitly that both are git-tracked and present in every clone/CI run (unlike the gitignored, CI-absent `.fvm/`), and are excluded because the whole tree has no `.claude/` source of truth, not because they're locally-absent.

(a) — added `prd.md` FR-1.16, recording the gate-registration-layer `args.exclude` carve-out as an explicit requirement (not left as a YAML comment contradicting FR-1.4/FR-1.5 as written): FR-1.5's no-exemption-schema guarantee is unchanged for every file `args.exclude` doesn't name, and FR-1.4 holds without exception for every mirror this plan itself produces.

repo-config.yml
wahidyankf: Fixed: added a `governance-readme-completeness` mention to `AGENTS.md` §Quality Gates ("READMEs need annotated indexes") linking a new convention doc, `repo-governance/conventions/structure/governance-readme-completeness.md`, which documents both `governance-readme-index` (`orphan`/`ghost`) and `governance-readme-completeness` (`missing`/`unannotated`) — the finding kinds, scan scope, remediation, and enforcement points — matching the pattern the word-budget gate already sets. Also linked from `repo-governance/conventions/structure/README.md`'s index (required to avoid a new `orphan` finding from the readme-index gate itself).

`AGENTS.md` is right at the word-budget ceiling (496w before this fix), so the addition is deliberately terse and offset by two unrelated trims elsewhere in the file (`Trunk Based Development` → `Trunk-Based Dev`, dropped a redundant "to" in the Git Workflow section) to land at 499/500 words — still WARN, not FAIL. Re-verified via `rhino-cli governance word-budget validate`.

repo-config.yml
wahidyankf: Fixed by amending the spec docs to the narrowed 4-entry reality (the option that matches `delivery.md`'s already-recorded, user-approved rationale — widening back to 6 would reintroduce the ~880 pre-existing `docs/`/`specs/` findings the narrowing was explicitly done to avoid):

- `prd.md` FR-3.7: now states the shipped 4-entry list directly, with a note recording the pre-narrowing 6-entry history and the Phase 9/16 narrowing rationale. FR-3.20 updated to stop calling it "the widened FR-3.7 6-entry scope." The FR-1.11 prose and its `args.paths`/`trigger` YAML sample block are updated to the actual 4/5-entry lists (dropping `docs/`, `specs/`).
- `prd.md` item 4 in the FR-3 rename-and-extend list also had a **separate**, pre-existing contradiction discovered while fixing this: it claimed `DEFAULT_PATHS` widens to 6 entries, which directly contradicts FR-3.19's own "current `DEFAULT_PATHS` (unwidened)" continuity guarantee two paragraphs later — and contradicts the shipped code (`readme_index.rs::DEFAULT_PATHS` is verified unchanged). Corrected to describe what actually changed: the new gate's own `--paths` flag, never the constant.
- `tech-docs.md` §1.1 and §4 (the mechanism table and the `--paths` bullet): same "6-entry"/"widened" language corrected to the narrowed 4-entry reality with a pointer to `prd.md` FR-3.7.
- `README.md` §Scope Boundaries: "the wider FR-3.7 list" replaced with the actual 4-entry list plus the narrowing rationale.

repo-config.yml
wahidyankf: Fixed in `plans/in-progress/optimize-governance-md/README.md` §Scope Boundaries — corrected the claim rather than the gate (the gate's blanket `apps/**`/`libs/**` coverage is the intended, already-shipped behavior per this PR's own `apps/rhino-cli/README.md` trim, so the doc was wrong, not the config):

- The word-budget "In scope" paragraph now explicitly names the blanket `**/README.md` surface and states it covers `apps/**`/`libs/**` regardless of tree.
- The "Out of scope" line now scopes the `apps/**`/`libs/**` exclusion to the README-index/completeness and frontmatter gates only, and adds an explicit carve-out: "`apps/**`, `libs/**`, and root `README.md` are not out of scope for the word budget," citing the `apps/rhino-cli/README.md` trim as proof.

While verifying this I found the same class of gap one level further: the original "Out of scope" line also claimed root `README.md` was excluded, and it is **not** — `rhino-cli governance word-budget validate` reports it directly (856 words, WARN band, 44 words from FAIL). Corrected that too, rather than leaving a second false claim of the same shape newly visible right next to the one this thread named.

plans/in-progress/optimize-governance-md/evidence/phase-9-trigger-matrix.txt
wahidyankf: Fixed by genuinely re-running both cited proofs under the six mandatory layers and recording the isolation preamble verbatim, rather than just annotating the existing transcripts.

Re-executed `evidence/phase-9-trigger-matrix.txt`'s 5-case matrix and `evidence/phase-9a-red.txt` Part 1's oversized-fixture proof in a fresh throwaway git fixture (built, exercised, and destroyed from this session's scratchpad, never the primary checkout — Standard 6), with `GIT_DIR`/`GIT_CEILING_DIRECTORIES` explicitly set to the fixture root and `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM=/dev/null` (Standards 1-3), a `git rev-parse --show-toplevel` escape guard asserted before every write — `init`, both `config` writes, every `add`, and every `reset --hard`/`clean -fd` between cases (Standard 4) — and exit-status checks on every invocation (Standard 5). Both transcripts now document this preamble and confirm the guard never tripped. All 5 trigger-matrix cases and the oversized-fixture case reproduced the exact same outcomes as the original recording, so the underlying proof stands — only the isolation evidence was missing, and that's what this fix supplies.

`phase-9a-red.txt` Part 2 (the frontmatter WARN baseline) is unchanged — it never invoked `git` to create or mutate a repository (a plain file add/delete against the real checkout, with no destructive git command), so it's outside `git-fixture-isolation.md`'s scope.

Note: this convention's scope is repo-wide and other pre-2026-08-15 evidence transcripts in this plan (e.g. `phase-1-ci-path-gated.txt`, which this very finding's cited pattern traces back to) likely share the same gap. Re-recording the entire plan's evidence corpus is out of scope for this fixer pass; flagging it here as a known follow-up rather than silently leaving it unaddressed.

repo-config.yml
wahidyankf: Fixed: deleted Enforcement Point 2 ("Pre-commit backstop: `rhino-cli convention audit` includes the word-budget category") from `governance-word-budget-remediation.md` and renumbered, with a note citing FR-1.14's rationale (no pre-commit surface; pre-push/CI already cover it) and the `convention_audit.rs::MEMBERS` fact that made the original claim false.

While fixing this I found the identical false claim, verbatim in spirit, in the sibling convention doc `governance-word-budget.md` §Enforcement Points ("Runs at pre-push ..., pre-commit (`convention audit`), CI, ..."). Fixed that occurrence too — same defect class, same root cause, would have re-surfaced this exact finding against a different file in a later cycle otherwise.

repo-governance/README.md
wahidyankf: Fixed: rewrote the annotations across all four affected sections (§Start Here lines 24-28, the §Navigate the Layers table lines 34-41, the §Follow the Trace mid-sentence interruption, and the §Choose the Right Home table) so each link's `— <annotation>` states something the surrounding prose or adjacent column doesn't already say — generally the target's own frontmatter `description` — instead of restating the lead-in clause or the adjacent column verbatim. Also gave the AI Agents table row (previously bare) an annotation for parity, since it's the shortest of the four sections to fix consistently and the inconsistency was called out explicitly.

Every target's annotation text is now reused verbatim at each of its repeated occurrences in the file (Start Here / table / Choose the Right Home / Follow the Trace / Read by Situation), which is what the annotation format is meant to do — restate the target's stable identity, not re-derive fresh prose per occurrence.

Re-verified: `governance readme-index validate --paths repo-governance/ --paths .claude/ --paths .codex/ --paths .pi/ --fail-kinds missing --fail-kinds unannotated` passes (no `unannotated` finding on this file), and the unconditional `orphan`/`ghost` gate passes too. Net effect also trims the file from 856 to 800 words (still WARN, not FAIL, against the 900-word `**/README.md` ceiling).

apps/rhino-cli/src/application/governance/word_budget.rs
wahidyankf: Fixed: documented the prefix-vs-glob semantics at all three declaration sites, per the suggested fix — `check_instruction_sizes`'s doc comment, `ValidateWordBudgetArgs::exclude`'s doc comment, and the `repo-config.yml` comment above `governance-word-budget`'s `args.exclude` block — each now states explicitly that this is a `str::starts_with` path-prefix match, not a glob, contrasting it with the identically-named glob-matched `--exclude` on `md links`/`md mermaid validate`, and giving the `.opencode/skills/` (matches) vs `.opencode/skills/*` (matches nothing) example from the suggested fix verbatim.

Kept the matching semantics as prefix rather than converging on `matches_any_glob`, since the finding itself notes prefix matching "may well be the right call here (it is simpler and sufficient for path-tree exclusion)" and no live bug exists today — the fix closes the documentation gap the finding actually identifies as the problem.

PR #309, opened 2026-08-26
feat(rhino-cli): scaffold F# project, dispatch shim, and CI wiring (Phase 2)

wahidyankf:
## Why

Phase 2 of `rewrite-rhino-cli-to-fsharp`: scaffold the F# project, add the dispatch shim, and wire
CI so the new binary participates in `pr-quality-gate.yml` alongside the still-live Rust crate.
Includes Phase 1's publish-mode spike (self-contained, non-AOT — see `learnings.md`), which carries
no reviewable change of its own and folds into this PR per the plan's Delivery Boundaries table.

## Scope

- New F# console project (`apps/rhino-cli/src-fsharp/`) targeting `net10.0`, self-contained non-AOT
  publish per Phase 1's decision.
- Dispatch shim wiring so CI can build/run both the Rust and F# binaries during the migration.
- `pr-quality-gate.yml`: new `dotnet` job, `detect` job's `has-dotnet-projects` output (new case in
  `ose-private`, pre-existing in `ose-public`), artifact download/export wiring in `format`,
  `enumerate`, and `gate` jobs.
- Placeholder modules with no executable `let` bindings, to avoid a 0%-coverage failure against the
  90% `test:coverage` threshold before real logic lands in later waves.
- `plans/in-progress/rewrite-rhino-cli-to-fsharp/delivery.md`, `learnings.md`: Phase 1 + Phase 2
  checklist items ticked, judgment calls recorded.

## Non-goals (deliberately out of scope, with reasons)

- **No namespace routing changes yet** — `FSHARP_NAMESPACES` ships empty, so every CLI namespace
  still dispatches to the Rust binary. Routing a real namespace to F# is Wave A's job, not
  Phase 2's; this PR only proves the scaffold and dispatch shim can carry a namespace once one is
  added.
- **No real F# business logic** — the five new projects hold placeholder modules only (see Scope
  above). Specs/behavior coverage for actual logic is deferred to Waves A-F, one namespace at a
  time, per the plan's wave map.
- **No RID/runner-architecture verification for `ose-private`** — `linux-x64` is assumed for that
  repo's self-hosted runner (see Known deviations below); confirming the real hardware
  architecture is out of this PR's scope and tracked as a pre-merge follow-up in `learnings.md`.

## Where to start reading

Start with `apps/rhino-cli/src-fsharp/RhinoCli.Program/src/Program.fs` (the entry point), then
`apps/rhino-cli/scripts/rhino-bin.sh` (the Rust/F# dispatch shim it plugs into). From there,
`apps/rhino-cli/scripts/shadow-diff.sh` shows how the two binaries are proved byte-identical, and
`.github/workflows/pr-quality-gate.yml`'s `build-rhino`/`dotnet` jobs show how CI wires the new
binary in.

## Paths to skip

`apps/rhino-cli/parity-manifest.sha256` and `apps/rhino-cli/Cargo.lock` are machine-regenerated by
the parity-manifest tooling and `cargo`, respectively — not hand-edited, safe to skip.

## Gates run and their results

- `dotnet build` — pass, both repos.
- `parity manifest validate` — pass, byte-identical `apps/rhino-cli/` across both repos (build
  artifacts under gitignored `bin/`/`obj/` excluded, as always).
- `git diff`/`grep` assertions per delivery.md's Phase 2 acceptance clauses — pass.
- CI: pending on this PR (will report once run).

## Known deviations from the plan text (recorded, not hidden)

- `Severity` DU renamed `Error`/`Warn` → `Blocking`/`Advisory` — `GRA-UNIONCASE-001` forces this;
  `RequireQualifiedAccess` alone does not suppress it (confirmed empirically).
- `linux-x64` RID assumed for `ose-private`'s self-hosted runner architecture.
- `shadow-diff.sh` was a pre-existing dead Go-era file in both repos, not new — repurposed, not
  created fresh (tech-docs.md's file-impact table was wrong on this point).
- `detect` job's `lang:fsharp` → `has-dotnet-projects` mapping was pre-existing in `ose-public` but
  entirely new in `ose-private` — delivery.md's assumption that "no edit is needed" was corrected
  to be per-repo accurate.

Full reasoning for all judgment calls is in `plans/in-progress/rewrite-rhino-cli-to-fsharp/learnings.md`.

---

## Review route record — cycle 1

- **Pinned base / head**: base `main` @ merge-base `851e83bf9b5e68d181784dfe83ae45f7f1d18fae`; head
  `f35425abd7c8c1014391fc8b0e8b4ef5a356f1be`. Every duty this cycle anchors to this head SHA.
- **Frozen outcome / scope**: Phase 2 of `rewrite-rhino-cli-to-fsharp` — F# project scaffold, the
  `rhino-bin.sh` namespace-dispatch shim (`FSHARP_NAMESPACES` empty), the rewritten
  `shadow-diff.sh` differential runner, the `dotnet-deps-audit.sh` gate wrapper, Nx project
  `rhino-cli-fsharp`, `repo-config.yml` coverage registration, `pr-quality-gate.yml` publish +
  artifact wiring, the parity-manifest/boundary-path update, and the Phase 1 tree-sitter drop
  folded in per the plan's Delivery Boundaries table. No namespace routes to F# yet.
- **Classification evidence**: 35 files, +2520 / -471 against the merge-base;
  `.github/workflows/pr-quality-gate.yml` is in the diff (CI/workflow = security-sensitive path).
- **Risk tier**: `full` — triggered independently by line count, file count, and the
  security-sensitive path; the path trigger alone forces `full` regardless of size.
- **Selected lenses (9 of 9)**: architecture (five-project layer split, shim seam),
  logic (shim routing, shadow-diff subcommand walk, audit-wrapper jq predicate),
  integrity (CI artifact flow, Nx target wiring, coverage/spec-gate scoping — test and workflow
  files present), types (F# DU/record scaffold — `.fs` files present), security (workflow edits,
  artifact download + `chmod +x`, dependency-audit gate), governance (Nx mandatory targets,
  registry hygiene, parity-manifest boundary), performance (self-contained publish, CI job cost),
  docs and instruction (plan prose and in-file rationale comments, reviewed once at cycle 1).
- **Skipped lenses**: none. The Content-Type Applicability Filter excludes neither `types` nor
  `integrity` on this diff, and it is re-derived fresh each cycle.
- **Current checks**: none reported on this branch yet at classification time.
- **Settled history**: none — this is cycle 1. No prior reviews, no prior review threads, no
  human dismissals to respect.
- **Probe (cycle 1)**: *scaffold honesty at the migration seams* — does each new artifact do what
  its own comment and the plan's acceptance clause claim, specifically at the four seams where a
  scaffold can ship green while being inert or wrong (shim routing, CI artifact flow, Nx target
  wiring, coverage/spec-gate scoping). First use of this class on this PR.
- **Cadence**: capped at 2 cycles by explicit user direction for this plan's execution phases; a
  cycle-3 continuation is replaced by a recorded orchestrator override.


apps/rhino-cli/scripts/shadow-diff.sh
```diff
@@ -1,82 +1,158 @@
...
-  echo "Divergence artefacts saved to: ${ARTIFACT}" >&2
-  exit 1
+	local mismatch=0
+	if [[ "${rust_exit}" -ne "${fsharp_exit}" ]]; then
+		mismatch=1
+	fi
+	if ! diff -q "${rust_stdout}" "${fsharp_stdout}" >/dev/null 2>&1; then
+		mismatch=1
+	fi
+	if ! diff -q "${rust_stderr}" "${fsharp_stderr}" >/dev/null 2>&1; then
```
wahidyankf, line 116: ### C1-F1 — `shadow-diff.sh`'s byte-identity comparison is unsound for most namespaces, and already fails two Gate clauses this PR commits to

**Severity**: HIGH | **Confidence**: 97
**Raised by**: logic

**What is wrong.** `compare_invocation` compares the two binaries' stdout with a raw `diff -q` and no field masking. Several rhino-cli namespaces print values that are *different on every run of the same binary* — wall-clock timestamps and elapsed-time measurements. So the harness reports a "difference" for output that is in fact identical in every respect that matters, and it will do so forever, no matter how faithful the F# port becomes.

**Reproduced** at head `f35425abd7c8c1014391fc8b0e8b4ef5a356f1be`, by pointing *both* sides at the same Rust gate binary — i.e. comparing a binary against itself, which must by construction report zero differences:

```
SHADOW_DIFF_FSHARP_BIN="$PWD/apps/rhino-cli/target/gate/rhino-cli" \
  bash apps/rhino-cli/scripts/shadow-diff.sh <namespace>
```

| namespace | result | exit |
| --- | --- | --- |
| `convention` | 15 compared, **0 differences** | 0 |
| `doctor` | 5 compared, **2 differences** | 1 |
| `repo-governance` | 20 compared, **4 differences** (3/3 consecutive runs, identical count) | 1 |
| `harness` | 55 compared, **23 differences** | 1 |

Two distinct causes, both visible in the emitted diffs:

- **Timestamps** — `repo-governance` prints `ran_at=2026-08-26T05:33:46Z` (`apps/rhino-cli/src/application/repo_governance/audit_orchestrator.rs`); `doctor -o json` prints `"timestamp": "2026-08-26T12:34:30+07:00"` and `doctor -o markdown` prints `**Generated**:` (`apps/rhino-cli/src/application/doctor/reporter.rs:206`). These are at least *pinnable* in principle, via a clock injection or a `--now` flag — neither exists today.
- **Elapsed-time measurements** — `harness claude validate` prints `Duration: 3.153959ms` vs `Duration: 4.697875ms`, and `-o json` prints `"duration_ms": 3`. This is the harder half: a duration is not pinnable by any flag, and two *different language runtimes* will never produce identical timings. No amount of port fidelity makes these fields match.

**Why it matters now, not later.** This is not a hypothetical. `delivery.md` already commits to acceptance clauses that this harness reproducibly cannot satisfy:

- `delivery.md:4104` and `delivery.md:4149` (Wave C Gate) — `shadow-diff.sh doctor test-coverage` must report zero differences. `doctor` alone reports 2.
- `delivery.md:12210` and `delivery.md:12255` (Wave E Gate) — `shadow-diff.sh harness specs repo-governance` must report zero differences. `harness` reports 23 and `repo-governance` reports 4, independently.

**Why the scaffold gate did not catch it.** Phase 2's own smoke test (`delivery.md:747`) runs only `shadow-diff.sh convention`, which is the one namespace above that prints no timestamp and no duration — so it passes cleanly and masks the whole defect class at exactly the moment the harness is being certified as "sound before it is trusted" (the script's own header comment, lines 5-7). Neither `tech-docs.md` nor `delivery.md` mentions non-determinism anywhere.

**What would fix it** (author's choice — this review does not prescribe): add field masking (a normalizing filter applied to both sides before `diff`), or add a clock/duration-suppression flag to the CLI and set it in the harness, or narrow the compared surface. Whichever route, the two Gate clauses above need to become satisfiable, and the non-determinism class needs to be written down.

**Refutation clause.** If

```
SHADOW_DIFF_FSHARP_BIN="$PWD/apps/rhino-cli/target/gate/rhino-cli" \
  bash apps/rhino-cli/scripts/shadow-diff.sh doctor harness repo-governance
```

exits 0 and prints `0 difference(s)` on three consecutive runs, this finding is void.

---

Generated by AI

wahidyankf (commented):
## PR Review — Cycle 1 of 2

**Cycle**: 1 of 2
**Head SHA**: f35425abd7c8c1014391fc8b0e8b4ef5a356f1be
**Diff size**: 2520 added / 471 deleted lines, 35 changed files (33 hand-authored, excluding the generated `apps/rhino-cli/parity-manifest.sha256` and `apps/rhino-cli/Cargo.lock`)
**Risk tier**: full
**Probe class**: scaffold-honesty-at-the-migration-seams (new on this PR)
**Per-specialist raw findings**: architecture 2, logic 1, governance 2, security 0, integrity 1, performance 0, docs 2, instruction 0, types 1
**Security-sensitive-path override applied**: yes (`.github/workflows/pr-quality-gate.yml` in diff)
**Prior-cycle human dismissals respected**: none this cycle (cycle 1)

**This review is blocking despite its `COMMENT` state** — `REQUEST_CHANGES` is structurally unavailable to this agent on its own PR. Two `HIGH` findings (C1-F1, C1-F2) carry the blocking status.

### Synthesis notes

- **One finding dropped by tool-verify.** The `types` specialist claimed `Types.fs`'s `GRA-UNIONCASE-001` justification was factually wrong — that the analyzer bails out when `[<RequireQualifiedAccess>]` is present, making the `Error`/`Warn` → `Blocking`/`Advisory` rename unnecessary. I ran the analyzer directly (this repo's pinned `G-Research.FSharp.Analyzers` 0.22.0 analyzer binaries via the repo's own `fsharp-analyzers` 0.36.0 tool manifest) against a throwaway project containing exactly `[<RequireQualifiedAccess>] type Severity = | Error | Warn`. It **did** emit `GRA-UNIONCASE-001` (exit 254). The finding's own refutation clause said "if it reports a diagnostic, finding is void" — so it is void. **`Types.fs:11-15` and `learnings.md`'s entry are correct as written**; no change is owed there.
- **Two findings merged into one.** `docs` and `architecture` independently found the same `tech-docs.md` diagram-direction defect (C1-F4). Independent convergence is confirmation, not two findings.

 CLI' plans/in-progress/rewrite-rhino-cli-to-fsharp/tech-docs.md"},
  {"id":"C1-F5","severity":"MEDIUM","discipline":"docs","confidence":88,"file":"plans/in-progress/rewrite-rhino-cli-to-fsharp/tech-docs.md","line":645,"raised_by":["docs"],"refutable_by":"grep -n 'No edit' plans/in-progress/rewrite-rhino-cli-to-fsharp/tech-docs.md"},
  {"id":"C1-F6","severity":"MEDIUM","discipline":"governance","confidence":95,"file":"plans/in-progress/rewrite-rhino-cli-to-fsharp/README.md","line":3,"raised_by":["governance"],"refutable_by":"git diff --numstat 851e83bf9b..f35425abd7 | awk '{a+=$1;n++} END {print a, n}'"},
  {"id":"C1-F7","severity":"MEDIUM","discipline":"governance","confidence":85,"file":"plans/in-progress/rewrite-rhino-cli-to-fsharp/README.md","line":4,"raised_by":["governance"],"refutable_by":"gh pr view 309 --json body -q .body | grep -Ei 'start (here|reading)|skip'"}]}
-->

---

Generated by AI

apps/rhino-cli/src-fsharp/project.json
```diff
@@ -0,0 +1,188 @@
...
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "if grep -rn -E --include='*.fs' 'Skip\\s*=' apps/rhino-cli/src-fsharp/tests/integration; then echo 'ERROR: xunit Skip= attribute found in test files above'; exit 1; fi && dotnet test apps/rhino-cli/src-fsharp/tests/integration/RhinoCli.IntegrationTests.fsproj"
+      },
+      "cache": false
+    },
+    "test:coverage": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "dotnet test apps/rhino-cli/src-fsharp/tests/unit/RhinoCli.UnitTests.fsproj /p:CollectCoverage=true /p:Threshold=90 /p:ThresholdType=line \"/p:ExcludeByFile=**/RhinoCli.Program/src/Program.fs\""
```
wahidyankf, line 77: ### C1-F2 — this `test:coverage` 90% gate is never invoked by any CI job

**Severity**: HIGH | **Confidence**: 95
**Raised by**: integrity

**What is wrong.** This target enforces a 90% line-coverage threshold through coverlet (`/p:Threshold=90 /p:ThresholdType=line`), and it works correctly when run by hand. But nothing in CI ever runs it, so the threshold it declares is not enforced on any pull request.

**Evidence** at head `f35425abd7c8c1014391fc8b0e8b4ef5a356f1be`:

- `.github/workflows/pr-quality-gate.yml` contains **exactly one** `test:coverage` invocation, at line 409, inside the `rust` job:
  `npx nx affected -t test:coverage --exclude='tag:lang:ts,tag:lang:fsharp,tag:lang:csharp,tag:lang:dart' ...`
  — which explicitly excludes `tag:lang:fsharp`.
- The `dotnet` job (lines 364-377), the one gated on `has-dotnet-projects` and therefore the only job that runs for this project, invokes `typecheck lint test:quick specs:behavior:coverage` only. `test:coverage` is not in that list.
- `rhino-cli-fsharp` carries `tag:lang:fsharp` (`project.json:6`), so it is excluded by the one invocation that exists and absent from the one job that would run it.

**This is a pre-existing gap being extended, not created.** The `tag:lang:fsharp` exclusion pre-dates this PR (it already applied to `crane-cli`) and this PR's workflow edits touch neither list. That is why this is HIGH and not CRITICAL, and why the finding is not a claim that the author broke something.

**Why it is still this PR's finding.** Two reasons, both specific to this diff:

1. `delivery.md:523-525` states the target is authored *"with the same `--fail-under-lines 90` threshold the Rust target enforces, wired through `coverlet`, so the threshold survives the language change rather than being re-derived at Phase 9d."* The first half is true; the conclusion is not. The threshold does not survive the language change, because the Rust target's threshold is enforced *by CI calling it* and the F# target's is not called. A future reader of that line will believe parity exists.
2. It bites the very next PR. Nothing ships broken today only because the placeholder modules contain no coverable code. Wave A starts landing real F# logic, which can land at 0% coverage with CI fully green.

**Refutation clause.** If `grep -n "test:coverage" .github/workflows/pr-quality-gate.yml` returns an invocation whose `--exclude` list does *not* contain `tag:lang:fsharp`, or if `test:coverage` appears in the `dotnet` job's target list, this finding is void.

---

Generated by AI

.github/workflows/pr-quality-gate.yml
```diff
@@ -229,6 +237,19 @@ jobs:
...
           path: apps/rhino-cli/target/gate/rhino-cli
+      # F# rhino-cli port (rewrite-rhino-cli-to-fsharp): publishes alongside
+      # the Rust binary from Phase 2 onward so every downstream job can
+      # resolve either binary through the dispatch shim. Self-contained,
+      # non-AOT per the Phase 1 publish-mode decision
+      # (learnings.md#publish-mode-decision) — bundles the .NET runtime so
+      # the eight consumer jobs below stay toolchain-free, unlike a
+      # framework-dependent publish.
+      - uses: ./.github/actions/setup-dotnet
+      - run: dotnet publish apps/rhino-cli/src-fsharp/RhinoCli.Program/RhinoCli.Program.fsproj -c Release -r linux-x64 --self-contained true -o apps/rhino-cli/src-fsharp/dist
```
wahidyankf: ### C1-F3 — CI hand-writes its own `dotnet publish` instead of invoking the Nx `build` target, and the two have already diverged

**Raised by**: architecture

**What is wrong.** There are now two independently-authored commands that produce the same artifact at the same path, and nothing keeps them in agreement:

- `apps/rhino-cli/src-fsharp/project.json:13` (the Nx `build` target):
  `dotnet publish ... -c Release --self-contained true --use-current-runtime -o apps/rhino-cli/src-fsharp/dist`
- this line (CI `build-rhino`):
  `dotnet publish ... -c Release -r linux-x64 --self-contained true -o apps/rhino-cli/src-fsharp/dist`

They already differ: `--use-current-runtime` versus `-r linux-x64`. CI never calls `npx nx run rhino-cli-fsharp:build`.

**Why it matters.** The two agree today only by coincidence — the GitHub runner happens to be linux-x64, so `--use-current-runtime` resolves to the same RID the CI command hardcodes. That coincidence is load-bearing and undocumented. The moment either side changes (a different runner architecture, an added publish flag, a trimming or `PublishReadyToRun` setting added to the Nx target), CI silently ships a *different* binary than the one `nx build` produces locally — and the difference will surface as a confusing shadow-diff or gate failure, not as a build error.

`apps/rhino-cli/scripts/rhino-bin.sh:41-43` compounds this by describing `dist/rhino-cli-fsharp` as *"the `build` Nx target's output; also what CI's `build-rhino` job uploads"* — a single sentence that presents two separately-maintained commands as one thing. A reader will reasonably conclude that changing the Nx target changes what CI publishes. It does not.

**Note on the PR's stated rationale.** The self-contained non-AOT decision itself is well-reasoned and recorded (`learnings.md#publish-mode-decision`); this finding is not about the publish mode. It is only about the command being duplicated rather than delegated.

**Refutation clause.** If `grep -n 'rhino-cli-fsharp:build' .github/workflows/pr-quality-gate.yml` returns a line inside the `build-rhino` job (i.e. CI invokes the Nx target), or if the CI command string is byte-identical to `project.json`'s `build` target command, this finding is void.

---

Generated by AI

plans/in-progress/rewrite-rhino-cli-to-fsharp/tech-docs.md
```diff
@@ -192,6 +199,16 @@ flowchart LR
   CLI --> PROG[RhinoCli.Program<br/>entry point, exit-code mapping]
```
wahidyankf: ### C1-F4 — this architecture diagram draws the `Program`/`Cli` edge backwards, and the correction was recorded only in the file this same document calls transient

**Severity**: MEDIUM | **Confidence**: 98
**Raised by**: docs, architecture *(independently found by both — convergence, not two findings)*

**What is wrong.** This line reads `CLI --> PROG`, meaning `RhinoCli.Cli` depends on `RhinoCli.Program`. The shipped code is the opposite:

```
apps/rhino-cli/src-fsharp/RhinoCli.Program/RhinoCli.Program.fsproj:42
    <ProjectReference Include="../RhinoCli.Cli/RhinoCli.Cli.fsproj" />
```

Every *other* edge in this diagram matches its `ProjectReference` exactly — `CLI --> APP` (`RhinoCli.Cli.fsproj:30`), `APP --> DOM` and `APP --> INF` (`RhinoCli.Application.fsproj:29-30`). So the diagram's convention is unambiguous: `X --> Y` means X references Y. Under that convention this one edge is inverted, and it is the edge governing the entry point.

**The code is right; the document is wrong.** `RhinoCli.Program` is the `Exe` and referencing the parser layer beneath it is the conventional layering — the author reasoned this out correctly in `learnings.md:362-373`.

**Why this is still a finding.** The correction lives *only* in `learnings.md`. This very document, twelve lines below this diagram, cites [the transient-log caveat](repo-governance/development/quality/knowledge-capture/the-transient-log-caveat.md) to establish that `learnings.md` "is a transient running log a future date may delete" — and uses that as the stated reason to keep durable facts in `tech-docs.md` instead. Recording the fix in the disposable file and leaving the durable file wrong is exactly the failure mode that paragraph exists to prevent. Fix the diagram here; the `learnings.md` entry can then stand as the rationale rather than as the only record.

**On the PR body's framing.** "Known deviations" does not mention this one at all (it lists the `Severity` rename, the `linux-x64` RID, `shadow-diff.sh`'s provenance, and the `detect` mapping). The deviation is real and should be listed, or — better — eliminated by correcting line 199.

**Scope note.** The `learnings.md` claim that "no wave's plan text depends on the literal arrow direction" is accurate as far as I can verify, so this is MEDIUM, not HIGH. It is a durable-record accuracy defect, not a blocked downstream step.

**Refutation clause.** If `grep -n 'PROG --> CLI' plans/in-progress/rewrite-rhino-cli-to-fsharp/tech-docs.md` returns a line, or if `grep -c 'RhinoCli.Cli.fsproj' apps/rhino-cli/src-fsharp/RhinoCli.Program/RhinoCli.Program.fsproj` returns 0, this finding is void.

---

Generated by AI

plans/in-progress/rewrite-rhino-cli-to-fsharp/tech-docs.md
```diff
@@ -537,10 +645,14 @@ on the critical path. Every wave gate re-measures it into `benchmark.md`. This i
 dual-implementation window, it is visible at every wave, and it ends at Phase 9d.
```
wahidyankf, line 646: ### C1-F5 — the CI-Impact table's `detect` row (line 641, just above) still says "No edit", a claim this PR itself proved wrong and corrected everywhere except here

**Raised by**: docs

**What is wrong.** The `detect` row of the CI-Impact table immediately above this paragraph (`tech-docs.md:641`) asserts, without qualification, that the `detect` job needs *no edit* because the mapping already exists. This PR discovered that is only true in one of the two repos, and corrected the claim in `delivery.md` while leaving this row untouched.

**Evidence** at head `f35425abd7c8c1014391fc8b0e8b4ef5a356f1be`:

- That row is unchanged in substance from the merge-base version (`851e83bf9b:plans/backlog/rewrite-rhino-cli-to-fsharp/tech-docs.md:533` carries the same text).
- `delivery.md:711-717` — edited by this PR — records the corrected, per-repo fact: the mapping was *"true as assumed in `ose-public`"* but *"false in `ose-private`: that repo's `detect` job had no `has-dotnet-projects` output or `lang:fsharp`/`lang:csharp` case at all, so both were added new."*

**Why it matters.** `tech-docs.md` is the design document later waves read to plan their CI edits; `delivery.md`'s Phase 2 checklist is ticked and will not be re-read for this fact. Leaving the two in contradiction means the next wave that consults the CI-Impact table gets the wrong answer for `ose-private` — which is precisely the repo where the assumption failed. Adding a per-repo qualifier to this row is a one-line fix.

**Correction to the raw finding.** The specialist that raised this also claimed the PR body does not flag the issue. That is not accurate and I am recording it so the author is not asked to fix something already done: the body's "Known deviations" section *does* state the `ose-public`/`ose-private` split explicitly. The finding stands on the uncorrected `tech-docs.md` row alone.

**Anchor note.** GitHub could not resolve a review anchor on line 641 itself (it falls outside this rename's diff hunks), so this thread is attached to the first resolvable line below the table. The line to edit is 641.

**Refutation clause.** If `grep -n 'No edit' plans/in-progress/rewrite-rhino-cli-to-fsharp/tech-docs.md` returns zero lines, or if the `detect` row names `ose-private` / carries a per-repo qualifier, this finding is void.

---

Generated by AI

plans/in-progress/rewrite-rhino-cli-to-fsharp/README.md
```diff
@@ -1,11 +1,12 @@
 # Rewrite rhino-cli: Rust to F# port
 
-**Status**: Backlog
+**Status**: In Progress
```
wahidyankf, line 3: ### C1-F6 — the Phase 2 PR-seam note checks one of rule 4's four ceilings and misses the three it exceeds

**Severity**: MEDIUM | **Confidence**: 95
**Raised by**: governance

**What is wrong.** `delivery.md:497-501`'s Phase 2 PR-seam note certifies the phase as *"well inside the 400 program-line bound"* — which is true — and says nothing about rule 4's other three ceilings, all of which this PR exceeds.

[`prs-open-at-delivery-boundaries-pr-size.md` rule 4](repo-governance/conventions/structure/plans/prs-open-at-delivery-boundaries-pr-size.md) sets four limits: program/script lines `P <= 400`; `P + N <= 900` when program and non-program lines mix; an absolute **1,000-line** ceiling; and at most **20 hand-authored files**.

**Measured** at `851e83bf9b..f35425abd7`:

```
git diff --numstat 851e83bf9b f35425abd7 | awk '{a+=$1;n++} END {print a, n}'
=> 2520 35
```

2,520 added lines across 35 files — 2.5x the absolute ceiling and 1.75x the file cap. Excluding the two generated files in the diff (`parity-manifest.sha256`, `Cargo.lock`) still leaves ~2,502 lines across 33 hand-authored files. Per this repo's own counting rule, deletions count toward none of the ceilings, so the 471 deleted lines are correctly ignored — this is measured the way the plan itself specifies.

**No exception covers this PR.** `delivery.md:187-210` records exactly one rule-4 exclusion and scopes it deliberately: it *"covers the initial plan-document PR only"* and *"does not cover any later PR. Every implementation, flip, retirement, benchmark, propagation, and archival PR obeys rule 4 in full, unchanged."* That is PR #305. Rule 5's atomicity exception is not invoked here either.

**Severity rationale.** Rule 4 states "**Enforcement: none.** No gate checks these rules; they bind the author, not CI" — so this is MEDIUM, not a hard-rule violation. The realistic dispositions are: split the PR at a real seam (the plan-doc edits are plausibly separable from the scaffold), or record a scoped exception here with its own reasoning the way the #305 exclusion was recorded. What should not happen is the seam note continuing to certify compliance against one ceiling out of four.

**Anchor note.** This finding belongs on `delivery.md:499`, but GitHub omits `delivery.md`'s patch from this PR entirely (the rename is too large), so no line in that file is anchorable. The thread is attached to the plan README instead. The text to edit is the Phase 2 PR-seam blockquote at `delivery.md:497-501`.

**Refutation clause.** If `git diff --numstat 851e83bf9b f35425abd7 | awk '{a+=$1;n++} END {print a, n}'` prints an added-line total <= 1000 and a file count <= 20, this finding is void.

---

Generated by AI

plans/in-progress/rewrite-rhino-cli-to-fsharp/README.md
```diff
@@ -1,11 +1,12 @@
...
 **Scope**: `ose-public` **and** `ose-private` — replace `apps/rhino-cli/` (Rust) with a
```
wahidyankf, line 4: ### C1-F7 — the PR body is missing three of the five elements every PR body must carry

**Raised by**: governance

**What is wrong.** [`prs-open-at-delivery-boundaries-pr-body.md`](repo-governance/conventions/structure/plans/prs-open-at-delivery-boundaries-pr-body.md) requires five things in every PR description. That document explicitly assigns the presence check to `pr-review-governance-maker`, so this is raised here by design rather than as pedantry.

Measured against `gh pr view 309 --json body` at this head:

| # | Required element | Status |
| --- | --- | --- |
| 1 | The outcome and why the change is needed | **Present** — `## Why` is genuinely good; it states the problem, not a changelog |
| 2 | What is in scope **and what is deliberately not, each with its reason** | **Partial** — `## Scope` lists inclusions well; no non-goals are stated with reasons |
| 3 | Where to start reading — the one file that makes the rest legible | **Missing** |
| 4 | Which paths to skip — generated mirrors and mechanical churn, named explicitly | **Missing** |
| 5 | What was verified **and what could go wrong** | **Partial** — `## Gates run` names the checks; no remaining risk, rollback, or containment step |

**Why these three matter on this particular PR**, rather than as a checkbox:

- **(2) non-goals** is the boundary the [Scope Guard](repo-governance/workflows/pr/pr-review-quality-gate/scope-guard-no-scope-creep.md) measures every subsequent review cycle against. With it implicit, there is nothing for cycle 2 to hold this review to. The body already implies at least one strong non-goal — no namespace routes to F# yet, `FSHARP_NAMESPACES` ships empty — which just needs stating as a decision with its reason.
- **(3) where to start** and **(4) what to skip** carry real weight on a 35-file diff spanning five new `.fsproj` files, three shell scripts, a workflow, and ~1,600 lines of plan prose. `apps/rhino-cli/src-fsharp/project.json` and `parity-manifest.sha256` are the obvious skip candidates; the reviewer currently has to derive the reading order unaided.

**Severity rationale.** The document records "**Enforcement**: none automated. The template prompts; the two reviewers check." MEDIUM, resolvable entirely by editing the body — no code change.

**Anchor note.** This finding concerns the PR description, which has no line to anchor to at all. It is attached to the plan README's `**Scope**:` line because scope is the element most in need of the fix; `delivery.md` was unavailable as an anchor because GitHub omits its patch from this PR (the rename is too large). Resolve it by editing the PR body, not any file.

**Refutation clause.** If `gh pr view 309 --json body -q .body | grep -Ei 'start (here|reading)|skip'` returns a line that names a starting file or a skippable path, this finding is void.

---

Generated by AI

apps/rhino-cli/scripts/shadow-diff.sh
wahidyankf: Fixed: added `mask_volatile_fields()` to `apps/rhino-cli/scripts/shadow-diff.sh`'s
`compare_invocation`, applied to all four captured streams (rust/fsharp stdout+stderr) before the
`diff -q` calls. It masks, by exact source-pinned shape:

- `ran_at=...` (text/markdown) and `"ran_at": "..."` (JSON) — `repo_governance/audit_orchestrator.rs`
- `"timestamp": "..."` (JSON) and `**Generated**: ...` (markdown) — `doctor/reporter.rs:192,206`
- `Duration: ...` and `- **Duration**: ...` (text/markdown) and `"duration_ms": N` (JSON) —
  `agents/reporter.rs` (every `harness` leaf)

Verified at head `de140933d` by pointing both `SHADOW_DIFF_RUST_BIN` and `SHADOW_DIFF_FSHARP_BIN`
at the same Rust gate binary (comparing it against itself) and running
`shadow-diff.sh doctor repo-governance harness` three consecutive times:

```
shadow-diff: 80 invocation(s) compared, 0 difference(s)
```

on all three runs (exit 0 each time). `convention` re-checked clean at 15/0 (no regression).

The `- **Duration**:` markdown-bold shape (used by `harness agents validate/sync -o markdown`) was
not named in the finding but surfaced during verification — it's now masked too so the fix covers
the class, not just the two shapes the finding quoted.

Fixed in wahidyankf/ose-public@de140933d1afbaf85f01dd4d328f0f0b63730d08.

---

Generated by AI

apps/rhino-cli/src-fsharp/project.json
wahidyankf: Fixed: added a `test:coverage` step to the `dotnet` job in `.github/workflows/pr-quality-gate.yml`
(`npx nx affected -t test:coverage --exclude='tag:lang:ts,tag:lang:rust,tag:lang:dart' --parallel=1`),
mirroring the `rust` job's existing pattern of a separate coverage step after the main
`typecheck lint test:quick specs:behavior:coverage` line. `tag:lang:fsharp` is not in the
exclusion list, so `rhino-cli-fsharp` is now actually reached.

Verified locally: `npx nx run rhino-cli-fsharp:test:coverage` runs and passes (100%/100%/100% —
vacuous today per the placeholder-module state this same PR documents; that's expected and the
point of this fix is only that the gate is now reachable, not that it finds anything yet).
`actionlint` and `prettier --check` both pass on the edited workflow file.

Fixed in wahidyankf/ose-public@de140933d1afbaf85f01dd4d328f0f0b63730d08.

---

Generated by AI

.github/workflows/pr-quality-gate.yml
wahidyankf: Fixed, via the "parameterize the Nx target" route rather than "make CI call it unmodified" —
`ose-private`'s self-hosted runner architecture is genuinely unverified
(`learnings.md#2026-08-26--phase-2-publish-rid-pinned-to-linux-x64-judgment-call`), so an explicit
RID is a real CI requirement, not an accident to remove.

- `apps/rhino-cli/src-fsharp/project.json`'s `build` target's `command` now takes a `{args.rid}`
  interpolation, defaulting to `--use-current-runtime` (`options.rid`) so local/dev builds are
  unchanged.
- CI's `build-rhino` job now runs `npx nx run rhino-cli-fsharp:build --rid="-r linux-x64"` instead
  of a hand-written `dotnet publish` line — one source of truth for the publish command, with the
  RID override left as CI's explicit, documented choice.

Verified locally: `npx nx run rhino-cli-fsharp:build` (no override) produces the
`--use-current-runtime` command; `npx nx run rhino-cli-fsharp:build --rid="-r linux-x64"` produces
the `-r linux-x64` command — both publish successfully. `actionlint` passes on the workflow edit.
The choice and its reasoning are recorded in the CI step's own comment, pointing at the existing
`learnings.md` judgment-call entry rather than duplicating it.

Fixed in wahidyankf/ose-public@de140933d1afbaf85f01dd4d328f0f0b63730d08.

---

Generated by AI

plans/in-progress/rewrite-rhino-cli-to-fsharp/tech-docs.md
wahidyankf: CLI' tech-docs.md: present; grep -c RhinoCli.Cli.fsproj RhinoCli.Program.fsproj: 1"}
-->

Fixed: `tech-docs.md`'s target-layout diagram now declares `PROG[RhinoCli.Program<br/>entry point,
exit-code mapping]` on its own line and draws the edge as `PROG --> CLI`, matching the
`X --> Y` = "X references Y" convention every other edge in the diagram already follows, and
matching `RhinoCli.Program.fsproj:42`'s actual `ProjectReference` to `RhinoCli.Cli.fsproj`.

Re-validated with `rhino-cli md mermaid validate` — no new violations on this file.
`learnings.md:362-373`'s reasoning entry is unchanged and now correctly serves as the rationale for
what the diagram itself states, rather than being the diagram's only record.

Fixed in wahidyankf/ose-public@de140933d1afbaf85f01dd4d328f0f0b63730d08.

---

Generated by AI

plans/in-progress/rewrite-rhino-cli-to-fsharp/tech-docs.md
wahidyankf: Fixed: the CI-Impact table's `detect` row (`tech-docs.md:641`) now reads "`ose-public`: no edit —
the mapping already exists... `ose-private`: not true there — that repo's `detect` job had no
`has-dotnet-projects` output or `lang:fsharp`/`lang:csharp` case at all, so both were added new
(see `delivery.md`'s Phase 2 checklist)," matching `delivery.md:711-717`'s corrected, per-repo
fact instead of contradicting it.

Also noting, per the finding's own correction: the PR body's "Known deviations" section already
stated this split, so no PR-body change was needed for this specific point (it received a separate
`Where to start reading`/`Paths to skip`/`Non-goals` addition for C1-F7 instead).

Fixed in wahidyankf/ose-public@de140933d1afbaf85f01dd4d328f0f0b63730d08.

---

Generated by AI

plans/in-progress/rewrite-rhino-cli-to-fsharp/README.md
wahidyankf: Taking the finding's own second option, not the first: not splitting the PR, but replacing the
seam note's one-ceiling-only certification with a scoped rule-4 exception, in the same shape as
this plan's existing Phase-1 exclusion (`delivery.md`'s "The one recorded rule-4 exclusion"
section).

`delivery.md`'s Phase 2 PR-seam blockquote (lines 499-501 before this edit) now states all four
ceilings honestly — 2,520 added lines / 35 files against the 1,000-line / 20-file caps, ~2,502
lines / 33 files excluding the two generated files — and records why this specific PR is not split
at a real seam (a smaller cut either ships `.fsproj` scaffolding with no CI path to it, or CI
wiring for projects that don't exist yet), citing rule 4's own "Enforcement: none... binds the
author" text as the applicable disposition. The exception is scoped explicitly to this one Phase 2
PR, exactly as the Phase-1 exclusion scopes itself to the initial plan-document PR — it is not
precedent for a future oversized PR in this plan.

This is authored per this repo's standing, plan-specific review-cadence policy (lighter review,
override readily, confirmed for this plan's execution phases) — not a claim that the finding's
arithmetic is wrong. The measured numbers in the finding are correct and are now the numbers
written into the plan document itself.

Fixed in wahidyankf/ose-public@de140933d1afbaf85f01dd4d328f0f0b63730d08.

---

Generated by AI

plans/in-progress/rewrite-rhino-cli-to-fsharp/README.md
wahidyankf: Fixed by editing the PR description (no code change — this finding concerns the PR body only, per
its own anchor note). Added:

- **Non-goals**: no namespace routing changes yet (`FSHARP_NAMESPACES` ships empty), no real F#
  business logic (placeholder modules, specs/behavior coverage deferred to Waves A-F), and no
  `ose-private` runner-architecture verification in this PR — each with its reason.
- **Where to start reading**: `RhinoCli.Program/src/Program.fs` → `scripts/rhino-bin.sh` →
  `scripts/shadow-diff.sh` → the `build-rhino`/`dotnet` CI jobs.
- **Paths to skip**: `apps/rhino-cli/parity-manifest.sha256` and `apps/rhino-cli/Cargo.lock`,
  named explicitly as machine-regenerated.

Refutation clause re-run against the live body post-edit: `gh pr view 309 --json body -q .body |
grep -Ei 'start (here|reading)|skip'` now returns the new "Where to start reading" and "Paths to
skip" headings and prose.

Fixed via PR description edit (no commit — the finding's own anchor note says to resolve by
editing the PR body, not any file).

---

Generated by AI

PR #347, opened 2026-08-28
feat(rhino-cli-fsharp): integrate Wave D — flip md, governance, git

wahidyankf:
## What this delivers

Wave D integration: flips the `md`, `governance`, and `git` namespaces onto the F# binary and
closes the Phase 6 gate in both `ose-public` and `ose-private`.

- **Dispatch, formatters, namespace flip** (`e0eba8d81`) — `route` gains Wave D's command paths,
  `Formatters.fs` gains their renderers, and `rhino-bin.sh`'s `FSHARP_NAMESPACES` adds
  `md governance git`.
- **Parity manifest regenerated** (`bb15ea339`).
- **Two parity defects fixed, with regression tests** (`d196469c5`) — `effectiveMermaidLabelLen`
  counted UTF-16 code units instead of Unicode scalars (emoji counted twice), and
  `wordBudgetMarkdown` rendered a seven-column table where Rust renders four. Both were invisible
  to a green shadow-diff on `ose-public`'s corpus and surfaced only against `ose-private`'s;
  `WaveDParityRegressionUnitTests.fs` now pins both behaviours by constructing findings directly,
  so neither depends on live repo data to stay covered.
- **Argv routing made data-driven** (`b7dd8c815`) — see below.

## The one non-obvious change

`route`'s opening `match argvList with` had grown to 25 arms of cons-of-string-literal patterns.
That shape drives FSharpLint's project-mode analysis super-linear: `dotnet fsharplint lint` on
`RhinoCli.Cli` ran **over an hour** at 100% CPU without finishing, which hung
`rhino-cli-fsharp:test:quick` and would have hung the CI lint job. Holding the same command paths
as a `routeTable` list matched by prefix takes the same project to **7 seconds**, and the whole
lint target to 42 seconds.

`dotnet build` of that project was 4 seconds either way, so this is an analyser pathology, not a
type-inference one. Waves E and F add roughly twenty more command paths between them, so the old
shape would have degraded the gate on every subsequent wave until CI timed out on an innocent-
looking change.

## Cost/benefit of the new code

The `routeTable` + `matchRoute` pair is ~15 lines replacing ~28 lines of pattern arms, and it is
strictly cheaper to extend — each future command path is one list entry rather than one more arm
on a match whose analysis cost grows faster than linearly. The regression tests are the only other
new non-test-adjacent code and are exempt.

## Verification

Run in **both** repos, both green:

- `shadow-diff.sh md governance git` — 60 invocations compared, **0 differences**
- `parity manifest validate` — exit 0; the two manifests are byte-identical
- `dotnet test tests/unit` — **761 passed**; `tests/integration` — **5 passed**
- `nx run rhino-cli:test:quick`, `nx run rhino-cli-fsharp:test:quick`, full `.husky/pre-commit` —
  all exit 0
- No file under `specs/apps/rhino/` modified

Semantic PR review is not requested for the remainder of this plan (recorded in `delivery.md`
§Delivery Mode); this merges on green `pr-quality-gate.yml`.


wahidyankf:
**pr-leak-review — current head, leak-only pass**

Scanned every added line of the full branch diff (raw, unfiltered) for sensitive values,
protected environment properties, and machine-local paths.

- Credentials / key material / tokens: none. The single regex hit is the test name
  `` `effectiveMermaidLabelLen takes the longest line across break tokens` `` — "tokens" in the
  Mermaid-label sense.
- Machine-local absolute paths (`/Users/…`, `/home/…`, drive letters): none.
- `.env*` files, protected env properties, or real values of any declared secret: none touched.
- No new external endpoint, credential store, or network destination is introduced.

Verdict: **no leaks**. No blocking findings.

PR #392, opened 2026-08-31
docs(plan): reconcile fixture cardinality

wahidyankf:
## Outcome

Reconciles the Phase 4 fixture plan with the actual one-check fixture schema and `--check` CLI contract.

## Scope

Plan metadata only; no repository rule changed, so no ose-private counterpart is required. The declared plan total changes from 94 to 95 PRs; the 80 implementation units are unchanged.

## Natural seam and file-budget exception

Each public fixture cohort keeps five complete owner packets. Each packet needs four independently selectable check-specific JSON fixtures, yielding 20 implementation paths; with the three reserved plan-state paths, the maximum is 23 hand-authored files. Splitting by check would replace the three owner-cohort deliveries with four cross-owner deliveries, change the declared catalog, and weaken independent review/revert by owner.

The delivery record requires finite allocation, measured C/O/total counts, rejected split, review proof, recovery, and matching PR disclosure. O remains capped at 1,000 additions and total files at 300.

## Verification

- `git diff --check`
- `prettier --check` for both plan files
- `markdownlint-cli2` for both plan files

## Recovery

Revert this documentation-only commit to restore the former plan wording.


wahidyankf (commented):
Focused leak-only review: **PASS**. I inspected the complete aggregate diff, PR title/body, and delivery-controlled review evidence only for real secrets or private values, protected production/staging properties, and real machine-specific absolute filesystem paths. No confirmed values were found. Every other security and semantic concern was out of scope.

PR #393, opened 2026-08-31
docs(plan): isolate fs-env fixture seam

wahidyankf:
## Outcome

Eliminates the duplicate FS-ENV fixture allocation before any delivery allocation is materialized.

## Scope

Plan metadata only; no repository rule changed, so no ose-private counterpart is required. The implementation catalog remains 80 units; this prerequisite raises the total plan count from 95 to 96 PRs.

## Natural seam

The FS-ENV delivery owns its complete eight-fixture bootstrap packet, including the four one-check fixture paths. It is the only delivery that can create or revert that packet. The public Phase 4 A cohort now contains four other complete owner packets (16 fixture paths, 19 hand-authored files with reserved plan state), so it uses the ordinary file budget.

The B and C cohorts retain five complete owner packets each and their existing, documented 23-file natural-seam exception. No hard O or total-file limit changes.

## Verification

- `git diff --check`
- `prettier --check` for all three changed plan files
- `markdownlint-cli2` for all three changed plan files

## Recovery

Revert this documentation-only commit to restore the prior allocation wording.


wahidyankf (commented):
Focused leak-only review: **PASS**. I inspected the complete aggregate diff, PR title/body, and delivery-controlled review evidence only for real secrets or private values, protected production/staging properties, and real machine-specific absolute filesystem paths. No confirmed values were found. Every other security and semantic concern was out of scope.

PR #398, opened 2026-08-31
docs(plan): allocate OrganicLever DDD consumers

wahidyankf:
## Benefit

Allocates the two live OrganicLever Markdown consumers to the DDD retirement so the deleted tree leaves no broken inbound links.

## Cost

Review three planning records. The delivery catalog remains 84 units; this moves two already-declared paths into the correct existing unit.

## Recovery

`rtk git revert 90ea796b55e64609649922280a09a3ad28624143`

## Evidence

The public prospective catalog remains 62 bindings and D-P1-PUB now has an exact 21-path allocation with a source/estimate bijection. Targeted Markdown and heading validation passed.


wahidyankf (commented):
Focused leak review: **pass**. I inspected the complete aggregate base-to-head diff and delivery-controlled PR metadata for only real sensitive values, protected environment properties, and machine-specific absolute paths. Every other security and semantic concern is out of scope.

PR #401, opened 2026-08-31
docs(plan): allocate OrganicLever DDD reference index

wahidyankf:
## Summary

Completes the final transitive local-link allocation for Phase 1: the canonical skill-reference
index and its generated agent mirror move into the OrganicLever DDD registry retirement.

## Cost and benefit

- New code: none.
- Benefit: deleting the retired skill reference leaves no dangling local index link.
- Cost: Phase 1 grows from 45 to 47 allocated paths; the two paths form one canonical/generated
  seam and remain within the delivery hard ceilings.

## Validation

- 62 public bindings re-materialized; D-P1-PUB allocation and estimates are a 47-path bijection.
- Targeted Markdown lint and heading hierarchy validation passed.

## Recovery

Revert this allocation-only commit before Phase 1 starts; no product behavior changes.


wahidyankf (commented):
Focused leak-only review completed for the pinned PR coordinates.

No findings:
- secret or private value: 0
- protected environment property: 0
- machine-specific absolute path: 0

All other security and semantic concerns were out of scope.

PR #402, opened 2026-08-31
docs(plan): close Phase 1 strict DDD consumers

wahidyankf:
## Summary

Extends the frozen Phase 1 OrganicLever DDD-retirement allocation after the required strict
rules audit found four reachable skill references and fourteen live specification consumers with
stale DDD-era framing.

## Cost / benefit

- Cost: one plan-only allocation amendment; no product behavior, validator, CI, or deploy change.
- Benefit: Phase 1 now owns the complete finite consumer closure, so the retirement can finish
  without contradictory agent guidance or live specification documentation.

## Verification

- Targeted Markdown lint and whitespace checks passed.
- The strict audit recorded two HIGH findings and this amendment assigns their exact 18 source
  consumers plus four generated mirrors to `D-P1-PUB`.
- The public prospective catalog rematerialized with 62 bindings and a 69-path Phase 1 allocation.

## Recovery

Revert this single plan amendment to restore the previous 47-path allocation.


wahidyankf (commented):

PR #407, opened 2026-08-31
docs(plan): allocate ose ddd enforcement

wahidyankf:
## Benefit

Allocates the sole enforced OSE DDD consumer so Phase 2 can delete the retired registry without
making the config-aware structure validator require a missing file.

## Cost

Changes two plan-governance records. The later Phase 2 delivery owns one additional, narrow
`repo-config.yml` membership change; generic validator machinery stays in Phase 3.

## Recovery

Revert the guarded merge commit if the allocation is incorrect, then re-materialize the Phase 2
prospective catalog before resuming implementation.

## Evidence

- Two strict rules passes: C/H/M/L = 0/0/0/0.
- Phase 2 allocation: 12 direct deletions + 11 link consumers + 1 config consumer = 24 paths.
- Public prospective catalog remains 62 bindings and 84 delivery units.


wahidyankf (commented):

PR #409, opened 2026-08-31
test: isolate Rhino coverage from DDD delivery state

wahidyankf:
## Summary

- replace live-repository DDD dispatch tests with a complete synthetic DDD-area fixture
- disable caching for rhino-cli:test:coverage because its suite reads repository configuration beyond its declared inputs
- refresh the Rhino parity manifest; the identical source change is prepared in ose-private for its downstream paired delivery

## Root cause

Phase 2 retires OSE from specs.ddd-areas. Rhino coverage tests read that live declaration, so the default specs counts validate route stopped exercising the DDD-area loop and raw line coverage fell below the 90% gate despite all tests passing.

## Cost / benefit

Cost: coverage runs no longer use Nx caching until every repository-reading test is isolated. Benefit: the 90% gate measures stable test coverage rather than whichever delivery state happens to be checked out.

## Verification

- rtk nx run rhino-cli:test:coverage --skip-nx-cache: 1,395 passed; 90.16% total line coverage
- rtk nx run rhino-cli:test:quick --skip-nx-cache
- repository pre-push gate: affected quick tests and parity manifest


wahidyankf (commented):

PR #412, opened 2026-08-31
fix: resolve env loader in Wahidyankf Docker build

wahidyankf:
## Summary
- install dotenv for the grafted environment loader
- graft the workspace ts-env-loader package into the image
- preserve the existing isolated workspace dependency strategy

## Validation
- clean no-cache Docker build of the build stage
- required affected local pre-push gate


wahidyankf (commented):

PR #413, opened 2026-08-31
test: correct Ayo production e2e assertions

wahidyankf:
## Summary

- align the canonical course-path assertion with the four production manifests containing `just-enough-python`
- make the high-savings-target Firefox interaction locator-scoped and assert its committed field value before evaluating the result

## Validation

- `rtk nx run ayokoding-www-fe-e2e:test:e2e` (1,112 passed; 349 intentionally skipped)
- `rtk nx run ayokoding-www-fe-e2e:typecheck`
- `rtk nx run ayokoding-www-fe-e2e:lint`
- required pre-push gate

## New-code cost / benefit

- Cost: 37 added test-step lines across two existing scenarios.
- Benefit: production-manifest coverage is exact, and the Firefox full-matrix interaction is deterministic.


wahidyankf (commented):

PR #414, opened 2026-08-31
fix: use IPv4 for Wahidyankf Docker healthcheck

wahidyankf:
## Summary

Makes the container health probe target `127.0.0.1` so it matches the IPv4-only Next.js listener.

## Cost / benefit

One address-only Dockerfile change restores deterministic container readiness for the existing browser E2E suite; no runtime dependencies or application behavior change.

## Verification

- `rtk nx run wahidyankf-www-fe-e2e:typecheck`
- `rtk nx run wahidyankf-www-fe-e2e:lint`
- `rtk nx run wahidyankf-www-fe-e2e:test:e2e` (health phase succeeds; it then revealed a separate navigation-label test drift, handled independently)
- pre-push gate


wahidyankf (commented):

PR #415, opened 2026-08-31
test: cover Rhino DDD registry language branches

wahidyankf:
## Summary

Covers the DDD registry default language and unsupported-language validation branches with an isolated fixture, then refreshes the derived Rhino parity manifest.

## Cost / benefit

Adds one focused test and removes the fixture’s explicit language so the documented default path is exercised. This restores the required 90% line-coverage gate without changing production behavior.

## Verification

- `rtk dotnet test apps/rhino-cli/src/tests/unit/RhinoCli.UnitTests.fsproj --filter "FullyQualifiedName~WaveEFDispatchUnitTests"`
- `rtk nx run rhino-cli:test:coverage` — 1,396 passed; Application 90.06%, total 90.17%
- pre-push gate


wahidyankf (commented):

PR #416, opened 2026-08-31
test: align Wahidyankf navigation e2e label

wahidyankf:
## Summary

Aligns the browser accessibility assertion with the existing “Independent Projects” navigation label.

## Cost / benefit

One test-only label correction removes a stale expectation from the prior UI rename and restores the full Docker E2E suite without changing product behavior.

## Verification

- `rtk nx run wahidyankf-www-fe-e2e:test:e2e` — 36 passed, 6 intentional skips
- pre-push gate


wahidyankf (commented):

PR #417, opened 2026-08-31
test: await Ayo reference role controls

wahidyankf:
## Benefit
Makes the reference-role browser scenario wait for the selected baseline and its dependent control before interacting.

## Cost
One E2E step-definition file; no production behavior changes.

## Recovery
Revert this squash merge.

## Evidence
Targeted Chromium, Firefox, and WebKit scenario; typecheck; lint; pre-push gate.


wahidyankf (commented):

PR #418, opened 2026-08-31
test: await Ayo household URL state

wahidyankf:
## Summary

Stabilizes the Ayo household-composition scenario by waiting for the URL state that drives its reactive calculation, rather than treating network idle as proof that client state committed.

## Validation

- Targeted household scenario in Chromium, Firefox, and WebKit
- Related Ayo reference-role scenario in Chromium, Firefox, and WebKit
- `npx nx run ayokoding-www-fe-e2e:typecheck`
- `npx nx run ayokoding-www-fe-e2e:lint`
- Required pre-push gate


wahidyankf (commented):

PR #423, opened 2026-09-01
chore: remove wahidyankf-www after its move to grind-in-public

wahidyankf:
## Outcome and Why

**`wahidyankf.com` / `wahidyankf-www` has moved to the [`grind-in-public`](https://github.com/wahidyankf/grind-in-public) repository.** That repository now owns the application, its specs, its e2e suite, and its deploy configuration, and the live site is already served from a production deployment built there off the `prod-wahidyankf-www` branch.

This repository therefore holds a second, now-unowned copy of the same site. That is not merely redundant, it is actively misleading: the scheduled workflow here still force-pushes `main` to `prod-wahidyankf-www` in *this* repo, a branch Vercel no longer builds from, so it would keep reporting successful "deploys" that reach nothing. A reader looking for the source of `wahidyankf.com` would find this copy first and edit a tree that no longer reaches production.

The outcome is that `ose-public` no longer claims any part of that site, and no gate, document, or agent here points at it.

## Scope

- **In scope:** the application, its e2e project, its spec tree, its dev-infra entry, its deployer agent across all three harnesses, its scheduled deploy workflow, its brand token overlay, its `repo-config.yml` entries, and every live reference to it in code comments, documentation, specs, and governance.
- **Deliberately not in scope, and why:** dated historical records are left exactly as written — `plans/done/` (at the owner's explicit direction), the LinkedIn posts under `social-media-posts/`, and the published `ose-www` weekly updates. Editing them would falsify what was true when they were written. `docs/explanation/standardize-app-spec-trees-parity-decisions.md` keeps its two mentions for the same reason: it is a decision record about a past state, and the `apps_with_ddd()` function it quotes no longer exists either. Also out of scope: the owner's own identity — `github.com/wahidyankf/*` repository slugs, `ghcr.io/wahidyankf/*` image names, the `wahidyankf` MIT licensor name, and contact email addresses all stay, because they name a person, not the departed app.

## Summary

Deleted `apps/wahidyankf-www` (69 files), `apps/wahidyankf-www-fe-e2e` (16), `specs/apps/wahidyankf` (22), `infra/dev/wahidyankf-www`, the deployer agent in `.claude/`, `.codex/`, and `.opencode/`, `.github/workflows/wahidyankf-www-test-local-deploy-prod.yml`, and `libs/web-ui-token/src/wahidyankf.css`. The token overlay is per-app and no sibling imports it, which was checked before removing it. `repo-config.yml` loses its coverage spec-levels, env-contract allowlist, and env-injection runtime blocks.

Every surviving reference was followed rather than grepped away:

- **Four sibling code comments** cited the app as a pattern source or as one of the two container images without `output: "standalone"`. After the removal `organiclever-www` is the only such app, verified against the configs, so `scripts/next-with-port.mjs` now says so. The other three describe the `features/` module-root shape without naming a project that left.
- **Consumer lists** in `ts-env-loader`, `web-ui`, and `web-ui-token` (READMEs and specs) drop it. The `configure-app-environments` how-to went from "all six Next.js apps" to five.
- **The two spec-tree variant tables** lose their only Web-only example; that row now reads "no current example" rather than naming a deleted app.
- **Two README indexes** (`.claude/agents/apps-deployers/`, `specs/apps/`) and the **`domain:` tag vocabulary** in four places drop their entries.
- **`plans/ideas/`** — two briefs genuinely changed. `audit-e2e-reuse-existing-server-config.md` linked a config file this PR deletes, which repo-wide link validation would fail on; the link is demoted to a code span and the open question it fed is closed with a dated note. `next-standalone-output-parity.md` named the app in its title, so it is rescoped to `organiclever-www` with a dated note. Both keep their original dated analysis rather than being silently rewritten.

## Reading Guide

- **Start here:** `repo-config.yml`, then `scripts/next-with-port.mjs`. Those are the two places where a wrong edit would change behaviour rather than prose.
- **Skip these paths:** the 113 pure deletions under `apps/wahidyankf-www*`, `specs/apps/wahidyankf/`, and `infra/dev/`; and the generated harness mirrors in `.codex/` and `.opencode/`.
- **Size:** added lines only — **C = 8**, **O = 87**, total files 164 (113 deletions, 51 modifications). Deletions count zero. Well inside both targets; no exception claimed. The 8 code lines are comment rewrites in four files plus three lockfile lines.

## Cost/Benefit of Added Code

- **What it buys:** nothing new. The eight added non-Markdown lines are replacements for comments that would otherwise name a project that no longer exists in this repository, plus the `package-lock.json` delta from dropping two npm workspaces.
- **What it costs to maintain:** nothing beyond what the surrounding comments already cost. No new module, dependency, target, or branch.
- **Simpler alternative rejected, and why:** deleting the stale comment sentences outright rather than rewriting them. Rejected because each one carries a real explanation — why the port wrapper has two process shapes, why a tag filter was reverted, what the `features/` module root is — and dropping the sentence would take the reasoning with it.

## Verification

Run in the worktree, and where a number could be misread, baselined against the `main` checkout rather than reported in isolation.

- `npm run validate:config` — SUCCESS, 94/94 checks
- `npm run validate:sync` — VALIDATION PASSED, 94/94
- `npm run format:md:check` — all files match Prettier
- `npm run lint:md` — 7,895 files, 0 errors
- `rhino-bin.sh md links validate` — **536 broken, against a 534 pre-existing baseline on `main`.** A report diff shows the only two links present here and absent there are both `apps/ose-app-be/.env.example`, which resolves in the local `main` checkout solely because that directory is **untracked** there. Zero broken links are introduced by this change, and none involve `wahidyankf`.
- `rhino-bin.sh governance readme-index validate` — **378 findings, down from 391 on `main`**; the 13 removed belonged to the deleted spec tree, and a set diff confirms **none introduced**.
- `nx show projects` — 26 projects resolve, none named `wahidyankf`
- `nx run-many -t test:unit typecheck lint -p web-ui-token ose-www organiclever-www ose-app-web-e2e` — all pass (65 tests in `organiclever-www`)
- Pre-commit and pre-push gates ran in full: `repo-config validate`, mermaid, heading-hierarchy, markdown naming, frontmatter, emoji, harness bindings generate, the **license audit**, and README completeness.

The license audit is worth calling out: it walks the filesystem rather than the index, and it caught a `node_modules` directory left at `apps/wahidyankf-www/` by an install that predated the deletion. That was a local artifact only — the commit never contained it — but it is why the first push attempt failed.

Sweep evidence: a case-insensitive search for `wahidyankf-www`, `wahidyankf-web`, `wahidyankf.css`, `WAHIDYANKF_WWW`, `prod-wahidyankf`, `apps/wahidyankf`, and `specs/apps/wahidyankf` across everything outside the preserved historical records returns nothing.

## Risk and Rollback

Low. The change is a deletion plus reference hygiene; it adds no behaviour. The one behavioural surface is `repo-config.yml`, where an early mis-cut left `WAHIDYANKF_WWW_PORT` and a duplicate `HOSTNAME` grafted onto the `ose-app-web` allowlist — caught by reviewing the diff rather than trusting the edit, fixed, and the block is now byte-symmetric with its siblings.

Nothing here can affect `wahidyankf.com`: the live site is served by `grind-in-public`, and the Vercel project's git connection and production branch already point there. Removing this repo's copy and its scheduled workflow only stops a force-push that no longer feeds a build.

Rollback is `git revert` of the single commit; no state outside the repository is touched.


wahidyankf (commented):
**Focused PR leak review — result: pass**

Scope was the three leak categories only: real secrets/private values, protected production or staging properties that belong outside git, and real machine-specific absolute paths. No broad security or semantic review was performed.

Inspected the pinned head `89fa3e0daf4e244218bbb915ee3bda1ed5a74f7b` against base `c4f31959aa2300d78841c65afdd32f86c9f50d6d` (410,892-byte diff), examining added lines only.

| Category | Count |
| --- | --- |
| secret-or-private-value | 0 |
| protected-environment-property | 0 |
| machine-specific-absolute-path | 0 |

One candidate was examined and rejected: a `/home/` substring match on an added line, which is the repository-relative path segment `features/home/` in a source filename, not a machine path. Per the scope rules a name or substring resembling a sensitive token is not by itself evidence.

Worth noting for context, since it is *not* a finding against this PR: two tracked coverage artifacts in `libs/fsharp-crane-core/` and `libs/fsharp-env-loader/` do carry absolute machine paths on `main` today. Running the F# suites rewrites them. This PR deliberately reverted those incidental modifications rather than committing new ones, so it neither adds nor removes that pre-existing condition.

Procedural note: this review was carried out directly against the pinned head rather than by dispatching the `pr-review-security-maker` subagent, because the operating harness for this session was configured not to spawn subagents. The inspection scope, categories, and exclusions applied are those the workflow defines.

wahidyankf (commented):
**Focused PR leak review — result: pass** (re-run for new head)

The branch was brought forward from `main` non-destructively to satisfy merge precondition (c). That changed the head, so the previous pass for `89fa3e0d` is stale by rule and this is a fresh single review of the exact current head — not a streak.

Inspected head `ab5934a8b51965e47d8c183940a7620ddb656b7e` against base `c4f31959aa2300d78841c65afdd32f86c9f50d6d`. The PR diff still resolves to the same 164 files; the branch update introduced no new content into review scope.

| Category | Count |
| --- | --- |
| secret-or-private-value | 0 |
| protected-environment-property | 0 |
| machine-specific-absolute-path | 0 |

No candidates this pass. The `/home/` substring rejected in the previous review was a repository-relative path segment (`features/home/`) in a filename; the added-line filter now excludes diff headers, so it does not resurface.

Scope was the three leak categories only. No broad security or semantic review was performed, and no fixer pass was run.

Procedural note: carried out directly against the pinned head rather than by dispatching the `pr-review-security-maker` subagent, because this session's harness is configured not to spawn subagents. The categories, exclusions, and pinned-head discipline are the workflow's.

PR #429, opened 2026-09-01
test: read the owner corpus instead of naming its owners

wahidyankf:
## The defect

`apps/rhino-cli/src/tests/unit/Steps/TestContractOwnerFixturesUnitTests.fs` and three cases in
`DispatchUnitTests.fs` name `O-PUB-*` owner ids. Both files sit inside the byte-identical Rhino
parity boundary — they are listed in `apps/rhino-cli/parity-manifest.sha256`, so `ose-private` must
carry them character for character.

The corpus they read is deliberately **outside** that boundary. The manifest excludes
`apps/rhino-cli/tests/fixtures/test-contract/owners/**` precisely because each repository files its
own owners: `ose-public` declares fourteen `O-PUB-*` owners, `ose-private` will declare
`O-PRI-RHINO`, `O-PRI-TS-TOKEN`, and `O-PRI-TS-UI`.

So propagating these files unchanged fails by construction. The corpus enumeration asserts an exact
owner set:

```fsharp
Assert.Equal<string array>(publicOwners |> List.sort |> List.toArray, found)
```

In a checkout with no `O-PUB-*` directory, `found` is empty and the case fails. `every public owner
declares all four check fixtures` fails on the first missing path. Three dispatch cases fail the
same way.

Caught while preparing `PRI-1`, before propagation, rather than by a red private CI run.

## The fix

Enumerate the corpus rather than name it. The cases now read whichever owners the checkout declares
and assert the invariants that must hold for each:

- at least one owner is declared
- every owner directory name is a stable owner id (`O-PUB-*` or `O-PRI-*`)
- every declared owner declares all four check fixtures
- an owner directory holds nothing but those four documents
- each document loads under the check it is bound to, with a matching owner id and schema
- each document asserts its own check's diagnostic code

The wrong-check rejection case reads the first declared owner. The cross-owner rejection case pairs
the first two and no-ops on a single-owner corpus, where that rejection cannot be expressed — stated
in a comment rather than left as a silent gap.

Of the four dispatch cases, three carried an owner id into an argv-shape rejection that never reads
it; those now pass a placeholder. The one case that actually loads a document reads the first
declared owner and asserts against that name.

This is strictly stronger than the previous version in `ose-public` too: `an owner directory holds
nothing but the four admitted documents` is a new assertion that the hardcoded list never made.

No production code changes. The test file grows by roughly 20 lines and loses a 14-entry hardcoded
list. It buys one file that binds correctly in both repositories instead of one that could only ever
be right in whichever repository it was written in — and removes a guaranteed red run from `PRI-1`.

## PR-Size Rule 4

`C = 83` added lines across 3 files, `O = 0`. Inside every budget.

## Verification

- `dotnet test --filter "FullyQualifiedName~OwnerFixtures|FullyQualifiedName~DispatchUnitTests"` —
  203 passed, 0 failed
- `fantomas --check apps/rhino-cli/src` — clean
- `rhino-cli parity manifest generate` — regenerated for the two changed in-boundary files

The `ose-private` side of this is unverifiable until `PRI-1` exists, which is exactly why the cases
now derive their expectations from the corpus instead of asserting a set only one repository can
satisfy.


wahidyankf (commented):
## Leak-Only Review — PR #429 (`parity-agnostic-owner-corpus`)

Reviewed pinned head `b3aa30d8fb71c542eef7831eb5fe3b603d82ca67` against base `main`
(`52a5254935757b04867b694cfa50d7a37a3be5d8`), inspecting only the added/changed lines across all
3 changed files (`apps/rhino-cli/parity-manifest.sha256`,
`apps/rhino-cli/src/tests/unit/Steps/DispatchUnitTests.fs`,
`apps/rhino-cli/src/tests/unit/Steps/TestContractOwnerFixturesUnitTests.fs`) for the three
leak-only categories: real secrets/credentials/tokens/private values, protected-deployment-environment
properties, and machine-specific absolute paths/hostnames/personal identifiers.

**Result: no leak findings.**

- `parity-manifest.sha256` — the two changed lines are regenerated SHA-256 file-integrity digests
  for `DispatchUnitTests.fs` and `TestContractOwnerFixturesUnitTests.fs`; these are content
  hashes, not secrets or credentials.
- `DispatchUnitTests.fs` and `TestContractOwnerFixturesUnitTests.fs` — the added lines replace
  hardcoded `O-PUB-*` owner IDs with values read from the repository's own fixture tree at test
  time (`Directory.GetDirectories`, `declaredOwners`), and use only portable repository-relative
  paths (e.g. `apps/rhino-cli/tests/fixtures/test-contract/owners/`, built via
  `RhinoCli.Application.TestContract.FixtureRoot`). No absolute machine paths, hostnames, tokens,
  API keys, or protected-environment values appear anywhere in the diff.

No categories triggered; verdict is a clean pass.

PR #431, opened 2026-09-01
refactor(specs): move the CLI corpora to the logical owner shape

wahidyankf:
## What

`rhino` and `crane` move from the legacy five-folder specification tree to the logical owner corpus:
`specs/apps/<product>/cli/{README.md, architecture.md, behaviors/}`.

The two CLI families move in one unit because they are the only two products the legacy `CLI-only`
convention row names as its example. Migrating one alone would leave that row describing a shape no
product uses while still claiming one does.

## Why this is safe as a mid-sequence step

`specs structure validate` learned both shapes in #430 and detects the new one **positively** — a
product has adopted it as soon as one immediate subdirectory carries an `architecture.md`. Every
family still on the legacy tree therefore keeps its legacy rules until its own unit lands. All five
products report `0 finding(s)` on this head.

## What changed

| Kind                | Count | Notes                                                                    |
| ------------------- | ----: | ------------------------------------------------------------------------ |
| Corpus moves        |     2 | `behavior/<p>-cli/gherkin/` → `cli/behaviors/`, content unchanged        |
| Hand-authored docs  |     6 | two `README.md`, two `architecture.md`, two product indexes              |
| Governance edits    |     4 | by hand — these documents describe the legacy tree and used it as example |
| Mechanical sweep    |   172 | prose, globs, `repo-config.yml`, Nx targets, F# `Path.Combine` segments  |

The C4 content the five folders carried is absorbed into each owner's `architecture.md` rather than
dropped: context, containers, components, and the constraints that bind them.

No production code is added. The two `architecture.md` documents and two `README.md` files replace
ten deleted documents, so the reader-facing surface shrinks while the per-owner story becomes one
directory instead of five. The mechanical sweep is a rename with no behavior change: `rhino-cli`'s
2045 covered steps and `crane-cli`'s 141 all still pass.

## Parity

`specs/apps/rhino/cli/behaviors` is inside the byte-identical boundary, so `Parity.fs` names the new
path and `parity-manifest.sha256` is regenerated over the same 226 files. The identical move lands
in `ose-private` next; `crane` exists only here.

## Verification

- `nx run-many -t build,typecheck,lint,test:quick -p rhino-cli,crane-cli --skip-nx-cache` — green
- `specs structure validate` for `rhino`, `crane`, `ayokoding`, `ose`, `organiclever` — `0 finding(s)` each
- `rhino-cli gate run --surface=pre-push` — exit 0, parity manifest current
- readme-index findings fall 378 → 354; none of the failing kinds (`missing`, `orphan`, `ghost`)


wahidyankf (commented):
## Focused Leak Review (`pr-leak-review`)

Scope: real secrets/private values, protected production/staging environment properties, and
real machine-specific absolute filesystem paths only — the aggregate base-to-head diff (319
changed paths: two `git mv`-moved Gherkin corpora, six hand-authored Markdown documents, four
hand-edited convention documents, and 172 mechanically path-rewritten files) plus delivery-controlled
PR title/body. No other security or semantic concern was in scope for this pass.

**Result: PASS — no findings.**

Checked and clear:
- No secrets, credentials, API keys, tokens, private keys, or connection strings in any added/changed
  line (matches on `token`/`secret` are all product identifiers — e.g. `web-ui-token`,
  `O-PUB-WEB-TOKEN` fixture owner IDs, `secrets.json` as a synthetic fixture filename, and a "PASS
  token" log-message string — none is a real credential).
- No absolute machine-specific paths (`/Users/<name>/...`, `/home/<user>/...`,
  `C:\Users\...`, `$HOME`, bare `~/...`) introduced anywhere in the diff, including the
  `parity-manifest.sha256` regeneration and the `plans/in-progress/adopt-beavernest-test-automation/evidence/**`
  captured-transcript files — all paths are repository-relative.
- No protected production/staging environment properties, hostnames, internal URLs, IPs, or
  `.env`-only values committed.
- No email addresses or other personal identifiers.
- PR title/body carry no sensitive values, only repository-relative paths and public plan/PR
  references.

PR #432, opened 2026-09-01
refactor(specs): move AyoKoding's specs to the logical owner shape

wahidyankf:
## What

AyoKoding moves from the legacy five-folder specification tree to a logical owner corpus at
`specs/apps/ayokoding/www/{README.md, architecture.md, behaviors/}`.

The three parallel behavior surfaces collapse into one corpus, split by **perspective** rather than
by deployable — because there is only one deployable to split on:

| Was                                  | Is now                                  |
| ------------------------------------ | --------------------------------------- |
| `behavior/ayokoding-www/gherkin/`    | `www/behaviors/frontend/`               |
| `behavior/ayokoding-be/gherkin/`     | `www/behaviors/backend/`                |
| `behavior/ayokoding-build-tools/gherkin/` | `www/behaviors/build-tools/`       |

The Next.js server tier and the browser tier are two runtime tiers of the same container, and the
build-time index generators run before that container is built. The old `api` slug was already
documented as a *perspective* slug rather than a container; this makes the tree say so.

## Two fixes found along the way

**`repo-config.yml` named a path that has never existed.** All eight AyoKoding corpus globs read
`specs/apps/ayokoding/behavior/www/**`, while the real surfaces were `behavior/ayokoding-www` and
its siblings — so every one of them matched nothing. They now name the corpus that does exist.

**`build` and `test:quick` were not composable in one workspace.** `ayokoding-www`'s vitest unit
project includes `**/*.unit.{test,spec}.{ts,tsx}` and excluded only `node_modules`. `next build`
copies the application — tests included — into `.next/standalone/apps/ayokoding-www/`, so a run
following a build discovered every unit test twice and the copies failed on their own relative
imports. Five phantom suite failures, nothing to do with the code under test. `**/.next/**` is now
excluded. Recorded as `L-9` in the plan's learnings.

## Why this is safe as a mid-sequence step

`specs structure validate` detects the new shape **positively** — a product has adopted it as soon
as one immediate subdirectory carries an `architecture.md`. `ose` and `organiclever` are untouched
and keep their legacy rules until their own units land. All five products report `0 finding(s)`.

## Hand-edited rather than swept

| Document                                                    | Why                                                        |
| ------------------------------------------------------------ | ---------------------------------------------------------- |
| `gherkin-feature-file-placement-and-lib-spec-structure.md`   | describes the legacy tree; used AyoKoding as its example   |
| `specs-application-sync/existing-patterns-to-follow.md`      | its AyoKoding section named a nonexistent `organiclever` path |
| `standardize-app-spec-trees-parity-decisions.md`             | a historical record — keeps the paths as they were         |

No production code is added. Two documents replace nine, so the reader-facing surface shrinks while
the owner's story becomes one directory instead of five. The one behavioral change is the vitest
exclusion, which removes false failures and cannot hide real ones: it excludes copies of tests that
also run from `src/`.

## Verification

- `nx run-many -t build,typecheck,lint,test:quick -p ayokoding-www,ayokoding-www-fe-e2e,ayokoding-www-be-e2e --skip-nx-cache` — green
- `ayokoding-www:test:coverage` — 165 files, 3501 tests, above the 82% line floor
- `specs structure validate ayokoding` — `0 finding(s)`
- `rhino-cli gate run --surface=pre-push` — exit 0


wahidyankf (commented):
## Leak-only review — pr-leak-review

**Head reviewed**: a78f5dbd551334a514c27afea0109ec7e7f79600 (pinned)

Scope: secrets/tokens/credentials, values that belong only in an uncommitted `.env*`, absolute
machine paths / home-directory prefixes, and internal hostnames/private URLs/personal identifiers,
across the full diff against `main`.

**Result: PASS — no findings.**

Checked specifically:
- `repo-config.yml` — all edits are corpus-glob path corrections
  (`specs/apps/ayokoding/behavior/www/**` → `specs/apps/ayokoding/www/behaviors/**`); no
  environment values, hosts, or secrets touched.
- `plans/in-progress/adopt-beavernest-test-automation/learnings.md` (L-9) — the quoted
  test-runner error message is a workspace-relative path
  (`.next/standalone/apps/ayokoding-www/src/features/...`), no machine/home-directory prefix.
- `apps/ayokoding-www/vitest.config.ts` — new `**/.next/**` exclude entry and its comment; no
  paths or values outside the repo tree.
- The full `specs/apps/ayokoding/**` corpus move (175 files) and every touched `project.json` /
  `playwright.config.ts` / `*.steps.ts` `@covers` comment — all path references are
  workspace-relative renames, no absolute paths introduced.
- Repo-wide grep of the diff for `/Users/`, `/home/<user>/`, secret/token/key/password patterns,
  cloud credential prefixes (`AKIA...`, `ghp_...`), private-key PEM headers, and GA measurement-ID
  patterns (`G-...`, `UA-...`) — zero matches in any added line.

No inline comments to post.

PR #433, opened 2026-09-02
refactor(specs): move the library specs to the logical owner shape

wahidyankf:
## What

Moves all four library spec corpora to the logical owner shape and teaches `specs counts validate` to recognize it at a folder root.

`specs/libs/<library>/` now holds exactly `README.md`, `architecture.md`, and a recursive `behaviors/` tree. A library owns exactly one surface, so there is no product directory to nest an owner under — the shape is identical to an app's corpus, one level shallower. That is the rule [`logical-owner-corpus.md`](repo-governance/conventions/structure/specs-directory-structure/logical-owner-corpus.md) already states; this PR makes the tree match it.

## Why this shape

The five folders it replaces were scaffolding, not specification. `product/`, `system-context/`, `containers/`, and `components/` each carried one `README.md` plus a single stub restating the library's own README, and `behavior/gherkin/` held the only content any runner consumed. `architecture.md` says what the library is and who depends on it, in one place a reader finds without traversing four directories that never diverged.

## The validator gap this exposed

`specs counts validate` detects the corpus by looking for an `architecture.md` in an *immediate subdirectory* — where an app keeps one. A library keeps it at the folder root, so after the move every library still failed the five-folder count. `isOwnerCorpusRoot` makes the same positive detection work at the root, and the counts validator then **applies** the corpus rules rather than skipping them: a library that drops its `behaviors/README.md` is a finding here exactly as it would be for an app. Two scenarios in `validate-counts.feature` pin both directions.

## The failure worth naming

`fsharp-crane-core`'s test project copies its feature files in through an MSBuild glob written with backslash separators:

```
<GherkinFiles Include="$(ProjectDir)..\..\..\..\specs\libs\fsharp-crane-core\behavior\gherkin\**\*.feature" />
```

No path rewrite driven by forward slashes touches that. It failed loudly — TickSpec reported `No data found for ... PDF to Markdown Routing` — rather than quietly passing over an empty corpus, which is the behavior you want from a glob that has gone stale.

## Cost

New code is one predicate (`isOwnerCorpusRoot`, three lines) and one branch in `validateSpecCounts`. Everything else is a move, a path rewrite, or a document. The benefit is that the library trees stop being a second shape the reader and the validator both have to know about; Phase 5's final unit can then delete legacy support outright instead of carving an exception for `specs/libs/**`.

## Parity

`apps/rhino-cli/src` and `specs/apps/rhino/cli/behaviors` are inside the parity boundary, so `Specs.fs`, `SpecsSteps.fs`, and `validate-counts.feature` open a propagation obligation against `ose-private`. The regenerated `parity-manifest.sha256` is in this commit; the mirror lands as its own PR there.

## Verification

- `nx run-many -t typecheck,lint,test:quick -p rhino-cli --skip-nx-cache` — green; 1,725 unit tests pass
- `nx run-many -t test:quick -p web-ui,web-ui-token,ts-env-loader,fsharp-crane-core --skip-nx-cache` — green
- `rhino-cli gate run --surface=pre-push` — exit 0
- `rhino-cli governance readme-index validate` — exit 0 (findings are all `unannotated`, which is reported and never fails)
- `rhino-cli md links validate` — 536 broken links, unchanged from the `main` baseline; none in `specs/libs/**`


wahidyankf (commented):
## Focused Leak Review (leak-only, security discipline)

Scope: real secrets/credentials/private values, protected production/staging environment
properties that policy requires outside git, and real machine-specific absolute filesystem paths.
No other security or semantic concern was in scope for this pass.

**Result: no findings.** Inspected the complete base-to-head diff (all 100+ changed files: the
library spec-corpus moves under `specs/libs/**`, `repo-config.yml` corpus-path updates, the
`isOwnerCorpusRoot` validator change in `Specs.fs`/`SpecsSteps.fs`, the `.fsproj` MSBuild glob
rewrite, `.steps.tsx`/`.steps.ts` renames, `bdd-missing-step.json` fixtures, and the
`parity-manifest.sha256` regeneration) plus the delivery-controlled PR title/body.

- No credentials, tokens, or other access-granting values appear in any tracked hunk.
- No production/staging configuration value was moved into a tracked artifact.
- No real machine-specific absolute path (e.g. `/Users/<name>/...`) appears in any added or
  modified file. The one backslash-separated path in
  `libs/fsharp-crane-core/tests/unit/fsharp-crane-core-unit-tests.fsproj`
  (`$(ProjectDir)..\..\..\..\specs\libs\fsharp-crane-core\behaviors\**\*.feature`) is a
  repository-relative MSBuild glob using Windows path separators, not a machine-specific absolute
  path — out of scope for this predicate.

PR #438, opened 2026-09-02
fix(crane-cli): restore 36 Gherkin scenarios lost to a retired corpus path

wahidyankf:
## What

`crane-cli`'s unit suite has been running 36 fewer tests than it looked like it was running. This restores them.

`apps/crane-cli/tests/unit/Suite.fs` resolved its corpus from `GHERKIN_ROOT`, falling back to `specs/apps/crane/behavior/cli/gherkin` — the path Phase 5 retired. `GHERKIN_ROOT` is set nowhere in `apps/crane-cli/project.json` or the test tree, so the fallback is what ran. `Directory.Exists` returned false, and `buildScenarioData` answered with a single no-op row:

```fsharp
if List.isEmpty loaded then
    // Placeholder: step definitions not yet implemented.
    // Returns a single no-op row so [Theory] does not fail with "No data found".
    Seq.singleton [| box "no-op" |]
```

That guard is what converted a corpus that could not be read into a green target.

**Measured, not inferred:**

| | `nx run crane-cli:test:unit --skip-nx-cache` |
| --- | --- |
| before | `Passed! - Failed: 0, Passed: 99, Total: 99` |
| after | `Passed! - Failed: 0, Passed: 135, Total: 135` |

The 36 scenarios were never broken. The 12 step-definition files under `tests/unit/Steps/` bind them correctly and always did — only the root path was wrong, so all 36 pass the moment they are loaded.

## The placeholder is removed, not repaired

A missing directory, an unreadable feature file, and an empty corpus now each raise with the offending path named. Falsified in both directions:

- corpus present → `Passed! ... Total: 135`, exit 0
- `GHERKIN_ROOT` pointed back at the retired path → exit 1, `System.Exception : the Gherkin corpus directory does not exist: .../specs/apps/crane/behavior/cli/gherkin`

That second run is the point: **the same path that now exits 1 previously exited 0.** An exit code that cannot distinguish "passed" from "read nothing" is not a test result.

## Registry corrections

`repo-config.yml` carried the retired crane glob in four places (219, 298, 302, 709) while `apps/crane-cli/project.json` had already been updated to `specs/apps/crane/cli/behaviors`. The registry and the project disagreed, and neither failed. All four now agree.

It also still declared `wahidyankf-www` and `wahidyankf-www-fe-e2e` across four blocks, against a corpus glob matching no file. Both projects were deleted in #423 and neither appears in `nx show projects`. `rhino-cli test-contract registry validate` went from **seven findings to one**; the survivor is `fsharp-env-loader`'s `bootstrap` lifecycle-state, which is by design until Phase 8B.

## What this PR deliberately does not do

The integration tier still reports a single no-op row. `tests/integration/Steps/{PdfSteps,OcrSteps}.fs` are one line each — a module declaration and no bindings at all — and every one of the 37 scenarios in the corpus is tagged `@unit`, so no scenario declares an integration-level obligation. Whether that adapter gets built or governed as `inapplicable` under AC-TEST-06 is a Phase 6 design decision, not a bug fix, and it does not belong in the same commit as a measured regression fix. Its corpus path is corrected here so the follow-up starts from one fewer confound.

## Cost / benefit

Net −64 lines. The suite gains three failure paths it did not have, each naming the path it could not read; it loses the branch that swallowed all three. The benefit is 36 scenarios of `crane-cli` behaviour that are now actually executed on every `test:quick`, plus a registry that no longer claims two deleted projects.

## Verification

- `nx run-many -t typecheck,lint,test:quick,test:integration -p crane-cli --skip-nx-cache` — exit 0.
- `rhino-cli gate run --surface=pre-push` — exit 0.
- `rhino-cli test-contract registry validate` — 7 findings → 1 (expected bootstrap).
- `fantomas` applied; no parity-boundary path touched, so no cross-repo propagation is owed.

---

## Second commit: plan learnings

`8c94de1e6` appends two entries to the plan's `learnings.md`:

- **L-10** — the no-op placeholder row, added so `[Theory]` would not fail with "No data found", is what turned an unreadable corpus into a passing suite. `specs:behavior:coverage` passed throughout, because it scans step sources statically and never asks whether the runtime found the corpus — so neither signal available at the time could have caught this.
- **L-11** — all six `test-contract` validators built in Phase 4 have **zero call sites** across `gates:`, every `project.json`, and every workflow. Four of them read only fixtures, which is the known L-8 gap and not a defect. But `registry validate` and `registry validate-mapping` read the real `repo-config.yml` and are gated nowhere — which is exactly why the `wahidyankf` orphans this PR removes survived from #423 until now. Wiring those two is filed as follow-up work, not done here.

### DIFF
--- a/apps/crane-cli/tests/integration/Suite.fs
+++ b/apps/crane-cli/tests/integration/Suite.fs
@@ -9,7 +9,7 @@ let private assembly = Assembly.GetExecutingAssembly()
 
 let private gherkinRoot =
     match System.Environment.GetEnvironmentVariable("GHERKIN_ROOT") with
-    | null -> Path.Combine(__SOURCE_DIRECTORY__, "../../../../specs/apps/crane/behavior/cli/gherkin")
+    | null -> Path.Combine(__SOURCE_DIRECTORY__, "../../../../specs/apps/crane/cli/behaviors")
     | root -> root
 
 /// Returns loaded Gherkin scenarios, or a single no-op placeholder when none can be loaded.
--- a/apps/crane-cli/tests/unit/Suite.fs
+++ b/apps/crane-cli/tests/unit/Suite.fs
@@ -9,36 +9,40 @@ let private assembly = Assembly.GetExecutingAssembly()
 
 let private gherkinRoot =
     match System.Environment.GetEnvironmentVariable("GHERKIN_ROOT") with
-    | null -> Path.Combine(__SOURCE_DIRECTORY__, "../../../../specs/apps/crane/behavior/cli/gherkin")
+    | null -> Path.Combine(__SOURCE_DIRECTORY__, "../../../../specs/apps/crane/cli/behaviors")
     | root -> root
 
-/// Returns loaded Gherkin scenarios, or a single no-op placeholder when none can be loaded.
-/// Prevents xUnit from failing with "No data found" when step definitions are not yet implemented.
+/// Every scenario in the owner corpus, one xUnit row each.
+///
+/// This function raises rather than degrading. A corpus that cannot be read is
+/// indistinguishable at the exit code from a corpus that passes, so an earlier
+/// version returned a single no-op row "so [Theory] does not fail with No data
+/// found" — which is exactly what let a retired `gherkinRoot` path hide 36
+/// scenarios behind a green target. A missing directory, an unreadable feature
+/// file, and an empty corpus are all failures here.
 let private buildScenarioData () : seq<obj[]> =
+    if not (Directory.Exists gherkinRoot) then
+        failwithf "the Gherkin corpus directory does not exist: %s" gherkinRoot
+
+    let files =
+        Directory.GetFiles(gherkinRoot, "*.feature", SearchOption.AllDirectories)
+
+    if Array.isEmpty files then
+        failwithf "no .feature file was found under %s" gherkinRoot
+
+    let defs = StepDefinitions(assembly)
+
     let loaded =
-        if Directory.Exists(gherkinRoot) then
-            let files =
-                Directory.GetFiles(gherkinRoot, "*.feature", SearchOption.AllDirectories)
-
-            let defs = StepDefinitions(assembly)
-
-            files
-            |> Seq.collect (fun path ->
-                try
-                    let feature = defs.GenerateFeature(path)
-                    feature.Scenarios |> Seq.map (fun scenario -> [| scenario :> obj |])
-                with _ ->
-                    Seq.empty)
-            |> Seq.toList
-        else
-            []
+        files
+        |> Seq.collect (fun path ->
+            let feature = defs.GenerateFeature(path)
+            feature.Scenarios |> Seq.map (fun scenario -> [| scenario :> obj |]))
+        |> Seq.toList
 
     if List.isEmpty loaded then
-        // Placeholder: step definitions not yet implemented.
-        // Returns a single no-op row so [Theory] does not fail with "No data found".
-        Seq.singleton [| box "no-op" |]
-    else
-        loaded :> seq<_>
+        failwithf "the %d feature file(s) under %s expanded to no scenarios" files.Length gherkinRoot
+
+    loaded :> seq<_>
 
 type CraneCliUnitSuite() =
     static member Scenarios() : seq<obj[]> =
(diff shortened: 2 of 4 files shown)


wahidyankf (commented):
## Focused Leak Review (leak-only)

Scope: exactly the three `pr-leak-review` predicates — real secrets/private values,
protected production/staging properties, and real machine-specific absolute filesystem
paths. Every other security/semantic concern is out of scope for this pass.

**Files inspected**: `apps/crane-cli/tests/unit/Suite.fs`, `apps/crane-cli/tests/integration/Suite.fs`,
`repo-config.yml`, plus the delivery-controlled PR title/body.

### secret-or-private-value: none found

No credentials, tokens, keys, or other access-granting values appear in the diff.

### protected-environment-property: none found

No production/staging configuration values are introduced. The `repo-config.yml` changes are
corpus-glob path corrections (`specs/apps/crane/behavior/cli/**` -> `specs/apps/crane/cli/behaviors/**`)
and removal of two retired-project registry blocks (`wahidyankf-www`, `wahidyankf-www-fe-e2e`) —
neither carries a protected property.

### machine-specific-absolute-path: none found

Both `Suite.fs` files build the `GHERKIN_ROOT` fallback from `__SOURCE_DIRECTORY__` plus a
portable relative segment (`Path.Combine(__SOURCE_DIRECTORY__, "../../../../specs/apps/crane/cli/behaviors")`
in the unit suite, `"../../../../specs/apps/crane/cli/behaviors"` in the integration suite) — no
hardcoded machine-specific absolute path is committed. The new `failwithf` format strings
(`"the Gherkin corpus directory does not exist: %s"`, `"no .feature file was found under %s"`,
`"the %d feature file(s) under %s expanded to no scenarios"`, `"expected a TickSpec Scenario row,
got %O"`) interpolate that same runtime-resolved, portable value; the format strings themselves
contain no hardcoded path.

The PR body's `## The placeholder is removed, not repaired` section quotes a runtime exception as
`System.Exception : the Gherkin corpus directory does not exist: .../specs/apps/crane/behavior/cli/gherkin`
— the leading path segment is elided (`...`), and neither the body nor any PR/review comment on
this PR contains a real `/Users/...` (or other machine-specific) absolute path. Verified by
grepping the fetched body and comment text for `/Users`, `/home`, and Windows-style `C:\Users`
patterns with zero matches.

### Result

**result: pass** — 0 findings across all three categories on this pinned head.

wahidyankf (commented):
ose-pr-leak-review:v1

result: pass

Scope: leak-only review of PR #438 at pinned head `8c94de1e6eaf8f99d7a084d19d4f1cc66b737d59`, covering the full base-to-head diff (merge-base `1d3dcef84f0d015e04f034154bb788b83f14630f`) plus delivery-controlled PR metadata (title, body, comments). A prior leak review (id 5085240920) already passed the earlier head `f221f6141751562e6755d66aeb544e4721d77f38`; this pass re-covers the full diff and gives particular attention to the one new commit, `8c94de1e6` ("docs(plan): record the placeholder-suite and unwired-validator learnings").

Findings by category:

1. Real secrets committed to a git-tracked file — none found. Scanned the full diff for API-key/token/secret/password patterns, private-key headers, and cloud-credential shapes; no matches.
2. Protected environment properties — none found. `repo-config.yml` and the `docs/` / `specs/` changes in this diff are registry/path corrections and documentation prose, not environment configuration.
3. Real hostnames/IPs/personal data/machine-specific absolute paths — none found. Checked specifically for the L-3-class risk called out in the task (an absolute `/Users/<name>/...` path leaking via a learnings entry that quotes local command output). The new commit's `learnings.md` additions (L-10, L-11) quote command output (`nx run crane-cli:test:unit`, `rhino-cli test-contract registry validate`) and reference only repo-relative paths (e.g. `apps/crane-cli/tests/unit/Suite.fs`, `repo-config.yml`); no absolute filesystem path, hostname, or IP appears in either entry. The PR body's exception-message excerpt (`the Gherkin corpus directory does not exist: .../specs/apps/crane/behavior/cli/gherkin`) is deliberately elided with `...` and carries no machine-specific path. A regex sweep for `/Users/<name>`, `/home/<name>`, and Windows `C:\Users\` shapes across the full diff returned zero true positives (one `/home/` grep hit was a false positive on the literal string `www/behaviors/frontend/home/home.feature`).

No leak-review findings to raise. This PR is clear on all three categories.

PR #449, opened 2026-09-03
test(organiclever-app-web): migrate to the target test contract (Phase 14)

wahidyankf:
## Summary

Phase 14 (`D-O-PUB-OL-WEB`) of the BeaverNest test-automation adoption plan — migrates `organiclever-app-web` to the target test contract already applied to `web-ui` (#443), `ts-ui` (ose-private #145), and `ayokoding-www` (#446).

- Relocates all 58 pre-existing unit test files from `src`-colocated `*.unit.test.ts(x)` / `test/unit/**` into `apps/organiclever-app-web/tests/unit/**`, mirroring `src/`. Imports rewritten to the new relative depth; zero behavioral changes to relocated files beyond that (one file renamed `journal-store.int.test.ts` → `journal-store-batch.unit.test.ts` to avoid colliding with the existing `journal-store.unit.test.ts` sibling, since its content — migration idempotency, batch atomicity, sort tiebreakers — folds into the unit tier per `repo-config.yml`'s `integration: disposition: inapplicable` for this project).
- Flattens `vitest.config.ts` from a two-project (`unit`/`integration`) config to a single top-level config; `test:integration` becomes a documented no-op per the registry's `inapplicable` disposition — no real integration boundary exists (PGlite is in-process, not an isolated local resource).
- Raises `test:coverage`'s line-coverage floor from 74% to 99% (the sole repository-enforced metric per AC-COVERAGE-01; functions/branches/statements keep their pre-migration 75% bar).
- Fixes a latent `test:unit` target bug: `cwd` was `{projectRoot}` (an Nx macro the F# `test-contract` static reader cannot interpolate), blocking `test:layout:validation` from resolving `vitest.config.ts`; changed to the literal `apps/organiclever-app-web`.
- Adds the three policy-validation Nx targets (`test:layout:validation`, `coverage:policy:validation`, `package-manifest:policy:validation`), each invoking `rhino-cli test-contract <kind> validate`.
- Closes every genuine coverage gap the raised floor exposed with **real, behavior-asserting tests** — no `istanbul-ignore`, no weakened assertions, no lowered bar. ~40 new tests added across 23 files (5 brand-new test files, 18 extended).
- `repo-config.yml`'s `testing.projects[]` and `compatibility.mappings[]`/`behavior.adapters` entries already matched the target shape (unit=required, integration=inapplicable, e2e=delegated) — verified, no changes needed.

## Before / after

| | Before | After |
|---|---|---|
[...]

PR #450, opened 2026-09-03
docs(adopt-beavernest): record D-O-PUB-OL-WEB delivery evidence (Phase 14)

wahidyankf:
## Summary
Records the Execution-state blockquote in `delivery.md` and the matching `EVIDENCE` row in `implementation-notes.md` for `D-O-PUB-OL-WEB` (Phase 14, PR #449).

Docs-only, no source changes.


wahidyankf (commented):
## pr-leak-review (exact leak-only mode)

Focused review of exactly three categories: secrets/private values, protected environment
properties, and machine-specific absolute paths. No other security or semantic concern was in
scope.

**Result: pass — 0 findings.**

Verified rather than assumed, across the full base→head aggregate diff (`git diff
0544d77f3e94ae93e203f846e5658ade1e355299...803b6563880a036eb7ed881ccbaf957744574377`,
2 files changed, +29/-0):

- Full-diff regex sweep for `/Users/...`, `/home/...`, `/tmp/...`, `/private/...`,
  `/var/folders/...`, `/opt/...`, `/etc/...`, and Windows `C:\Users\...` paths: zero hits across
  both changed files. Every path referenced in the added prose (`apps/organiclever-app-web`,
  `tests/unit/steps/landing`, `test/unit/steps/landing`, `routine-store.ts`, `repo-config.yml`)
  is repository-relative, not a machine-local absolute path — including the narrative describing
  a local verification run (`forced-fresh test:quick run`), which reports counts/outcomes only,
  no filesystem path from the operator's machine.
- Secret/token/credential shape patterns (`api_key`, `secret`, `password`, `access_key`,
  `private_key`, `bearer`, `credential`, AWS/PEM/PGP header shapes): zero hits. The one `TOKEN`
  substring match is `D-O-PRI-TS-TOKEN`, a pre-existing plan-item identifier name from an
  unrelated prior EVIDENCE row, not a credential value.
- No `.env`, `process.env`, or protected production/staging property reference appears in either
  changed file.
- No binary file is touched — confirmed via `git diff` returning zero `Binary files ... differ`
  markers across the full diff, not assumed from the description.
- PR title/body inspected as delivery-controlled metadata: describes the doc-only evidence
  recording purpose, no additional leak-relevant content beyond the diff already covered.
- No PR comments or prior reviews exist on this PR to inspect as additional delivery-controlled
  metadata.

Every other security and semantic concern is out of scope for this focused review.

PR #463, opened 2026-09-03
test(adopt-beavernest): roll out real BDD adapter coverage (Phase 20A)

wahidyankf:
## Summary

Rolls out `test:behavior:coverage:{unit,integration,e2e}` to every remaining migrated project in this repo, fixes two latent rhino-cli bugs the rollout's first real-scale run surfaced, and closes every real BDD-coverage gap the fixed validator found.

- **Mechanical rollout** (26 files): adds the 4-target block to every migrated project not yet carrying it, plus switches `rhino-cli`'s own 3 `test:behavior:coverage:*` targets from raw `dotnet run` to `rhino-bin.sh` (a concurrent `nx run-many` fan-out was racing on the shared build output).
- **rhino-cli fixes** (2 files, `TestContractProject.fs` + `repo-config.yml`):
  - Scenario Outline steps only matched substituted per-example text; `@amiceli/vitest-cucumber`'s `ScenarioOutline` API binds only the raw `<placeholder>` template text, so every such step read as unbound. Fixed with the same OR-fallback pattern `Specs.stepCovered` already uses.
  - A delegated adapter's corpus resolved from the *originally requested* project's row instead of the *actual driver-hosting* project's row — latent since every delegate row left `corpus:` empty (inheriting the owner's) until `ayokoding-www-fe-e2e`'s row was given its own broader corpus for its second, in-situ binding of `libs/web-ui`'s `resizable-panel.feature`.
- **Real gap closure** (120 files): every genuine gap the fixed validator found — missing `When` keywords (`bdd-missing-required-keyword`, since binding matches on step text but a full-feature structural parser like `describeFeature` is keyword-strict at runtime), two silent-default-to-fail-fast production fixes (`NatsClient.fs` in `organiclever-be`/`ose-be`), and new real drivers for previously-uncovered scenarios.
- **Follow-up fix** (18 files, `fix(specs): split messaging behaviors...`): the pre-push gate's own `test:quick` caught a real regression from the gap-closure — `organiclever-be`/`ose-be`'s `specs:behavior:coverage` blanket-excluded their whole `messaging/` directory since it was added (back when nothing in it had a real unit driver). The gap-closure's new `nats-config.feature` driver made that exclusion stale, so its steps read as orphans. Splits the directory: scenarios needing a live NATS/JetStream broker move to `messaging/live/` (still excluded), `nats-config.feature` (now genuinely unit-covered) stays scanned.
- **Parity propagation** (`chore(rhino-cli): regenerate parity manifest`): `TestContractProject.fs` is a byte-identical parity-boundary file with ose-private; a companion PR there ports the same two bug fixes.

## Verification (independently re-run, not just agent-reported)

- `test-contract bdd validate --project <P> --adapter <A>` for every rolled-out project: all `behavior-coverage-valid` or correctly `not-applicable`/`delegated`
- `nx affected -t test:quick --base=origin/main --head=HEAD`: 26 projects, 42 tasks, all green (confirmed twice — once triggering the real regression, once clean after the fix)
- `dotnet test` rhino-cli's own unit suite: 2460 passed, 0 failed
- `governance readme-index validate`: passed
- `parity manifest validate`: current

## Test plan

- [x] Full affected sweep passes clean
- [x] rhino-cli's own unit suite passes
- [x] README index audit passes
- [x] Parity manifest current, companion fix pushed to ose-private

https://claude.ai/code/session_01SAVoSfk5AochhrgQU2s343


wahidyankf (commented):
## Focused PR Leak Review (`pr-leak-review`)

Scope: exactly three categories — real secrets/private/access-granting values, production/staging
properties that policy requires in environment or secret storage, and real machine-specific
pass.

**Result: PASS** — no leak found across the complete base-to-head diff (163 changed files) or the
delivery-controlled PR title/body.

PR #477, opened 2026-09-05
ci(actions): bound storage usage

wahidyankf:
## Summary

- keep non-main GitHub Actions caches restore-only while main owns cache saves
- publish the Rhino PR handoff as a framework-dependent artifact
- provision .NET for every artifact consumer
- add the propagated GitHub Actions storage-budget rule

## Cost / benefit

The workflow adds explicit cache-scope conditions and relies on the runner toolchain for framework-dependent handoffs. In return, the representative compressed Rhino artifact drops from about 39.8 MB to about 2.7 MB and cache churn is bounded.

## Validation

- pre-push quality gate
- actionlint
- Rhino quick tests and coverage
- artifact-retention policy
- parity manifest
- rules-propagation strict checker, double-clean

## Parity

Matching private delivery: https://github.com/wahidyankf/ose-private/pull/158


wahidyankf (commented):
Focused leak review: pass.

- Secret or private values: 0
- Protected environment properties: 0
- Machine-specific absolute paths: 0

This review covered only the three focused leak categories. Every other security and semantic concern was out of scope.

PR #480, opened 2026-09-05
fix(hippo): read cache timestamps through the platform branch

wahidyankf:
## What

`prune_cache` read release-directory timestamps through
`stat -f '%m' "$path" 2>/dev/null || stat -c '%Y' "$path"`. GNU coreutils reads `-f` as
`--file-system`, so on Linux `stat` wrote a block of filesystem detail to **stdout** before failing
on the format operand, and the fallback then appended the correct epoch. `2>/dev/null` suppresses
only stderr, so the command substitution captured the report concatenated with the value. The
numeric guard `case "$modified" in '' | *[!0-9]*) continue ;; esac` therefore skipped **every**
candidate and cache retention reclaimed nothing at all on Linux.

It fails safe — it never deletes — so the only symptom is a consumer cache that grows without
bound. macOS is unaffected, which is why no local run ever caught it.

## Evidence

- First-ever execution of `hippo-consumer-smoke.yml` failed on `ubuntu-24.04` and passed on
  `macos-15` (runs `33994712170`, `33994713323`).
- Reproduced in a Linux container and traced with `sh -x` to the retention assertion `[ 3 -eq 2 ]`.
  The container needs `procps` installed, or release-claim identity fails first and masks the fault.
- After the fix the Linux bootstrap suite exits 0 and the macOS consumer contract exits 0.
- Both guard controls run: reintroducing the fallback chain exits 1 with the message, the fixed tree
  exits 0.

## Why the surface changed too

The defect reached `main` through a green pull request because the consumer suite ran on **no**
pull-request surface — only on a schedule that had never fired. The workflow now carries a
path-filtered `pull_request` trigger, so this very PR proves the boundary on both operating systems
before it merges.

Twelve lines of new non-test code: a static guard in the bootstrap suite and a trigger block.

- **Cost** — one extra CI job pair on pull requests that touch the consumer boundary, roughly
  21 seconds of suite time each plus runner setup. The path filter keeps every other pull request
  untouched. The guard asserts code shape rather than behaviour, which is a weaker kind of test.
- **Benefit** — the shape assertion is the only check that runs on the platform where this class of
  defect is *invisible*, so it stops a macOS-only workstation reintroducing a Linux-only fault. The
  trigger converts a twice-daily scheduled signal that had never fired into a pre-merge one.
- **Rejected alternative** — asserting retention behaviour on macOS only. That is what already
  existed, and it is exactly what missed this.

### DIFF
--- a/.github/scripts/test-hippo-bootstrap.sh
+++ b/.github/scripts/test-hippo-bootstrap.sh
@@ -10,6 +10,18 @@ trap 'rm -rf -- "$temporary_root"' EXIT HUP INT TERM
 subject="$temporary_root/consumer/hippo"
 mkdir -p "$temporary_root/consumer" "$temporary_root/fake-bin" "$temporary_root/payload"
 cp "$repository_root/hippo" "$subject"
+
+# Timestamp reads must go through the platform-branching helper. The BSD form
+# `stat -f` means "filesystem status" to GNU coreutils, which writes a block of
+# filesystem detail to stdout before failing, so a `stat -f ... || stat -c ...`
+# fallback chain captures that detail alongside the real value and yields a
+# non-numeric result. Retention then silently treats every release as undatable
+# and reclaims nothing. That failure is invisible on macOS, so assert the shape
+# here rather than waiting for a Linux-only run to catch it.
+if grep -qE "stat -[fc] .*\|\| *stat -[fc] " "$repository_root/hippo"; then
+	echo "hippo must read timestamps through file_modified_epoch, not a stat fallback chain" >&2
+	exit 1
+fi
 chmod 755 "$subject"
 
 # The suite runs on both Linux and macOS runners. Pinning the fake uname to one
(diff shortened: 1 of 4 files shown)


wahidyankf (commented):
Focused leak-only review of the exact current head. Scope is limited to three categories: real
secrets or private values, protected production or staging environment properties that belong
outside git, and machine-specific absolute paths. No broad security or semantic review was
performed, and every other concern is out of scope for this pass.

Inspected the complete aggregate base-to-head diff across all four files, both added and removed
lines, plus the delivery-controlled title and body. Every path introduced is repository-relative
(`hippo`, `hippo.lock`, `hippo.local.json.example`, `.github/scripts/*`, `.github/workflows/*`),
the workflow reads only `RUNNER_TEMP` and `github.workspace`, and the assertion prints observed
contract values rather than any environment content.

Result: clean. No finding in any category.

PR #484, opened 2026-09-07
chore(hippo): pin the consumer to v0.5.1

wahidyankf:
Pins the checksum-verified `./hippo` consumer from v0.4.0 to v0.5.1.

## What v0.5.1 brings

- `run --wait-for-admission <duration>` retries a capacity deferral in-process, backing off from 100ms to a 2s ceiling and still reporting `75` if capacity never frees. Zero remains the default, so nothing changes unless a call site opts in.
- `status --json` reports `abandonedProcessGroups`: payloads still running under a guard that died. HIPPO reports these and never signals them, because a recycled process group would name an unrelated process.
- Conformance gained a `deferralRetryProbe` that verifies a consumer survives exit `75` rather than reading it as an admission. No consumer in this repo ships a manifest yet, so nothing is wired to it here.

## Cost and benefit

Cost is one pinned release bump: four checksums and this commit. Benefit is that local compute stops needing a bespoke deferral loop at every call site, and an operator can finally see a payload left running by a dead guard, which is the one leak `status` previously could not surface.

## Verification

Verified against the published release rather than a local build: the bootstrap installs v0.5.1, runs work under it, and still fails closed with exit `78` on a tampered host checksum while a tampered non-host checksum correctly does not block. `npm install` and `npm run doctor -- --fix` both completed in this worktree under v0.5.1.

The consumer bootstrap itself needs no change; it already branches on `$goos` for `stat`.

https://claude.ai/code/session_019QXPT6kJZZZsNrxQDsPRne


wahidyankf (commented):
## Focused PR Leak Review — exact leak-only mode

Reviewed the complete aggregate base-to-head diff (`2b7b0b4...08f0e2b`), added and removed lines, plus the PR title and body, against exactly three categories: real secrets/private values, protected production or staging environment properties that belong outside git, and real machine-specific absolute paths.

The aggregate diff touches one tracked file, `hippo.lock`, replacing a pinned upstream release identifier, an upstream commit identifier, and four per-platform release-artifact integrity checksums. Verification digests of published release artifacts are public integrity metadata: they grant no access, carry no environment configuration, and are inherently portable. The PR title and body describe the bump and its verification and contain no protected property or absolute filesystem path; the only paths referenced are repository-relative.

No leak in any of the three categories. Every other security and semantic concern — including injection, unsafe git/FS behaviour, logic, architecture, and governance — was out of scope for this pass and was not reviewed.

wahidyankf (commented):
## Focused leak review — exact current head

Scope was the three leak categories only: real secrets or private values, protected production/staging
properties that belong outside git, and real machine-specific absolute paths. Every other security and
semantic concern was out of scope for this pass, including injection, unsafe git/FS behaviour, and all
convention review.

Reviewed the complete aggregate `main`...head diff (added and removed lines) plus the delivery-controlled
PR title and body. The head moved since the earlier leak review, so this is a fresh pass pinned to the new
head; the prior review does not authorize it.

The aggregate diff touches a single tracked file, replacing a pinned upstream release identifier and four
per-platform artifact integrity digests. Those digests are published release-verification values for a
public upstream repository, not access-granting credentials, so they fall under the documented-public-value
exclusion. No environment or secret-storage property, no production or staging configuration, and no
absolute filesystem path appears on any added or removed line. The PR title and body carry only release
notes, exit-code behaviour, and the repository-mandated commit trailer.

Result: no leak in any of the three categories at this head.

PR #485, opened 2026-09-07
chore(hippo): pin the consumer to v0.5.2

wahidyankf:
Pins the HIPPO consumer to v0.5.2.

## What changed upstream

`--wait-for-admission` backs off from 100 ms to a 2 s ceiling, so a ten-minute
budget is roughly three hundred attempts. Every attempt printed the same deferral
sentence, which buried the surrender line and every other notice under hundreds
of identical copies. v0.5.2 prints that notice once; the surrender still reports
the total.

Quieting is scoped to the deferral alone. Storage refusals, warning-pressure
admissions, shedding notices, and deferred coordination cleanup stay audible on
every attempt, so a waiting caller can never miss a refusal it has to act on.

## Verification

- All four v0.5.2 checksums recomputed locally from the downloaded release assets
  rather than copied from the published `checksums.txt`; all four matched.
- The released `darwin_arm64` binary was exercised against a saturated
  coordination root: 12 attempts, one notice, surrender intact, exit 75.
- The pinned consumer bootstrap was run in this worktree against the new pin, and
  printed a warning-pressure admission notice — direct evidence that non-deferral
  notices remain audible.

## Cost/benefit

No new product code — one pinned-version line change, six lines. The benefit is
that a deferred build's stderr stays readable, so a real refusal is not lost in
the noise.


wahidyankf (commented):
Focused leak-only review of the exact current head. Scope is limited to three categories: real
secrets or private values, protected production or staging environment properties that belong
outside git, and machine-specific absolute paths. No broad security or semantic review was
performed, and every other concern is out of scope for this pass.

Inspected the complete aggregate base-to-head diff — one file, `hippo.lock`, six lines removed and
six added — plus the delivery-controlled title and body. Every added value is a public release
identifier: a semantic version, an upstream commit SHA, and four SHA-256 digests of published
release archives. There is no credential, token, or private value; no production or staging
environment property; and no absolute path of any kind, machine-specific or otherwise.

Result: clean. No finding in any category.

PR #486, opened 2026-09-07
feat(governance): add the rules-grooming sweep and cut repo-governance frontmatter to two keys

wahidyankf:
## What this does

Two pieces of corpus-wide governance work, both aimed at the same thing — the rule tree
carrying less that binds nothing — and both proved to have lost no obligation.

1. **`rules-grooming`** — a new recurring sweep that removes *representation* carrying no
   obligation, hands every reduction to `rules-propagation` to write, and is run three
   times here against the live corpus.
2. **The `repo-governance/` frontmatter allow-list** — every `.md` in the tree reduced to
   exactly `description` and `when_to_use`, with the schema rewritten from a minimum into
   an allow-list so the metadata cannot regrow.

---

# Part 1 — `repo-governance/` frontmatter: thirteen keys to two

## The rule

> Every Markdown file under `repo-governance/` carries exactly two frontmatter keys,
> `description` and `when_to_use`. Both are required and non-empty. Any other key is a
> validation failure.

Falsifiable both ways: a file with a third key fails; a file missing either key fails.
Canonical home: [`conventions/structure/governance-frontmatter.md`](repo-governance/conventions/structure/governance-frontmatter.md).

## What was there, and who read it

| Key | Files | Read by |
|---|---:|---|
| `title` | 2,418 | `readme-index generate` only, and only for entries *absent* from an index |
| `description` | 2,415 | index generation — **kept** |
| `when_to_use` | 2,414 | index generation — **kept** |
| `category` | 1,798 | nothing |
| `subcategory` | 1,795 | nothing |
| `tags` | 1,794 | nothing |
| `created` | 1,774 | nothing |
| `name`/`goal`/`termination`/`inputs`/`outputs` | 33–34 | nothing — read by humans, from the body |
| `status` | 14 | nothing |

None of `category`, `subcategory`, `tags`, `created` is read anywhere in `apps/`, `libs/`,
`tools/`, `scripts/`, or `.github/`. No site build ingests `repo-governance/`.

**`title` was the interesting one, and it was settled by measurement rather than argument.**
Its only consumer is `governance readme-index generate`, and only for entries not already
in an index; the `validate` path never opens frontmatter. So nothing has ever compared a
generated entry against the `title` it came from — and **19.6% of the 4,183 index entries
have already drifted away from it**, undetected. A field written once, read almost never,
and checked never is not load-bearing. It is deleted; the generator falls back to the
title-cased filename stem, which `generatedTreeEntry` now does too (it previously fell back
to the raw kebab stem, so the two paths disagreed).

`created:` was the other notable case: the repo *already* forbids `updated:`, `**Last
Updated**` footers, and `- **Created**:` body bullets, on the stated rationale that git
history is the single source of truth for change dates. Frontmatter `created:` was the gap
in that rule, not an exception to it. It is now audited as forbidden — but only under
`repo-governance/`, since `docs/` and site content still carry a legitimate `created` date.

## The reduction

| | Before | After |
|---|---:|---:|
| Distinct frontmatter keys | 13 | **2** |
| Files carrying exactly `{description, when_to_use}` | 0 | **2,419 / 2,419** |
| Corpus words | 782,501 | 739,811 |
| Frontmatter lines | 26,547 | 7,257 |
| Frontmatter share of the tree | 20.8% | **6.6%** |

−42,690 words, −19,290 lines of frontmatter. This matters beyond tidiness because the word
budget counts frontmatter fully — there is no prose-only carve-out — so every file was
spending budget on metadata nothing read.

## Nothing was deleted before it was moved

Content migration ran per file, *before* the strip, so no obligation passed through an
intermediate state where it existed nowhere:

- **34 workflow files** — `goal`, `termination`, `inputs`, `outputs` became body sections
  (`## Goal and Termination`, `## Inputs`, `## Outputs`). Rendered as **bullets, not
  tables**, deliberately: `wc -w` counts every `|` in a Markdown table as a token, and four
  of these files sit near the 750-word fail line. The plan called for tables; measurement
  overruled it.
- **`name:`** was dropped rather than migrated, after verifying per file that it equals the
  filename stem in all 34.
- **2 files** were progressive-disclosure shards *of an inputs block* — their entire payload
  was frontmatter. They are now ordinary body documents.
- **6 files had no H1** and would have lost their only title; each gained one.
- **4 files** in the previously-unvalidated subtrees lacked `description` or `when_to_use`
  and were backfilled.
- **14 `status: "Pilot — initial issue"`** markers were resolved against current state.

## Enforcement: an allow-list, not a minimum

A minimum schema would have permitted every removed key straight back. The change is in
`Md.fs`:

- `governanceDocPrefixes` now covers **all of `repo-governance/`** rather than four named
  sub-trees. `glossary/`, `vision/`, `repository-governance-architecture/`, and the tree's
  root files previously classified as `UnknownArea` and received **zero** validation — which
  is exactly where the missing-key gaps were.
- `validateGovernanceSchema` requires `description` + `when_to_use` non-empty, drops the
  `title` requirement, and **fails any key outside the allowed set**.
- `validateSoftwareSchema` is untouched; `docs/explanation/software-engineering/` keeps its
  five-key schema, with a scenario pinning that it is unaffected.

Enforcement is automatic — `md frontmatter validate` is already wired into `lint-staged` for
every staged `*.md`.

**Gherkin first**, per the BDD contract: 7 new scenarios in `docs-validate-frontmatter.feature`
(missing either key fails, both keys pass, `title` fails the allow-list, `category` fails the
allow-list, the `glossary/` subtree is now validated, the SE schema is unaffected) and 1 in
`repo-governance-frontmatter-audit.feature` binding `created:` as forbidden. Unit and
Integration bindings for all of them.

## Conflicting rules were amended, not left to contradict

Five surfaces asserted the old schema. Each was corrected as part of the same delivery — a
rule that lands while its contradiction survives has not landed:

| Surface | Was |
|---|---|
| `workflow-identifier/workflow-structure.md` | Canonical template mandating `name`/`goal`/`termination`/`inputs`/`outputs` |
| `criticality-levels/…-critical.md` | Classed a missing `subcategory` as **Critical** |
| `criticality-levels/domain-specific-examples-….md` | Classed missing `goal`/`termination` as a defect |
| `workflow-identifier.md` + its `Contents` | Described "the frontmatter and body structure" |
| `indentation/yaml-and-code-block-indentation.md` | 2-space frontmatter rule — narrowed in scope, not deleted |

Two `no-date-metadata` FAIL/PASS example pairs were also repaired along the way: the FAIL
block was byte-identical to the PASS block, demonstrating nothing.

## Preservation

Two independent proofs, same standard the grooming runs are held to:

- **Obligation diff** — 4,352 → 4,322 normative sentences. Of the 87 lines that moved, 69
  are the deleted keys themselves, 11 are block-scalar reflow confirmed present elsewhere,
  and **7 are rules deliberately amended above**. Real losses: **0**.
- **Declaration containment** — every input/output `name`, `type`, enum `value`, `default`,
  `pattern`, description token, and `required` flag from all 34 workflow files checked
  against the post-migration file. **275/275 preserved.** The two initial misses were
  prettier escaping `(PASS/BLOCKED_*)` to `(PASS/BLOCKED\_\*)` — the same string.

Broken internal links: the repository's own `md-links` gate reports **zero broken links**
before and after. A stricter ad-hoc sweep (which also follows links inside fenced examples)
reports the same count either way, **zero newly broken and zero fixed**. Its first comparison
appeared to show 104 new *and* 104 fixed; that was line numbers shifting as frontmatter
shrank — comparing file→target pairs instead gives an exact match.

---
# Part 2 — `rules-grooming`: the sweep that found this work

Adds `rules-grooming` — a recurring, corpus-wide sweep that removes **representation that
carries no obligation**, and hands every reduction to `rules-propagation` to write.

No workflow owned this before. `rules-propagation` places one rule at a time;
`rules-quality-gate` audits one rule and its points of use and is forbidden from auditing
unrelated governance. Neither is allowed to look at the corpus as a whole, so nothing did.

The PR also **runs** the workflow, three times, and lands what each run found — including
the defects the runs found in the workflow itself.

## The reduction — grooming runs only

| | Before | After |
|---|---:|---:|
| Files | 3,173 | 3,054 |
| Lines | 165,006 | 163,512 |
| Metadata ratio | 19.64% | 18.88% |

These are the grooming runs' own figures, measured across the whole repo before Part 1 ran.
Part 1's frontmatter reduction is counted separately above.

**−1,494 lines, zero obligations lost**, across three runs: 99 shards reabsorbed in run 1,
27 more in run 3 (the first sweep with all four classes running by default).

Preservation was proved two independent ways on every run — line-level containment
(1,045/1,045 then 281/281 merged lines present in their parent) and an obligation-set diff
(3,661→3,661, then 3,679→3,679). Two proofs rather than one because containment catches a
truncated move that an obligation diff misses when the lost text carried no modal.

Run 3 also re-confirmed the packing rule at a sharper ratio: **582 shards qualified
individually, 27 survived packing per parent at 80% of budget** — a 21× over-report had it
been measured per shard. No parent finished above 520 of its 650 words. Entry points were
checked explicitly before any deletion; all 27 removals sat inside `<name>/` shard folders.

## What running it taught us

Every one of these was found by executing the workflow, not by reviewing it. Each is fixed
in the doc it affects.

**Packing to the budget ceiling was wrong.** The naive plan merged 226 shards and would
have left 52 parents above 90% of their word budget — satisfying the letter of the budget
while destroying the headroom it exists to protect. The convention says thresholds are
"capacity ceilings, not desired lengths or permission to fill the available space." Packing
to 80% instead merged 99. Yield deliberately cut by more than half.

**Per-shard packing over-reported yield 3.6×.** 811 shards fit their parent individually;
only 226 fit once packed together. Testing each shard against the parent alone counts the
same headroom repeatedly.

**"No machine consumer" is not "no obligation."** The sweep admitted the `created:` key
across 1,892 files, which no gate reads — but the Timestamp convention mandates it.
Withdrawn; removing it is a convention change with its own placement obligation.

**The duplication backlog was measured wrong.** Step 3b lacked the navigation exclusion
Step 7 already carries, so 6,809 annotated index-entry lines were counted as normative
prose. The real backlog is 46 groups over 101 files, not 65 over 129 — and the entries it
proposed removing are exactly the ones `governance-readme-completeness` requires.

**The biggest duplication group fails target-completeness.** Its nearest canonical home
governs *convention* documents with SHOULD; the duplicate binds *development practice*
documents with MUST. Different audience, different strength — a complete-the-target-first
item, never a deletion.

**Seven pre-existing broken links surfaced across the runs.** Already broken at HEAD, invisible because the
index gate inspects index references rather than shard bodies. Inlining made them visible.
All repaired; whole-tree broken links went 92 → 91 on run 3 with none newly broken.

**Retirement's only nomination was a false positive.** `RTK.md` has zero inbound Markdown
links but binds through an `@RTK.md` import — precisely the case 3c warns about when it
says absence of inbound links is not evidence.

## The scaffolding class, and the paraphrase question it settles

A fourth class was added for non-normative prose: **deletion-only**, never rewriting. Since
zero-obligation text cannot change the extracted inventory, Step 7's existing exact-match
diff *is* its admission test — no new instrument, no relaxed invariant.

Then it was measured. **Yield is ~5 safely removable lines in 163,867.** The surviving
candidates open with scope or exception content, which Step 7 protects. Its
heading-restatement form was withdrawn outright: implemented as word overlap with the
heading above, its highest-scoring matches were progressive-disclosure `See` links, because
a good `See` link necessarily echoes its heading. The form's best signal and the repo's most
protected construct are the same string shape.

That measurement answers the open question directly: since the *safe* class recovers almost
nothing, a paraphrase class would buy its marginal yield entirely by rewriting text that
carries obligations — the one operation propagation's semantic-preservation hard gate
forbids. **Paraphrase is dropped, on evidence rather than caution.**

## Invariants

- **Propagation is the sole writer.** Grooming discovers, ranks, and hands off.
- **The graph stays acyclic.** Grooming → gate → propagation; propagation calls neither.
  The gate's authorization was *amended* to name grooming rather than letting grooming
  invoke it on inferred authority.
- **Every run is gated.** Step 8 runs `rules-quality-gate` in `EFFECTIVE` mode on every run
  and every class. Mechanical preservation cannot tell whether what survived still coheres —
  a reduction can preserve every obligation and still move one into a parent that
  contradicts it.
- **Entry points are protected.** No class removes or hollows a `<name>.md` naming a
  workflow, principle, or convention without authorization naming it. Its `<name>/` shard
  folder stays ordinary corpus.
- **Prose is never rewritten.** Corrects a mis-citation along the way: anti-fix 2 forbids
  dense-prose *formatting*; the paraphrase prohibition is propagation's hard gate.

---

## Cost / benefit

**Part 1** removes 42,690 words and 19,290 lines of frontmatter, and adds ~480 words of
convention plus ~90 lines of validator and test code. The ongoing benefit is not the
one-time deletion — it is that the allow-list makes the deletion permanent. A minimum
schema would have let every removed key back in on the next document created by copy-paste.

**Part 2** is ~1,900 lines of new governance describing a workflow that has already removed
1,446 lines and is designed to run indefinitely. It pays for itself on the second sweep and
is the only surface that can measure the corpus as a whole.

Both parts are governance and test code; the BDD contract exempts tests from the cost/benefit
statement, so the figure above covers the non-test additions.

## Verification

All run on the current head.

| Gate | Result |
|---|---|
| `md frontmatter validate` | **PASSED — no findings** across 2,419 files |
| `governance word-budget validate` | 18 warnings, **0 FAIL**; `repo-governance/` over-target 9 → 8 |
| `governance readme-index validate` | **PASSED** — no orphan or ghost references |
| `parity manifest validate` | **current** |
| Broken internal links | **0**, before and after |
| `rhino-cli:test:quick` | **PASSED** — typecheck, lint, **751/751 unit**, specs structure, and all four coverage adapters (unit, integration, e2e, behaviour) |
| markdownlint | **0 errors** across 2,419 files |
| `gate run --surface=ci --group=governance` | **exit 0 — 12/12 PASS** |
| `gate run --surface=ci --group=markdown` | **exit 0 — 7/7 PASS**, including `md-links` |
| `rules-quality-gate` (EFFECTIVE) | **PASS_EFFECTIVE** |
| Obligation preservation | **0 real losses** (see Part 1) |

Six files in `repository-governance-architecture/` turned out to have **invalid YAML
frontmatter** — unquoted `:` inside plain scalars. They had never been validated, because
that subtree classified as `UnknownArea`. Extending the prefix surfaced them; they are fixed
here. That is the clearest evidence the prefix change was worth making on its own.

A second defect surfaced while re-reading `pr-leak-review.md` for the merge protocol:
reabsorbing its `success-criteria.md` shard removed the shard's index bullet from both
`## Contents` listings but left the bullet's *wrapped annotation line* behind, leaving each
index ending in a dangling sentence fragment glued to the previous entry. Neither
`readme-index validate` nor markdownlint can see this — the entry is still well-formed
Markdown and references no missing file. Both fragments are removed. A fence-aware sweep of
every index entry in `repo-governance/` for the same signature (a mid-entry sentence boundary
followed by a lowercase fragment) found **no other instance**; the 44 other matches are all
`vs.`/`e.g.` abbreviations. `ose-private` is unaffected — it never received the shard
reabsorptions, so it still carries the shard intact.

## The quality gate found three things, and they are fixed

`rules-quality-gate` was run read-only in `EFFECTIVE` mode over the affected rule and its
points of use. Eight of its nine semantic decisions passed outright. Decision 5 — one
canonical source, point-of-use links, no duplication — produced three findings: the canonical
workflow template, the criticality classification, and the YAML indentation convention each
restate part of the two-key rule to justify their own content, and none linked the convention
that owns it. A reader arriving at any of them had no route to the rule binding them.

All three were handed to `rules-propagation`, resolved with a point-of-use link each (no
surface gained a second copy of the rule), and both gate groups re-run clean. Ledger:
`local-tmp/repo-rules/rules-quality-gate__governance-frontmatter__ledger.md`.

## `ose-private` parity

`apps/rhino-cli` is declared byte-identical across the two repos, and this PR changes it.
The identical change is applied to `ose-private` in
[ose-private#166](https://github.com/wahidyankf/ose-private/pull/166) on its own
`worktree/rules-update` branch — **9/9 parity-scoped files byte-for-byte identical**, including
both repos' regenerated `parity-manifest.sha256`, which hash to the same value. That PR is
**merged** (`f6979f9016`) with all checks green and its own authenticated current-head
`pr-leak-review` pass.

`ose-private` also receives the `rules-grooming` workflow itself, so both repos carry the same
rule surface. It does **not** replay the 126 shard reabsorptions from this PR's three grooming
runs: those are specific to this repo's shard structure and budget headroom, most of those shards
do not exist there, and the workflow's own Step 5 requires per-class human authorization before
any reduction. That sweep is left for its own delivery.

`repo-governance/` is **not** in the byte-identity scope; the two corpora genuinely differ
(1,190 common paths, only 284 byte-identical at HEAD). What is held identical there is the
*transformation*, not the bytes: the same two-key rule, the same allow-list enforcement, the
same migration order, and the same preservation standard, with repo-accurate counts and
repo-accurate link targets.

## Known gap

The first grooming run wrote its 99 merges directly instead of routing them through
propagation, violating the sole-writer invariant the workflow itself asserts. The invariant
is now documented three ways; that run remains its counterexample.


wahidyankf (commented):
## Focused Leak-Only Review (`pr-leak-review`)

Scope: exactly three predicates — real secrets/private values, protected production/staging
environment properties belonging outside git, and real machine-specific absolute paths. No
other security or semantic concern was reviewed.

**Result: pass — no leak found.**

Inspected the complete aggregate base→head diff (2,554 files changed) and the PR title/body.
The change set is almost entirely `repo-governance/` Markdown prose plus a small F# change in
`apps/rhino-cli/` (`Governance.fs`, `Md.fs`, unit/integration/e2e test steps, two `.feature`
files) and a regenerated `apps/rhino-cli/parity-manifest.sha256`.

- **Secret / private value**: none found. The manifest's hex strings are SHA-256 content
  checksums of tracked source/spec files (expected regenerated output), not credentials. No
  PEM headers, known token-prefix formats, JWT-shaped strings, or embedded-credential URLs
  appear in any added line. One code example (`process.env.CLIENT_SECRET`) is illustrative
  Markdown-indentation-convention sample text, not a real value.
- **Protected production/staging environment property**: none found. No `.env`/`.env.example`
  file is touched, and no added line contains a database/service connection string, IP
  literal, or production/staging config value that policy requires in env or secret storage.
- **Machine-specific absolute path**: none found. No `/Users/…`, `/home/…`, Windows
  drive-letter, or other real machine-specific path was added. The tilde-prefixed paths that
  appear (e.g. `~/.claude/settings.json`, `~/projects/`, `source ~/.zshrc`) are generic
  documentation examples with no real username or hostname, consistent with the
  examples/placeholders exclusion.

Every other security and semantic concern (broad injection review, git-fixture isolation,
destructive-git-operation safety checks, etc.) is out of scope for this focused pass.

wahidyankf (commented):
## Focused Leak Review (pr-leak-review)

Exact leak-only pass for the current head. No broad security or semantic review, fixer pass, CI
wait, or retry was performed — those are out of scope for this workflow.

**Scope inspected**: aggregate base(`8c33c7ad`)→head(`7c3c5106`) diff (2,554 files touched, mostly
frontmatter-key deletions across `repo-governance/`), the new merge commit's own contribution
(`e3966bbf2`→`7c3c51069`, isolated to `hippo.lock`), and delivery-controlled PR title/body.

**Result**: no real secrets/private values, no protected production/staging properties committed
outside secret/env storage, and no real machine-specific absolute paths.

Notes:
- `hippo.lock`'s changed lines are a version bump plus release-asset content checksums and a
  public upstream commit hash — content digests, not secrets.
- `apps/rhino-cli/parity-manifest.sha256` entries are likewise content digests.
- The touched `secrets-and-env-standards/*` governance files only had frontmatter keys
  (`title`/`category`/`subcategory`/`tags`/`created`) removed; body content (including documented
  example values such as `conduct@example.com`) is unchanged in substance.

Every other security and semantic concern was out of scope for this focused pass.

PR #492, opened 2026-09-08
docs(governance): prove a run finished, and stop double-wrapping guarded npm scripts

wahidyankf:
## What

Two governance findings from `plans/in-progress/lms-init` Phase 1, propagated through
`repo-governance/workflows/rules/rules-propagation.md`.

**Rule 7 — Prove the Run Finished** (new). Trustworthy Measurement Rule 1 guards a command that
never *started*. Nothing guarded the other end: telling a finished run from one still working from
one that died. A backgrounded four-command wrapper produced empty stdout and an output file ending
in a plausible `PASSED` line. I read that as the wrapper having died; it had not, and was still
running 52 minutes later. Earlier in the same episode I made the opposite error. Without a terminal
marker those three states are one observation and a confident reading of it is a coin flip.

**Rule 1 amendment.** `${PIPESTATUS[0]}` is a `bash` array reference that expands to the empty
string under `zsh`, so an exit-code line prints a label with no number — the mechanism that made two
gate results here have no verified status. This belongs to the builtin-transform family Rule 1
already catalogues.

**Guarded Admission amendment.** The rule already said "never add another outer guard"; it did not
say *which* commands are already guarded. Every worked example in `AGENTS.md` and `README.md` wraps
`npm install` or `npm exec nx` — both correct, neither a package.json script. Verified against
`package.json`: 26 scripts invoke `./hippo run` directly, five delegate to one that does, and only
`prepare` is plain. So no `npm run <script>` here ever takes an outer guard.

## Enforcement dispositions

- Rule 1's `PIPESTATUS` half: **Covered** by the existing shellcheck gate, verified in both
  directions — violating input exits 1 with SC3028/SC3054, the conforming idiom exits 0.
- Rule 7's core and the double-wrap rule: **Unenforced by decision**. Both govern how an agent reads
  or types an ad-hoc uncommitted command that no repository gate observes; declaring a gate that
  cannot fire would be decoration. `git grep -nE 'hippo +run\b.*-- +npm run'` is recorded as the
  ready-made detector should the shape ever reach committed text.
- The pipe-loses-the-exit-code half was already stated in the propagation workflow's Step 7, and the
  double-wrap prohibition was already stated in Guarded Admission — so both are recorded as
  **violations of existing governance**, not as governance gaps. No new rule, no eviction.

## Cost/benefit

Three governance documents and two index entries; no code, no runtime surface, no dependency. The
benefit is that three distinct false readings that each cost real wall-clock in this plan become
mechanically checkable.


wahidyankf (commented):
## Focused PR Leak Review (exact leak-only mode)

Scope: exactly three categories — real secrets/credentials/private values, production/staging
properties that belong in environment/secret storage rather than tracked files, and real
machine-specific absolute filesystem paths. Every other security and semantic concern is out of
scope for this focused pass.

**Result: pass** — no leak found in any of the three categories across the aggregate base-to-head
diff (5 changed governance markdown files, all prose/examples) or the PR title/body.

PR #493, opened 2026-09-08
feat(repo): enable java projects across the quality surfaces

wahidyankf:
Delivery Unit 2 of [`plans/in-progress/lms-init/`](https://github.com/wahidyankf/ose-public/blob/lms-init/du2-java-enablement/plans/in-progress/lms-init/delivery.md). Follows #491 (DU1, doctor `extra-tools`).

## What this does

Java has no home in this repository yet. The behaviour-coverage validator cannot read a `.java` step definition, no formatter touches `.java`, and the PR quality gate has no job that would run a Java project's targets. The LMS backend arrives in DU3 and needs all three to exist first.

**Validator** — `scripts/behaviour-coverage.mjs` now scans `.java` sources for Cucumber-JVM `@Given`/`@When`/`@Then` bindings. Java resolves a step only against its own keyword and treats the annotation argument as a Cucumber expression, so the extractor sets `keywordSensitive: true` rather than reusing the TypeScript path — which would otherwise match `Given("...")` inside the annotation and look correct while proving nothing. The `specs/*.feature` quoted-literal scan the three language extractors shared is now one function.

**Formatting** — `scripts/format-java.sh` resolves each staged file's owning Gradle root and runs Spotless once per root, mirroring the Elixir wrapper's reason for existing (the formatter is invoked from a project root, not on bare file paths). Registered as the `format-java` / `format-verify-java` gate pair and emitted into the generated lint-staged block.

**CI** — a `setup-java` composite action following `setup-dotnet`'s self-hosted split, `has-java` detection in `detect`, a `java` quality-gate job, `tag:lang:java` excluded from the three sibling language jobs, and `java` added to the aggregate gate's `needs` so a Java failure cannot pass unnoticed.

**Also** — `java` and `springboot` admitted to the four-dimension tag vocabulary, three Java style guides plus their README, the `swe-programming-java` skill and `swe-java-dev` agent with regenerated OpenCode/Codex/`.agents` mirrors, `java` declared under `doctor.extra-tools` (which exercises DU1's stderr probe against a real JDK for the first time), and the `ose-lms-be` port rows.

## Expected CI behaviour

The **`Java quality gate` job must report `skipped`** on this PR. No Java project exists yet, so `has-java` is `false`; a job that *runs* here would mean the detection is wrong. DU3 is where it first executes.

## Cost and benefit

New code is one 66-line shell wrapper and one extractor function. Everything else is configuration or documentation. The benefit is that DU3's Java service lands into surfaces that already measure it, rather than arriving with its gates disabled and a follow-up PR to enable them. Tests are exempt from the cost accounting.

## Notes

- The rules-propagation outcome for this unit's three rules is recorded in the plan; R1 (the tag vocabulary) is dispositioned **Unenforced-by-decision** with a four-way proof that nothing today validates project tags — `repo-config.yml` has no tag key, no gate reads `project.json` tags, `nx.json` declares no tag constraints, and no F# validator reads them.
- Every Nx gate in this unit was re-run with `--skipNxCache`. A cache hit is not evidence that a gate ran, and the first run of each was a full cache hit.

https://claude.ai/code/session_015a9spDQ3PoMFNAoNLMT1VX


wahidyankf (commented):
## Focused PR Leak Review (exact leak-only mode)

Scope: exactly three categories — real secrets/credentials/private values, production/staging
properties that belong in environment/secret storage rather than tracked files, and real
machine-specific absolute filesystem paths. Every other security and semantic concern is out of
scope for this focused pass.

**Result: fail** — examined the full base-to-head diff (37 files) including the new
`plans/in-progress/lms-init/evidence/*.txt` fixtures and `repo-config.yml`'s new
`doctor.extra-tools` entry; found no secrets or protected environment properties, but the failing
Node test-runner capture committed at `evidence/du2-red-validator.txt` was not host-path
sanitized and carries real machine-specific absolute paths from the authoring machine.

### Findings

1. **Category**: machine-specific-absolute-path
   **Location**: `plans/in-progress/lms-init/evidence/du2-red-validator.txt:65`, `:92`, `:109`
   **Remediation**: Replace the captured local absolute source path in these Node test-runner
   stack-trace lines (`file:///Users/...`) with a repository-relative path or a sanitized
   placeholder, per this PR's own `delivery.md` history of stripping `/Users/...` paths from
   evidence files before commit.

2. **Category**: machine-specific-absolute-path
   **Location**: `plans/in-progress/lms-init/evidence/du2-red-validator.txt:86`, `:87`, `:88`, `:97`
   **Remediation**: Replace the captured OS per-user temp-directory path in these assertion-diff
   lines with a sanitized placeholder (e.g. `<tmpdir>/...`), per this repo's own established
   `/var/folders/...` sanitization rule (documented as a prior real leak and fix in this same PR's
   `delivery.md`).

wahidyankf (commented):
## Focused PR Leak Review (exact leak-only mode)

Scope: exactly three categories — real secrets/credentials/private values, production/staging
properties that belong in environment/secret storage rather than tracked files, and real
machine-specific absolute filesystem paths. Every other security and semantic concern is out of
scope for this focused pass.

**Result: pass** — re-review at the fixed head `74a091c4f`: the aggregate base..head diff, the PR
title/body, and all 18 files under `plans/in-progress/lms-init/evidence/` were re-examined (not
only the previously-dirty `du2-red-validator.txt`), and the prior 7 `machine_specific_absolute_path`
findings are confirmed fixed — `du2-red-validator.txt` now uses the `<public-worktree>` and
`<tmpdir>` placeholder tokens the other evidence files already use, with the file's own trailer
recording the substitution and a zero-match re-scan across the four host-path families; no new
secrets, protected-environment properties, or real machine-specific absolute paths were found
anywhere else in the diff (CI workflow, `repo-config.yml`, `scripts/format-java.sh`, the new Java
docs/agent/skill files, `delivery.md`, `learnings.md`).

Two items were explicitly examined and dismissed as non-leaks, not overlooked:
- `delivery.md` and `learnings.md` prose describing the leak class uses already-anonymized
  `<user-hash>`/`<session-hash>` placeholders, not real captured paths.
- The committed lint output's relative source path `components/home/entry-item.tsx` has no
  leading slash and is not a host path (the substring `home/` there is a component directory, not
  `/home/<user>`).

A third item was also checked and dismissed for the same reason: `du2-doctor-java.txt`'s own
sanitizer note lists the prefixes it scanned for (`/Users/, /home/, /private/, /opt/, /var/`) as
plain text describing the check performed — it is not an occurrence of any of those paths.

PR #497, opened 2026-09-08
feat(specs): add the islamic-be behaviour corpus and OpenAPI contract

wahidyankf:
## Why

DU2 of `plans/in-progress/islamic-be-init/`. Establishes the specification corpus that `islamic-be`
(DU3) and `islamic-be-e2e` (DU4) will implement, before either project exists — so the Gherkin is
written against requirements rather than reverse-engineered from an implementation.

## What

- `specs/apps/islamic/` — product README and overview, registered in the `specs/apps/` annotated index
- `specs/apps/islamic/be/` — README plus a C4 architecture doc (context + container diagrams)
- 8 Gherkin scenarios across `behaviours/health/` (3) and `behaviours/config/` (5)
- An OpenAPI 3.1 contract split into `paths/` and `schemas/` fragments, with `.spectral.yaml`
  copied byte-identically from the `ose-be` contract
- The `islamic-contracts` Nx project, tagged `["type:lib","domain:islamic"]` with `namedInputs.specs`
- Five sanitized evidence captures from the Phase 1 and Phase 2 gate runs

## Cost / benefit of new code

No production code. The added cost is one Nx project (8 targets, all spec-linting) and 21 spec files
that must be maintained alongside the service. The benefit is that DU3's implementation has a fixed,
diff-verified acceptance target: the scenario names and step lines are byte-identical to `prd.md`
US-1 and US-3, checked mechanically, so the corpus cannot silently drift from the requirements.

## Decisions worth review

- **The config scenarios carry an E2E exemption**, not only the Integration one the health scenarios
  take. A caller can observe *that* the service listens but not *which source supplied the port*, and
  a process that refuses to start on a malformed port exposes no public boundary at all. Both name
  `islamic-be:test:unit` as the alternative proof. Unit is never exempt.
- **`islamic-contracts` declares tags; its `ose-contracts` sibling declares none.** An
  exclusion-based CI selector treats "tag absent" and "tag unknown" identically — both fail open —
  so the tags are deliberate rather than cosmetic.
- **The corpus names `apps/islamic-be` in prose instead of linking it.** `md-links` is an
  `all-file-type` gate, so a forward link to a project DU3 has not created yet would fail the whole
  tree. Those links are explicit steps in DU3 and DU4.
- **`contracts/generated/` uses a glob-plus-negation ignore**, keeping a hand-written README while
  ignoring bundles. `ose-be` tracks its bundles; `ose-lms-be` ignores the directory wholesale with no
  README. Three shapes now coexist; the divergence is recorded in `learnings.md`, not resolved here.

## Verification

All with `--skip-nx-cache`:

- `islamic-contracts:lint` — "No results with a severity of 'error' found!"
- `islamic-contracts:test:quick` — exit 0
- `islamic-contracts:specs:structure-validation` — exit 0
- `specs structure validate` — 0 findings for "islamic"
- Scenario/step diff against `prd.md` US-1 and US-3 — byte-identical

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,6 +178,11 @@ coverage.xml
 # Bundled OpenAPI spec + docs
 specs/apps/a-demo/contracts/generated/
 specs/apps/ose/lms-be/contracts/generated/
+# The islamic contract keeps a hand-written README in its generated/ folder, so the folder
+# itself is ignored by glob and the README negated back in. The two siblings above ignore
+# the whole directory instead; theirs hold nothing but build output.
+specs/apps/islamic/be/contracts/generated/*
+!specs/apps/islamic/be/contracts/generated/README.md
 
 # MemSearch vector index (local user-specific data)
 .memsearch/
--- a/plans/in-progress/islamic-be-init/delivery.md
+++ b/plans/in-progress/islamic-be-init/delivery.md
@@ -689,12 +689,54 @@ platform linux`. Root cause: the `gate` matrix job provisions the **union** of a
 
 > All checks below must pass before starting Phase 3. Phase 2 (DU2) may proceed in parallel.
 
-- [ ] [AI] `npm run test:validators` — exits zero with the new Go extractor cases present
-- [ ] [AI] `npm exec nx -- run rhino-cli:repo-config:validation` — exits zero
-- [ ] [AI] `npm exec nx -- run-many -t test:quick --projects=ose-be,ose-be-e2e` — exits zero, proving no regression to existing lanes
-- [ ] [AI] Confirm the merged `pr-quality-gate.yml` excludes `tag:lang:go` in the `typescript`, `dotnet`, `flutter`, **and** `java` jobs — acceptance: `rtk grep -c 'tag:lang:go' .github/workflows/pr-quality-gate.yml` reports 5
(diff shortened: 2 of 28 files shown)


wahidyankf (commented):
## Focused Leak Review (`pr-leak-review`, exact leak-only mode)

Scope: exactly three categories — secret-or-private-value, protected-environment-property,
machine-specific-absolute-path — across the aggregate diff and delivery-controlled PR
title/body. Every other security and semantic concern was out of scope for this pass.

**Result: pass — no findings.**

Checked in particular:
- All five `plans/in-progress/islamic-be-init/evidence/*.txt` captures, including the
  `du1-gate-ose-be.txt` sanitization that replaced the absolute worktree prefix with
  `<WORKTREE_ROOT>` — no residual `/Users/`, `/home/`, `/private/tmp`, `/var/folders`, Windows
  drive path, hostname, username, or secret/token pattern remained in any of the five.
- `specs/apps/islamic/be/contracts/openapi.yaml` `servers:` entry — `http://localhost:8402`,
  a portable local-dev placeholder, not a real environment hostname.
- Remaining spec/README/feature files touched by this PR — no absolute paths or credentials.
- PR title and body — no leak-pattern matches.

PR #499, opened 2026-09-08
feat(islamic-be-e2e): drive the health scenarios through real HTTP

wahidyankf:
## Why

DU4 of `plans/in-progress/islamic-be-init`. The behaviour corpus has named an E2E project since DU2
and DU3 deferred the `e2e` adapter to whichever unit could make it resolve. This is that unit:
`apps/islamic-be-e2e` exists, the three `health/` scenarios run against a real `islamic-be` process
over HTTP, and `apps/islamic-be` declares its `e2e` adapter and `test:coverage:e2e` target.

## Cost / benefit of the new code

**Cost.** One new Nx project (~230 lines of TypeScript plus config), one new shell runner, and two
`@playwright/test` / `playwright-bdd` dependencies already pinned at the same versions by nine
sibling suites. `test:coverage` on `islamic-be` gains a third validator.

**Benefit.** The health contract is now proven at the boundary a consumer actually uses, not only
in-process. Without it, `islamic-be`'s `health/` scenarios have Unit proof only, and their
`@integration-exempt` tags name `islamic-be-e2e:test:e2e` as the alternative proof — a promise this
PR makes true.

## The suite starts the process it observes

`ose-be-e2e`'s harness starts the service only if nothing already answers on the port. Copied
verbatim, that made this suite pass against a service it never launched: a stale `islamic-be` from
the DU3 manual health check still held 8402, and a health handler deliberately broken to return
`"ok"` instead of `"healthy"` still reported three green scenarios.

`ensureBackendStarted` now reuses only a process this harness started and still owns, and treats any
foreign listener as a hard error. `evidence/phase-4-e2e.txt` records three runs: the passing one,
the mutation proving it can fail, and the earlier one that passed against day-old code.

## Deliberate divergences from `apps/ose-be-e2e`

| Divergence | Reason |
| --- | --- |
| `retries: 0` rather than `CI ? 2 : 0` | AGENTS.md forbids retrying a flaky test instead of fixing its root cause |
| No `e2e-coverage-baseline.json` | `allowedUnbound` is read by nothing in this repository; the live mechanism is the `@e2e-exempt` tag, which both `playwright.config.ts` and `behaviour-coverage.mjs` already read |
| No `docker-compose` in `run-e2e.sh` | The service owns no database, broker, or disk |
| Reuse only our own process | See above |
| No `@vitejs/plugin-react` dependency | A React plugin has no place in a headless API suite |

## Verification

- `nx run islamic-be-e2e:test:e2e` — 3 passed against a real process; fails as expected under a
  mutated health handler
- `nx run islamic-be-e2e:test:quick` — exits zero (`--skip-nx-cache`)
- `nx run-many -t test:quick --projects=islamic-be,islamic-be-e2e,islamic-contracts` — exits zero,
  Go line coverage still 100.0% against a 99% floor
- `rhino-bin.sh md links validate --exclude plans/done` — all links valid
- Deleting one `@e2e-exempt` tag makes `test:coverage:e2e` fail with `undefined E2E binding`, then
  passes again once restored

## Skip

No `test:unit` or `test:integration` target on the E2E project — both belong to `islamic-be`, and
the anti-echo convention forbids a naming-symmetry stub.

### DIFF
--- a/apps/islamic-be-e2e/.gitignore
+++ b/apps/islamic-be-e2e/.gitignore
@@ -1,0 +1,5 @@
+test-results/
+playwright-report/
+.features-gen/
+node_modules/
+dist/
--- a/apps/islamic-be-e2e/README.md
+++ b/apps/islamic-be-e2e/README.md
@@ -1,0 +1,69 @@
+# islamic-be-e2e
+
+End-to-end tests for [`islamic-be`](../islamic-be/README.md), driven through its public HTTP
+boundary. Playwright-BDD executes the same behaviour examples that describe the service — no
+browser involved. 🧪
+
+## Run locally
+
+```bash
+# Install test dependencies once on this machine
+./hippo run --class transactional --disk-path . -- npm exec nx -- run islamic-be-e2e:install
+
+# Build the service and run the scenarios against a real process
+./hippo run --class service --disk-path . -- npm exec nx -- run islamic-be-e2e:test:e2e
+```
+
+`test:e2e` runs [`apps/islamic-be/scripts/run-e2e.sh`](../islamic-be/scripts/run-e2e.sh), which
+builds the binary through Nx — so `codegen` runs first and the binary matches the current contract —
+and then hands process lifecycle to the harness. There is no `docker-compose` stack: the service
+owns no database, broker, or disk, so there is nothing to stand up.
+
+Use `islamic-be-e2e:test:e2e:ui` for Playwright's UI and `islamic-be-e2e:test:e2e:report` for the
+most recent report.
+
+## The suite starts the process it observes
+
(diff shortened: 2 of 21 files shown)


wahidyankf (commented):
## Focused PR Leak Review

Exact leak-only mode per `pr-leak-review`. Scope: real secrets/private values, protected
production/staging environment properties that belong outside git, and real machine-specific
pass.

**Result: pass** — no leak found in any of the three categories across the full base-to-head
diff, including the two captured-terminal-output evidence files
(`plans/in-progress/islamic-be-init/evidence/du3-ci-routing.txt` and
`plans/in-progress/islamic-be-init/evidence/phase-4-e2e.txt`), which were inspected specifically
for machine-specific absolute paths and found to contain only repository-relative paths, public
CI run/PR URLs, and public commit SHAs.

PR #500, opened 2026-09-08
feat(rhino-cli): scan Go source for environment reads

wahidyankf:
Teaches `env validate` the `go` language, so an `env-contract:` surface may declare `lang: go` instead of being rejected with `unsupported lang: go`.

**Paired with `wahidyankf/ose-private` — this PR is not independently shippable.** `Env.fs` and `env-validate-app-drift.feature` are inside `apps/rhino-cli/parity-manifest.sha256`, which both repositories enforce. Merging one alone makes the other red on a gate it did nothing to break, so both must merge in the same session. Counterpart PR linked in a follow-up comment.

## What it does

`scanGoReads` mirrors `scanFsharpReads`, with two documented divergences:

1. **Scans the module root, not `root/src`** — a Go module has no `src` directory; its packages sit directly beneath `go.mod`.
2. **Three regexes, not two.** Alongside direct `os.Getenv("K")` and `os.LookupEnv("K")` calls, it matches the injected-reader form `os.LookupEnv, "K"`.

`_test.go` files are skipped for the same reason `scanTsReads` skips `.test.`/`.spec.` — tests name keys to build fixtures, not to read configuration. `generated-contracts/` is skipped because generated transport types are not an authored environment contract.

## Why the third regex is load-bearing

Go lets a reader be passed as a function value, and `apps/islamic-be` does exactly that: `main.go` hands `os.LookupEnv` to a pure resolver and keeps the key in a `const`, so no literal ever reaches a call site. Verified empirically — a two-regex scanner returns **zero** keys for that module.

Left unfixed, that surfaces later as a genuinely-read variable reported as declared-but-unread, and the obvious remedy at that moment would be an `allowlist:` entry — silencing a drift detector on a variable that really is read.

This is not a new convention. `Env.fs` already carries the F# analogue for the identical dependency-injection shape (`fsharpReaderWrapperRegex`, matching `ose-be`'s injected `readEnvironment "OSE_BE_PORT"`). Injecting the reader is good design — it is what lets the resolver be Unit-tested without touching the OS — so the scanner meets the codebase where good design puts it.

## Proof

Bound at all three adapters for one scenario: Unit through the pure `validateAppKeys`, Integration against real temp directories, E2E across the published CLI boundary.

Both the injected-reader regex and the `| "go" ->` dispatch were mutation-tested — each mutation fails **exactly** the one test that targets it, and restoring returns to green. `rhino-cli:test:quick` exits zero in both repositories (757 unit tests).

## Cost/benefit

~60 lines of new scanner plus its proof, against a drift detector that would otherwise have to be switched off for every Go app via an allowlist entry. Tests are exempt from the cost side.


wahidyankf:
Paired counterpart: wahidyankf/ose-private#169. Both carry the byte-identical `Env.fs`, feature file, and regenerated `parity-manifest.sha256`; they must merge in the same session.

wahidyankf (commented):
## Focused PR Leak Review

Exact leak-only mode per `pr-leak-review`. Scope: real secrets/private values, protected
production/staging environment properties that belong outside git, and real machine-specific
pass.

**Result: pass** — no leak found in any of the three categories across the full base-to-head
diff, including the four captured-terminal-output evidence files
(`plans/in-progress/islamic-be-init/evidence/du5-red.txt`,
`plans/in-progress/islamic-be-init/evidence/du5-manifest-diff.txt`,
`plans/in-progress/islamic-be-init/evidence/du5-parity-preflight.txt`, and
`plans/in-progress/islamic-be-init/evidence/du5-preflight-diff.txt`), which were inspected line
by line specifically for machine-specific absolute paths and found to contain only
repository-relative paths, worktree/branch labels, public CI run and PR URLs, and public commit
SHAs.

PR #501, opened 2026-09-08
feat(islamic-be): register the Go env-contract surface and the app registry

wahidyankf:
Final delivery unit of the `islamic-be-init` plan. Puts the service on the surfaces that already describe every other app: the `env-contract:` registry, the port/domain tables, the monorepo app list, and the apps product map.

## The registration surfaced a real defect, not a paperwork step

`scanGoReads` (added in #500) finds **nothing** in this module as it stood. `main.go` passed `os.LookupEnv` as a function value and kept the key in a package const, so no key literal ever reached a call site.

The surface would therefore have validated only by adding `ISLAMIC_BE_PORT` to `allowlist:` — silencing a drift detector on a variable that genuinely *is* read. That is the failure mode the allowlist exists to avoid, not to enable.

Instead, `ResolvePort` now takes the variable name rather than hardcoding it, and `cmd/islamic-be` names `ISLAMIC_BE_PORT` beside the reader it injects. This is the shape `ose-be`'s `PortResolver` already uses (`readEnvironment "OSE_BE_PORT"` at its composition root), and it is better design independent of the scanner — the pure resolver stops knowing which app it serves.

`allowlist:` is empty and stays empty.

## Proof the registration is not vacuous

Reverting the key back into a const reproduces the drift; restoring it clears the finding:

```
MUTATION  -> DRIFT  apps/islamic-be  declared-but-unread  ISLAMIC_BE_PORT
RESTORED  -> env validate: no drift detected across all surfaces
```

Full transcript in `plans/in-progress/islamic-be-init/evidence/du6-env-validation.txt`.

## Gates

- `nx run rhino-cli:env:validation` — no drift across all surfaces
- `nx run-many -t test:quick -p islamic-be islamic-be-e2e islamic-contracts` — all green
- Coverage unchanged at 100% for `config`, `health`, `router` (the signature change adds no branch)
- `markdownlint-cli2` — 0 errors; no stray `.env` files

## Incidental fix

`repo-config.yml`'s `env-contract:` comment documented only `"rust"` or `"typescript"`. It had already fallen behind `fsharp`; it now names `go` too.

## Cost/benefit

The app-source change is net-neutral in size and removes a hardcoded name from a pure function. The registry edits are documentation that keeps the app discoverable. Tests are exempt from the cost side.

### DIFF
--- a/apps/README.md
+++ b/apps/README.md
@@ -11,6 +11,7 @@
 | AyoKoding                                    | [ayokoding-www](./ayokoding-www/README.md)                                                                  | Educational content platform.                                            |
 | OrganicLever's public presence               | [organiclever-www](./organiclever-www/README.md)                                                            | Marketing website for the OrganicLever productivity platform.            |
 | OrganicLever's life journal                  | [organiclever-app-web](./organiclever-app-web/README.md) and [organiclever-be](./organiclever-be/README.md) | Local-first journal and productivity tracker, with its REST API backend. |
+| Islamic tools                                | [islamic-be](./islamic-be/README.md)                                                                        | REST API backend for the Islamic tooling surface.                        |
 
 ## Tools
 
@@ -30,6 +31,7 @@ End-to-end projects keep browser and API behaviour separate from the application
 | AyoKoding                   | [ayokoding-www-fe-e2e](./ayokoding-www-fe-e2e/)                  | [ayokoding-www-be-e2e](./ayokoding-www-be-e2e/)         |
 | OrganicLever public website | [organiclever-www-fe-e2e](./organiclever-www-fe-e2e/README.md)   | Not applicable; the site has no backend public boundary |
 | OrganicLever product        | [organiclever-app-web-e2e](./organiclever-app-web-e2e/README.md) | [organiclever-be-e2e](./organiclever-be-e2e/README.md)  |
+| Islamic tools               | Not applicable; the service has no browser surface               | [islamic-be-e2e](./islamic-be-e2e/README.md)            |
 
 ## Work with an app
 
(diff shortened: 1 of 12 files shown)


wahidyankf (commented):
## Focused leak review (`pr-leak-review`)

Scope: exactly three categories — real secrets/private values, protected production/staging
environment properties that belong outside git, and real machine-specific absolute filesystem
paths. Every other security and semantic concern is out of scope for this pass and was not
reviewed.

**Result: pass — no leak found in any of the three categories.**

Inspected the full base-to-head diff at the pinned head, plus the PR title/body, with particular
attention to the files most likely to carry a leak:

- `plans/in-progress/islamic-be-init/evidence/du5-parity-audit.txt` and
  `plans/in-progress/islamic-be-init/evidence/du6-env-validation.txt` (new captured terminal
  output) — read line by line at the pinned head via the Contents API. All paths shown are
  repository-relative (`apps/rhino-cli/src`, `apps/islamic-be`, `specs/apps/rhino/cli/behaviours`,
  etc.) or truncated Go package paths (`.../internal/config`); no `/Users/...`, `/home/...`, or
  other machine-specific absolute path appears. Git commit SHAs, GitHub Actions run numbers, and a
  public Actions run URL are present but are public identifiers, not secrets.
- `repo-config.yml`'s new `env-contract:` surface entry for `apps/islamic-be` declares only
  `root`/`kind`/`lang`/`allowlist: []` — structure and names, no values.
- `apps/islamic-be/cmd/islamic-be/main.go`'s literal `"ISLAMIC_BE_PORT"` is a variable name, not a
  value.
- `docs/reference/web-sites.md`'s new port (8402) and variable name are public, non-secret
  configuration identifiers.

No other candidate values were found anywhere in the diff.

---

PR #509, opened 2026-09-09
docs(governance): expand git clean-up into dev artifact clean-up

wahidyankf:
## What

Renames the teardown workflow `git-clean-up` → `dev-artifact-clean-up` and widens the gate from Git artifacts alone to the development artifacts a session creates.

## Why the name changed

The old name already understated the workflow: its step 5 purged build output, which is not Git at all.

## Changes

- **Docker as a fourth artifact class** — scoped to Compose stacks a session started and images it built locally. Pulled base images, named data volumes, and stacks another session owns are retained. Machine-wide prunes (`docker system prune`, `image prune -a`, `volume prune`) stay out of the gate for the same reason `git gc` does.
- **Containers come down before the worktree** — `infra/dev/ose-app` and `infra/dev/organiclever-app` bind-mount the worktree read-write, so a running stack holds the directory the next step removes.
- **Secrets are out of scope everywhere** — `.env*` files and local infrastructure state exist nowhere else in the repository, so deleting one is unrecoverable from git.
- **Primary checkout is bounded** — only regenerable build output is removable there. This supersedes the previous worktrees-only purge scope, recorded in the propagation manifest.
- **Plans reference, not restate** — the delivery-checklist DAG's terminal node links the workflow instead of inlining its steps.
- **Enforcement** — `plan-verifying-execution` gains HIGH/CRITICAL lines for the new bounds.

## Design note

An earlier draft proposed per-worktree Compose project names as a prerequisite. Withdrawn: every dev stack binds fixed host ports, so two worktrees cannot run the same stack concurrently regardless of naming. The remedy is a proof-of-ownership rule, not compose-file edits. This delivery is governance-only.

## Cost / benefit

Cost: five governance files touched plus one new convention document, and one skill trimmed to stay under its 750-word fail limit.
Benefit: the gate now names the artifacts it actually leaves behind, and cannot delete an operator's only copy of a secret.

## Verification

`governance` gate group exit 0 (10/10). `markdown` group 6/7 — the one failure is a pre-existing `markdownlint-cli2: command not found` PATH defect; content verified green via `npm run lint:md` (0 errors, 7,640 files). Word budget exit 0, no FAIL. Prettier clean. Bindings mirror byte-identical.

https://claude.ai/code/session_01NmvNFotgLaEKQW1fMdY1FG


wahidyankf (commented):
## PR Leak Review — pass

Reviewed the whole diff at head `89397058619aba1ecf461ebd1dc652c1cb1af10f`.

| Category | Count |
| --- | --- |
| Real secrets, credentials, or access-granting values | 0 |
| Properties belonging in environment or secret storage | 0 |
| Real machine-specific absolute paths or host evidence | 0 |

The diff is sixteen Markdown files. The literal `.env*` occurrences are a **pattern name inside a prohibition** — the rule being added is precisely that such files are never deleted — not a value. Every path in the diff is repository-relative.

**Record posted after merge.** The scan itself was run against this exact head before merging and returned zero in all three categories, but the record was not posted first, so the ordering of the merge precondition was not met even though its substance was. Noting it here rather than leaving the gap silent.

PR #510, opened 2026-09-09
docs(governance): reconcile the task list when new direction arrives

wahidyankf:
Standard 4 covers work the agent **discovers** on the way. Nothing covered direction **arriving from outside** while a list is already open.

The two differ in a way that matters. Discovery is found by the agent and only ever *adds* to the list. Direction is given to it and can *invalidate* items already on it. Only the second can make an open item wrong — which is why appending the new work without re-reading the old is insufficient, and why this is not a restatement of Standard 4.

Standard 8 states the ordering: new, follow-on, or changed direction reconciles against every open item **before** any work responds to it. The reconciliation records what the direction invalidates, supersedes, leaves untouched, and creates. An item the direction kills is closed with that reason rather than left to be discovered later as apparently abandoned work.

Acting first and updating afterwards produces a list describing the task as it was originally *requested* rather than as it is being *performed* — the same stale state Standards 2 and 3 exist to prevent, arriving by a different route. The reconciliation is also the moment a contradiction between old and new direction becomes visible; carrying both forward silently resolves it by accident, and the resolution is never recorded as the decision it was.

## Verification

`npm run lint:md` — 7641 files, 0 errors. `npm run format:md:check` — exit 0.

Documentation only. No source, no specification, no behavior change.


wahidyankf (commented):
## Leak review — pass

Read the complete diff at `99b45f81f24decd6bac1b4663cb5d4b3303606ea`. Three Markdown files under `repo-governance/`: one new standard, and two index entries registering it.

- **Category 1 — access-granting values (0).** The diff introduces no value of any kind. Every hunk is prose or an index bullet.
- **Category 2 — properties belonging in environment or protected storage (0).** Nothing configurable is introduced. No endpoint, port, hostname, account, credential, or tunable appears.
- **Category 3 — machine-specific absolute paths or real host evidence (0).** Every path in the diff is repository-relative and resolves inside `repo-governance/development/practice/`. No captured host state, timing, user, or machine name.

The added prose is generic governance language describing task-list behavior, with no repository-external reference and nothing drawn from this machine.

PR #513, opened 2026-09-09
chore(hippo): pin the consumer to v0.5.3

wahidyankf:
## What

Bumps the pinned HIPPO release from `v0.5.2` to `v0.5.3` in `hippo.lock`, with all four platform digests.

## Why

`v0.5.3` fixes duplicate environment-key resolution in the guard. HIPPO read the **first** occurrence of a variable, while `os/exec` — and HIPPO's own writer — keep the **last**. The guard could therefore admit against one value while the child it launched read another.

The trigger is ordinary: `append(os.Environ(), "HIPPO_SESSION="+token)` leaves a duplicate, so an ambient session from an outer guard shadowed the caller's explicit override and produced a spurious exit `75` instead of session inheritance. The same reader backs `--concurrency-env` mapping, the reservation clamp, and launcher identity detection — all of which this repository exercises through its guarded `package.json` scripts.

## Verification

- Digests recomputed locally from the downloaded release assets rather than copied from the published `checksums.txt`; they match byte for byte.
- The consumer bootstrap was exercised in this worktree against the new pin: it downloaded `v0.5.3`, verified the checksum, and reported `{"version":"v0.5.3","commit":"b13567145a86942c76a32ea3371d5f75a817e3d9"}`.

## Scope

`hippo.lock` only. The `v0.5.2` strings remaining in `apps/ose-www/content/updates/2026-09-09-week-42-*.md` are a dated historical record of that week's state and are deliberately left unchanged.


wahidyankf:
## Leak review — pass

Read the complete diff at `f0711ec770ec51e5dce57270d716812a1ba0f37a`. One file, `hippo.lock`, six lines changed.

- **Category 1 — access-granting values (0).** The changed values are a semantic version string, a 40-character Git object ID, and four SHA-256 digests of publicly published release archives. A digest of a public artifact grants no access; it is a verification input, and its whole purpose is to be published. The commit ID names a public commit on `wahidyankf/hippo`.
- **Category 2 — properties belonging in environment or protected storage (0).** No endpoint, port, credential, or tunable is introduced. `hippo.lock` is a pin file and is meant to be tracked.
- **Category 3 — machine-specific absolute paths or real host evidence (0).** No path of any kind appears in the diff. No hostname, user, address, or captured host state.

The digests were recomputed from the downloaded public release assets and match `checksums.txt` published on the `v0.5.3` release, so nothing in the diff is derived from this workstation.

PR #514, opened 2026-09-09
docs(governance): fix the ordinal width at three digits

wahidyankf:
## Outcome

The Ordinal Filename Prefixes Convention now states a **width**: an ordinal that is kept is zero-padded to three digits. The eight files that pass the rule are renamed accordingly.

## Why

The convention already said *when* a governed filename may keep a leading number, and *which* number it must be. It never said how wide. The eight passing files carried two digits; the sibling repositories in this workspace number the ordered companions of a split plan document with three, and each repository stated only its own half. A reader crossing repositories had to remember which width applied where, and one question with two answers is a rule nobody can quote back.

Three digits wins because it is the width the larger population already carries, so aligning moves the smaller set. The convention now says that out loud, which is what stops a later reader from shortening it back.

## Scope

**In scope.** The width sentence in `ordinal-filename-prefixes.md`; the eight renames; the sixteen link targets in `repo-governance/` that pointed at them; the convention's own worked examples, which quote the names being renamed and would otherwise document a state the tree no longer has.

Also in scope, and worth stating rather than smuggling: **five Mermaid labels** in the two renamed `003-user-review.md` steps are shortened to the declared 20-character maximum. The violations predate this change. Renaming the files is what brought them into a changed-path gate, and the repair belongs at the cause rather than behind a new exclusion.

**Deliberately not in scope.**

- **Widening the rule.** This states a width; it does not license an ordinal anywhere one was banned. A filename that fails either of the two questions still *sheds* its ordinal rather than gaining a digit, so word-budget shards stay plain-named with their order in the parent index. Turning this into permission to number shards would undo the decision the convention was written to record.
- **Archived plans.** `plans/done/**` keeps its two-digit references. Those are records of what was delivered; editing a record to match a rule written afterwards falsifies it. `local-tmp/` and captured audit evidence are untouched for the same reason.
- **The verification regex.** It stays `[0-9]+`, never `[0-9][0-9][0-9]`. A glob pinned to the prescribed width reports a clean tree exactly when the tree is wrong — the one case the check exists for. The convention now says so where the regex appears.

## Where to start reading

`repo-governance/conventions/structure/ordinal-filename-prefixes.md` — the new `## Width` section, then the worked-cases table. Everything else follows from it.

**Paths to skip.** `.codex/agents/**` — 89 regenerated agent mirrors, emitted by the `harness-bindings-generate` pre-commit gate, no hand edits. The eight renames themselves are pure `R` entries with no content change; `git log --follow` reads them as moves.

## Verified

Pre-commit gate: `prettier --check` clean across `repo-governance/**/*.md`; `markdownlint-cli2` 0 errors; `md mermaid validate` clean after the five label fixes; `harness-bindings-generate` and `commitlint` pass. Pre-push gate ran on the exact head being reviewed.

The keep-clause's own non-emptiness check still returns 8, so the rule has not collapsed into the ban it replaces.

## Risk and rollback

The residual risk is a stale link: something outside `repo-governance/` pointing at a two-digit name. Verified absent — grep across the working tree returns matches only in `plans/done/`, `local-tmp/`, and captured evidence, all of them backticked prose rather than links. Rollback is reverting one commit; the renames revert with it.

## Why this is one natural seam

One rule and every consequence of stating it. The width sentence and the eight renames cannot land apart: a rule its own repository violates is worse than no rule, and renames without the stated rule are unexplained churn. The link repoints and the convention's own examples are the same obligation — they quote the names being renamed. The Mermaid fixes are here because this change is what makes the gate read those files; deferring them would mean landing a branch whose own gate is red.

No unrelated purpose is included.

## Why `main` stays production-deployable

Nothing user-reachable moves. No application code, no build input, no configuration, no route, no content under `apps/` is touched — the diff is governance Markdown plus a regenerated agent mirror. There is no feature flag to describe because there is no behaviour to gate. Build, test, and lint proof is the pre-commit and pre-push gate above; rollback is a single revert.

https://claude.ai/code/session_014ZAMZKyV4bzuWoHiua5ag2


wahidyankf (commented):
## Focused PR Leak Review

Leak-only inspection of the whole diff at the pinned head. The diff is governance
Markdown, eight pure renames, and a regenerated agent mirror.

- **Real secrets or private values** — none. No credential, token, key, or
  connection string appears in any hunk.
- **Protected production or staging properties** — none. No environment file,
  no `repo-config.yml` value, and no deployment input is touched.
- **Real machine-specific absolute paths** — none. Every path in the diff is
  repository-relative. The only new path-like tokens are the three-digit example
  filenames the convention quotes.

Result: `pass`.

PR #516, opened 2026-09-13
chore: replace private-shaped values in examples, plans, and dev defaults

wahidyankf:
## Outcome and Why

A whole-tree scan with the public-safety gate's shape set flags 2,656 lines in this repository. None is a live secret,
but every one blocks wiring that gate into this repository's hooks and CI, because the gate blocks on any finding. The
flagged lines fall into five groups:

- private-range addresses and internal-style hostnames in tutorials and governance examples;
- home-directory paths carrying a real account name, mostly in completed plans;
- literal database credentials in documentation examples and local development defaults;
- a traceroute capture with private hops;
- text that only resembled those shapes.

This PR clears 2,378 of them. A rescan of the result finds 278, every one already present before this PR. Of those, 25
need a follow-up edit or a validator fix, and 253 need a rule decision rather than an edit.

## Scope

- **In scope:**
  - Maintainer home paths in plans, tutorials, rules, and four generated tutorial artefacts.
  - Example and local-default credentials.
  - Teaching addresses and hostnames that do not depend on a private range.
  - Captured output and prose derived from those addresses.
  - The traceroute capture.
  - Wording that only resembled a private shape.
- **Deliberately not in scope, and why:**
  - **225 private-range addresses in runnable code.** Docker subnets, pod and service CIDRs, VPC declarations, NAT and
    firewall rules, SSRF block lists, and subnet calculators teach the range itself, so changing them is a per-lesson
    decision.
    Two of them are rows in the advanced-networking private-versus-public classifier example. They are unchanged, so
    their comment still points at the address Example 5 used before this PR.
  - **17 teaching hosts that sit inside those ranges.** They change only together with their ranges.
  - **28 scanner boundary artefacts.** Terraform `local_file`, a Rust local-address call, and an Erlang local-time call
    only match because the scanner's hostname pattern treats an underscore as a boundary; the fix belongs in the
    scanner, not in the examples.
  - **Seven home paths in `plans/done/2026-05-11__rename-governance-to-repo-governance/`.** Any commit touching those
    files fails pre-commit: the frontmatter validator's governance-area check matches the plan's own path as a
    substring. That needs a validator fix first.
  - **One systemd `ExecStart` path.** Its only clean runnable form is the `%h` specifier, which is a behaviour choice.
  - **Git history.** It still holds the original values.

## Summary

- `docs(ayokoding-www): mask private hops in the traceroute capture` (3 files): private hops are masked with a note;
  public hops are unchanged.
- `docs: normalise maintainer home paths in plans, tutorials, and rules` (199 files): the account-specific prefix
  becomes `~/`, `$HOME/`, `<repo-root>`, or `/Users/<name>/` by context. Generic names become `/home/<user>/`,
  `$GITHUB_WORKSPACE`, or `WORKDIR /app`. Where a normalised line also named a sibling checkout, that name reads
  `<sibling>`. Each touched completed plan gets a one-line normalisation note.
- `docs(ayokoding-www): drop the home prefix from profiler and debugger artefacts` (4 files): rewritten consistently,
  not regenerated.
- `docs(ayokoding-www): reword examples that only resembled private-network shapes` (22 files), including one example
  file renamed to an `sshd.conf` drop-in, with its references updated.
- `chore: take literal database credentials out of examples and dev defaults` (96 files): examples use placeholders,
  Spring and environment lookups, or `${VAR:-default}`. Local development defaults keep the same default value when the
  variable is unset.
- `docs: move teaching addresses and hostnames to reserved example shapes` (103 files):
  - Arbitrary hosts move to RFC 5737 blocks, one mapping per file, and internal-style hostnames to `.test`, `.invalid`,
    or `.example`.
  - Where the private range is the lesson, the leading octets stay and the rest is masked.
  - Values derived from an address, such as integer and packed forms, raw header hex, and the IPv4-mapped form, are
    recomputed.

## Reading Guide

- **Start here (runtime-reachable code):**
  - `infra/dev/ose-app/docker-compose.yml`, `apps/ose-be/scripts/run-e2e.sh`, and `apps/organiclever-be/scripts/run-e2e.sh`:
    the password is now `${POSTGRES_PASSWORD:-<same default>}`.
  - `apps/ose-be/tests/unit/Tests/*.fs` and `apps/organiclever-be/tests/unit/Tests/*.fs`: unreachable test connection
    strings drop the literal password, and a test base URL moves to an `.invalid` host.
  - `apps/ayokoding-www/src/features/cost-of-living-calculator/`: the `local` field of `toDisplayCurrencies` is renamed
    `localAmount`, with its two readers and its tests updated.
- **Skim:** tutorial, plan, and governance rewrites are mechanical per-file mappings.
- **Worth a look:** the raw IPv4 header hex in
  `apps/ayokoding-www/content/en/learn/legacy/software-engineering/networking/by-example/advanced.md` now carries a
  correct total-length field. The original never matched the example's own 28-byte length.

- **Natural seam:** one remediation class, making the tracked tree scan clean except for the out-of-scope rows above,
  split into one commit per category.
- **Production-deployable state:** no deployed configuration, environment file, or production code path changes. Local
  development and E2E scripts behave as before when `POSTGRES_PASSWORD` is unset.
- **Feature flag lifecycle:** not applicable.

## Verification

- `test:quick` for `ose-be` (rerun without cache after the test change), `organiclever-be`, and `ayokoding-www` (rerun
  after the address commit) all pass. The runs were made per project, under the local resource guard.
- `docker compose config` is valid for the edited development compose file.
- The rewritten Python networking examples were rerun, and their output matches the `# =>` comments. One rewritten
  pytest companion and one shell companion pass.
- Credential replacements were checked one per synthetic file against the scanner's credential detectors, with a
  flagged control.
- A whole-tree rescan of this branch reports 278 findings, all of them pre-existing and none new.
- Every commit's added lines and message, and the whole diff, pass the workspace outbound screen as well.
- **Not executed:**
  - the Clojure Migratus example restructure;
  - the Java MySQL demo URL form, checked against driver documentation only;
  - the anti-pattern Java examples, which are not meant to run.

## Risk and Rollback

- **Readers of examples** now substitute a value for `${VAR}` placeholders. A few examples introduce a variable name
  without defining it, and a GitHub Actions `env:` block and a quoted heredoc keep `${POSTGRES_PASSWORD}` literally.
- **Masking conventions** differ slightly between tutorial areas: the information-security pages mask the second octet,
  and the other address pages mask the third. Each file is internally consistent.
- **Rollback** is a revert of the commit for the affected category.


wahidyankf (commented):
## Focused PR Leak Review

Leak-only inspection of the whole diff at the pinned head: 399 files across six commits. The PR replaces values that
look private with placeholders and reserved example shapes, so the review reads the added side of every hunk.

- **Real secrets or private values:** none added. Removed lines carried documentation-example and local-default
  database credentials. Added lines carry placeholders, environment lookups, or `${VAR:-default}` whose default is the
  local development value the file already had. Added addresses are RFC 5737 documentation blocks, masked private-range
  octets, or `.test`, `.invalid`, and `.example` hostnames.
- **Protected production or staging properties:** none. No environment file, deployment input, or deployed
  configuration is touched; the only compose file is the local development stack.
- **Real machine-specific absolute paths:** none added. Home prefixes become `~/`, `$HOME/`, `<repo-root>`,
  `/Users/<name>/`, `/home/<user>/`, `$GITHUB_WORKSPACE`, or `WORKDIR /app`.

Result: `pass`.

PR #517, opened 2026-09-13
fix(ci): provision main formatting verify and backend image builds

wahidyankf:
## Outcome and Why

The push-triggered run for #516 failed four checks on `main`: `Auto-format affected (lint-staged)`, `Quality gate`,
`Publish ose-be`, and `Publish organiclever-be`. Both causes were already on `main`; #516 was the first push to reach
them.

- **Formatting verify on `main`:** the format job's push-only step runs the whole `formatting-verify` group, and
  `format-verify-ruff` found no `ruff` binary. Only the gate matrix job installs Ruff, and no earlier `main` push had
  changed a Python file.
- **Backend images:** both Dockerfiles never copy `libs/fsharp-env-loader`, which both projects have referenced since
  #176, so `dotnet publish` fails with FS0039. Every `main` publish since then skipped these images, because neither
  backend was affected.

## Scope

- **In scope:** the format job's Ruff install for the push-only verify, and the env loader copy in both backend
  Dockerfiles.
- **Deliberately not in scope, and why:**
  - **Making Ruff a doctor-managed tool.** That changes the tool registry for every surface. Here the install mirrors
    the matrix job's existing scoped step.
  - **A pull-request image build.** `publish-images` runs only on `main`, so a broken image build still surfaces only
    after merge. Adding a PR-side build is a workflow design decision.

## Summary

- `fix(ci): provision main formatting verify and backend image builds` (3 files):
  - `.github/workflows/pr-quality-gate.yml`: a `pipx install ruff==0.15.9` step, the matrix job's pin, runs before
    `Verify affected formatting on main`, scoped to push events.
  - `apps/ose-be/Dockerfile` and `apps/organiclever-be/Dockerfile`: the env loader's project file is copied before
    `dotnet restore`, and its `src/` before `dotnet publish`.

## Reading Guide

- **Start here:** the two Dockerfiles. Each gains two `COPY` lines that match the `ProjectReference` path in the app's
  project file.
- **Then:** the one workflow step, placed directly above the verify step that needs it.

- **Natural seam:** the provisioning gaps behind one failed `main` run, in one commit.
- **Production-deployable state:** merging runs `publish-images`, which builds and publishes both backend images for
  the first time since the env loader was introduced. Runtime images still contain only the published output.
- **Feature flag lifecycle:** not applicable.

## Verification

- Both images build locally from the repository root with `docker build -f apps/<app>/Dockerfile .`, under the local
  resource guard.
- `actionlint` passes on the workflow, and `hadolint` passes on both Dockerfiles.
- The failing `main` run's logs show both causes: `ruff: not found` in the formatting verify, and FS0039 on
  `FsharpEnvLoader` after the build skipped the missing project.
- **Not executed:** the push-only Ruff step itself, which runs only on `main` and is exercised by the next push that
  changes a Python file.

## Risk and Rollback

- **Image publication:** the first successful publish replaces whatever tags these images last carried, as intended by
  the existing workflow.
- **Rollback:** revert this commit. The formatting verify on `main` and both image builds would fail again as before.


wahidyankf (commented):
## Focused PR Leak Review

Leak-only inspection of the whole diff at the pinned head: three files in one commit. The diff adds one workflow step
and four Dockerfile lines, so the review reads every added line.

- **Real secrets or private values:** none added. The workflow step installs a pinned public package; the Dockerfile
  lines copy repository paths into the build stage.
- **Protected production or staging properties:** none. No environment file, secret reference, deployment input, or
  runtime environment variable changes. The image runtime stage is unchanged.
- **Real machine-specific absolute paths:** none added. Every path is repository-relative.

Result: `pass`.

PR #518, opened 2026-09-14
fix: clear the remaining public-safety baseline findings

wahidyankf:
## Summary

Clears the findings that remained in the repository's public-safety baseline after the first remediation cycle. Four
commits, no product behaviour change outside the frontmatter validator.

- **Hermes gateway unit.** `ExecStart` runs `/bin/sh -c 'exec "$$HOME/.local/bin/hermes" gateway'` instead of naming a
  home directory. systemd sets `HOME` from `User=hermes`; `$$` passes a literal `$` past systemd's own expansion; `exec`
  keeps the gateway as the main process, so `Type=simple` and `Restart=on-failure` still track it. The `%h` specifier
  does not fit: in a system unit it resolves to the service manager's home.
- **Frontmatter classifier.** `classifyDocArea` counted a docs-area prefix anywhere in a path, so a folder whose name
  merely ends in `repo-governance` was held to the governance schema. A prefix now counts only where a path segment
  starts. A new scenario in `docs-validate-frontmatter.feature` is bound at the unit, integration, and e2e layers, and
  the parity manifest is regenerated for `Md.fs` and the feature file; the sibling repository receives the identical
  change. The completed rename plan's records now name the parent workspace's `CLAUDE.md` by a home-relative path.
- **Lesson host addresses.** Lessons that named individual hosts inside private ranges use RFC 5737 documentation
  addresses, derive each host from its network with Python's `ipaddress`, or mask the octet the lesson does not depend
  on, with a note above any masked output block. Network prefixes stay as written, and every configuration example
  keeps its gateway, static address, and cluster DNS address inside its own subnet.

## Verification

- Public-safety shape scan (ose-rules v0.2.3 `check.sh`, CI surface) over the committed tree: 0 findings, down
  from 278.
- rhino-cli: the new scenario fails before the fix (unit and integration) and passes after it at all three layers; full
  suites pass (unit 758, integration 1482, e2e 368).
- Edited lesson scripts compile and run.
- Outbound screens of the diff, commit messages, branch, title, body, and leak review: clean.
- Pre-commit and pre-push hooks ran without bypass.


wahidyankf (commented):
## Focused PR Leak Review

Leak-only inspection of the whole diff at the pinned head: 32 paths in four commits. Every added line was read.

- **Real secrets or private values:** none added. The systemd line names a shell, an environment variable, and the
  lesson's own binary. The classifier change adds a path-segment rule, a Gherkin scenario, and fixtures that use
  repository-relative temporary paths. The lesson changes add RFC 5737 documentation addresses, addresses derived at run
  time from networks the lessons already publish, and masked octets; no address names a real machine.
- **Protected production or staging properties:** none. No environment file, secret reference, deployment input, or app
  runtime configuration changes; the changed files are tutorial content, CLI source and tests, a parity manifest, and a
  completed plan's records.
- **Real machine-specific absolute paths:** none added. The change removes a hardcoded home directory from the Hermes
  unit and replaces four plan records' machine-specific home paths with a home-relative form.

Result: `pass`.

PR #519, opened 2026-09-14
feat(repo): pin RHINO and adopt the public-safety gate

wahidyankf:
## Outcome

OSE Public carries the pinned RHINO release and the adopted public-safety gate. Neither is wired into a hook, a gate, or
the pull-request quality gate yet; the next foundation unit does that through the pinned binary.

## Why

The governance foundation lands before any corpus migration, in two parts so each stays reviewable. This first part adds
only files that change no existing behaviour: the executable pin with its bootstrap proof, and the outbound gate copied
from its catalog release.

## Change

- `rhino` and `rhino.lock` pin RHINO `v0.3.0` by version, commit, and archive digest. The bootstrap installs into a
  machine-wide cache outside the repository and verifies the binary and its reported identity on every use.
- `.github/scripts/test-rhino-bootstrap.sh` is BeaverNest's RHINO bootstrap adapter. As with the HIPPO adapter, no copy
  of the feature file lives here: the exact scenario titles are bound in the adapter, and a caller may set
  `RHINO_CANONICAL_FEATURE` to BeaverNest's feature to compare them before any fixture runs.
- `.github/workflows/rhino-consumer-smoke.yml` runs that adapter and a real install of the pin on Linux and macOS for
  every change to the bootstrap, and twice a day, in the same shape as the HIPPO consumer smoke.
- `scripts/public-safety/` is the outbound public-safety gate from ose-rules `v0.2.4`, byte for byte, with its
  fifteen-case suite.
- The workflow map and the scripts README name both.

## Verification

- The bootstrap adapter passes all 21 scenarios with `RHINO_CANONICAL_FEATURE` set to BeaverNest's feature and without
  it; a renamed scenario title fails before any fixture runs.
- `bash scripts/public-safety/tests/run.sh` passes all 15 cases, and every file under `scripts/public-safety/` is
  byte-identical to the ose-rules `v0.2.4` release.
- `shfmt -d` and `shellcheck --severity=warning` report nothing for the new shell files; `actionlint` accepts the new
  workflow.
- The pre-commit and pre-push gates ran without bypass, and the diff, commit message, title, and body were screened
  clean.

## Risk

- No product surface, hook, or quality gate changes. The new workflow runs only for changes to the bootstrap paths, on
  its schedule, or on dispatch.
- A future catalog correction arrives as a new release copy; the gate files here are not edited in place.


wahidyankf (commented):
Leak review of the pinned head: whole diff, title, and body read. The change adds the RHINO pin and bootstrap, its hermetic test adapter and smoke workflow, the public-safety gate copied byte for byte from a public catalog release, and two README edits. The adapter builds its fake releases under temporary directories at run time, the catalog's probe values are assembled from fragments, and the only URLs are the public release and catalog locations. No real secret or credential, no real address or hostname, no property that belongs in environment storage, and no machine-specific absolute path; every path is repository-relative or built at run time.

PR #520, opened 2026-09-14
feat(repo): route hooks and CI through RHINO with repo-config v2

wahidyankf:
## Outcome

The pinned RHINO binary now owns every gate surface in OSE Public. The three Git hooks and the pull-request quality gate
call `./rhino gate run`, the outbound public-safety gate runs first on each surface, and rhino-cli keeps running every
existing gate beneath it.

## Why

The governance foundation lands before any corpus migration. Unit 1a added the pin and the safety gate without wiring
them in; this unit wires them, so later units inherit one entry point per surface.

## Change

- `repo-config.yml` is `ose/repo-config/v2`. Its core declares public visibility and five gates: `public-safety` on all
  four surfaces, then one registry gate per surface that hands the surface to rhino-cli. Every v1 section moved
  unchanged under `extensions.rhino-cli`. The two YAML anchors were expanded, because RHINO does not resolve aliases.
  Two comments in the moved env-injection section no longer name a private repository.
- `governance-word-budget` now excludes `scripts/public-safety/`. The gate is path-gated on `repo-config.yml`, so this
  is the first change to run it since unit 1a adopted that directory. Its README is byte-identical with the catalog
  release and is over the 1000-word limit, and editing it here would break that identity.
- rhino-cli reads v1 and v2 alike: `RepoConfig.normalize` gives every reader the extension of a v2 document, so each v1
  fixture parses as before. Every child a gate run launches receives `OSE_GATE_SURFACE` for the surface it runs,
  overriding an inherited value.
- `.husky/commit-msg`, `pre-commit` and `pre-push` call `./rhino gate run --surface <s>`, and `gate validate` now
  requires that form. The CI matrix and the push-only formatting verify dispatch through
  `./rhino gate run --surface ci -- --group=<id>`.
- A new extension gate, `public-safety-tests`, runs the public-safety suite in the `shell-docker-actions` CI group.
- The `package.json` lint-staged block is regenerated from the registry, the parity manifest is regenerated for the
  changed sources and scenarios, and the documents that name what the hooks invoke use the new form. Two Mermaid labels
  in `git-hook-lifecycle.md` that already exceeded the 20-character limit are shortened, because the label check runs
  on every staged Markdown file.

## Verification

- `gate list --format=json` on all four surfaces, and `--by-group` on ci, is byte-identical between the v1 file and the
  v2 extension, apart from the one added CI gate and the word-budget exclusion.
- New scenarios cover the v2 reader, the surface export, a hook that bypasses the pinned binary, and a matrix group
  dispatched straight to the repository CLI; each failed before the change. The full unit, integration and e2e suites
  pass.
- `./rhino repo-config validate`, rhino-cli `repo-config validate`, `gate validate` and `parity manifest validate`
  pass; `actionlint`, `shellcheck` and `shfmt` accept the workflow and hooks.
- The commit-msg surface passes a conventional message and refuses a malformed one. The pre-commit and pre-push hooks
  ran without bypass, and the diff, commit message, title and body were screened clean.

## Risk

- Every `./rhino gate run --surface ci` call scans the tracked tree first, as RHINO requires for a public repository.
  Measured locally, that is about three minutes per matrix group job.
- RHINO reports only `[gate] <id> failed` and drops the gate's own output. A failure-only step after each dispatch
  reruns the same group through the repository CLI, so the job log still shows the cause.
- The lint-staged block had already drifted from the registry: the `*.go` entry lacked `scripts/lint-golangci.sh`.
  Regenerating it with the v1 file produces the same correction.
- The changed rhino-cli sources and scenarios sit inside the byte-identical parity boundary, so a repository that shares
  that boundary needs the same change before its own manifest validates.


wahidyankf (commented):
Leak review of the pinned head: whole diff, title, and body read. The change moves repo-config.yml to schema v2 with the existing rhino-cli sections nested unchanged under an extension, adds a v2 reader and a surface environment export to rhino-cli, points the three Git hooks and the CI matrix at the pinned RHINO binary, updates the tests and documents that name the old hook form, regenerates the lint-staged block and the parity manifest, and adds the public-safety suite as a CI gate. Test fixtures build their repositories under temporary directories at run time; the only commands and paths are repository-relative. No real secret or credential, no real address or hostname, no property that belongs in environment storage, and no machine-specific absolute path.

PR #521, opened 2026-09-14
feat(rhino-cli): add plan validate and the plan-structure gate

wahidyankf:
## Outcome

OSE Public checks its own plans against the shared plan structure. rhino-cli gains `plan validate`, held to the pinned
RHINO binary's output byte for byte, and a `plan-structure` gate runs it in pre-commit and CI.

## Why

Unit 1b put every gate surface behind the pinned RHINO binary. This unit adds the structural plan check that the shared
plan-validator contract requires, so plan decay fails a gate instead of waiting for review.

## Change

- `specs/fixtures/plan-structure/` is the shared fixture corpus, copied from the published catalog and verified by its
  `SHA256SUMS` and `CORPUS-DIGEST`. Prettier and markdownlint skip it, its two deliberately misnamed companions are
  exempt from Markdown naming, and the link check excludes it. The README index now walks `specs/apps/` and
  `specs/libs/` rather than all of `specs/`, so the new `specs/fixtures/` gets its own README.
- `RhinoCli.Application/src/Plan.fs` ports RHINO v0.3.0's plan validator: lifecycle roots, the six documents, technical
  companions, acceptance identifiers and the delivery checklist. Rule identifiers, messages, diagnostic order and exit
  classes match RHINO (`0` clean, `1` findings, `2` a plan document it cannot read), in text and JSON.
- `plan validate` walks `plans/` and prints that report, and the help text lists it.
- `specs/apps/rhino/cli/behaviours/plan/plan-structure.feature` has six scenarios over the corpus, bound in the unit,
  integration and e2e suites.
- The `plan-structure` gate runs in pre-commit when a `plans/` file is staged, and in CI with the `markdown` group. It
  always walks the whole tree, because a plan's checks span its documents. The lint-staged block is regenerated from
  the registry.
- The parity manifest is regenerated for the new sources and scenarios, and the rhino-cli README documents the command.

## Verification

- The F# command and the pinned RHINO binary produce identical exit codes, stdout and stderr, in text and JSON, over
  every corpus case, a set of edge trees, and this repository's own `plans/` tree.
- Unit tests over edge trees the corpus never reaches (CRLF, unterminated text, fences, link suffixes, an extensionless
  companion, JSON escapes, an unreadable document) assert RHINO's own output for the same trees.
- The new scenarios failed before the implementation. The full unit, integration and e2e suites pass, and unit line
  coverage stays above the 99% floor.
- In a throwaway copy of the tree the gate passes, then fails with `PLAN-LIFECYCLE-003` once one archived plan loses its
  completion-date prefix.
- `./rhino repo-config validate`, rhino-cli `repo-config validate`, `gate validate`, `parity manifest validate` and
  `plan validate` pass. The diff, commit message, title and body were screened clean.

## Risk

- The shared plan-validator contract classes exit `2` as invalid configuration or usage and exit `3` as a dependency or
  execution failure. RHINO v0.3.0 exits `2` when a plan document cannot be read, which reads more like the latter. This
  command follows RHINO so both implementations stay identical, and the mismatch is left to the contract's owner.
- Markdown naming exemptions match by base name, so a file elsewhere with the same name as one of the two exempt
  companions would also be exempt.
- The new sources and scenarios sit inside the byte-identical parity boundary, so a repository that shares that
  boundary needs the same change before its own manifest validates.


wahidyankf (commented):
Leak review of the pinned head: whole diff, title, and body read. The change adds the shared plan-structure fixture corpus by digest, a plan validator and its `plan validate` command to rhino-cli, one feature with its unit, integration and e2e bindings, a plan-structure gate on pre-commit and CI, the ignore and exemption entries the corpus needs, two README files, and the regenerated lint-staged block and parity manifest. The corpus holds synthetic plan documents only. Tests build their plan trees in memory or under temporary directories at run time; the only commands and paths are repository-relative. No real secret or credential, no real address or hostname, no property that belongs in environment storage, and no machine-specific absolute path.

PR #522, opened 2026-09-14
feat(rhino-cli): hand RHINO-owned validators to the pinned binary

wahidyankf:
## Outcome

rhino-cli keeps only the governance rules the pinned RHINO v0.3.0 does not provide. Validators RHINO owns now run the
pinned `./rhino`, and `repo-config.yml` records which commands delegate, which split, and which stay in F#.

## Why

Unit 1b put every gate surface behind the pinned RHINO binary, and unit 2 added the shared plan check. Two
implementations of the same Markdown and governance rules would drift, so the rules RHINO provides move to it, and a
source scan keeps them from coming back.

## Change

- `RhinoCli.Infrastructure/src/RustRhino.fs` starts `./rhino` with no shell and inherited stdio, and returns its exit
  code unchanged. A missing binary exits `3`. It does not recheck the digest, because `./rhino` already does.
- `repo-config.yml` gains RHINO's validator sections (`scan`, `governance-word-budget`, `md-frontmatter`,
  `md-heading-hierarchy`, `md-internal-link`, `md-naming`, `md-readme-index`, `convention-emoji`) and an
  `extensions.rhino-cli.delegation` block. The block has one row per F# validator command, classed `delegate`, `split` or
  `stay`, and `gate validate` refuses a missing, duplicate or extra row. Gate `args` that now live in the sections are
  removed, and the lint-staged block is regenerated.
- `md naming`, `md heading-hierarchy` and `convention emoji` delegate whole. `repo-config`, `md frontmatter`,
  `governance word-budget`, `md links` and both README gates run RHINO first, then their F# remainder. A retired policy
  flag exits `2` and names the section that now holds the policy.
- The F# per-file word budget, emoji, naming and heading rules, and the missing-README check, are deleted with their
  renderers, tests and scenarios. `KeepListSourceScanTests.fs` fails when a rule site is on no `keeps` list or is a rule
  RHINO provides; `plan-structure` is the named exception.
- `gate/rust-delegation.feature` holds the contract, bound in the unit, integration (stub `./rhino`) and e2e suites. The
  e2e suites give their temporary repositories a stub `./rhino`.
- RHINO's stricter checks needed repairs: 40 fragment-named Markdown files are renamed along with their references, 16
  broken internal links are fixed, one forbidden `updated` front-matter key is removed, and `session-card.tsx` writes its
  star glyph as an escape that renders the same.
- The rhino-cli README and the SDLC gate standard describe the boundary, and the parity manifest is regenerated.

## Verification

- Before the repairs, the pinned RHINO found 40 naming and 16 internal-link findings over the tracked tree. After them,
  all eight RHINO validators this unit relies on exit `0` with no findings.
- The delegation scenarios and the keep-list scan failed before the implementation. The full unit (741), integration
  (1409) and e2e (351) suites pass, and unit line coverage is 99.15% against the 99% floor.
- `./rhino repo-config validate` and rhino-cli `repo-config validate`, `gate validate`, `parity manifest validate`,
  `plan validate` and `md links validate` pass. Regenerating the harness bindings changes nothing.
- The pre-commit and pre-push surfaces and all six CI groups pass through `./rhino gate run`. The diff, commit message,
  title and body were screened clean.

## Risk

- Delegated validators walk their declared surface instead of the staged files, because RHINO validators take no file
  list. A matching staged file still decides whether the gate runs; the delegated pre-commit set adds about 3 s.
- Coverage follows the declared sections. The README index covers 18 named trees, and the heading check lists app and
  library READMEs by name, because a RHINO `*` crosses `/`. A new tree or README outside them is not checked until it is
  listed.
- RHINO's emoji set differs from the old F# set: it adds three ranges and drops two that have no tracked occurrences.
- Split commands refuse `--output json`, and delegated commands print RHINO's report format. Nothing in CI or the
  scripts reads the old JSON.
- Long lists in `repo-config.yml` must stay block sequences. RHINO's reader cannot parse a flow sequence that Prettier
  splits across lines, and every validator then exits `2`.
- `repo-config-schema`, `md-frontmatter`, `governance-word-budget` and `governance-readme-completeness` keep an F#
  remainder, because RHINO lacks the rule. Mermaid and harness validators stay in F# until later units.


wahidyankf (commented):
Leak review of the pinned head: whole diff, title, and body read. The change adds a launcher for the pinned `./rhino` binary to rhino-cli, RHINO validator sections and a delegation block in `repo-config.yml`, one feature with its unit, integration and e2e bindings, removes the F# rules RHINO now provides with their tests and scenarios, renames 40 governance Markdown files and updates their references, repairs 16 internal links, and regenerates the lint-staged block and parity manifest. Test fixtures build temporary repositories with a stub `rhino` script at run time; every command and path is repository-relative. No real secret or credential, no real address or hostname, no property that belongs in environment storage, and no machine-specific absolute path.

PR #524, opened 2026-09-14
docs(governance): make repo-governance diagrams accessible

wahidyankf:
## Outcome

Every diagram in the frozen `repo-governance/` diagram paths is accessible. Mermaid diagrams carry `accTitle` and
`accDescr`, conceptual text diagrams are Mermaid, and diagram colours meet the palette and contrast rules.

## Why

Units 1a to 4 moved governance validation onto RHINO and adopted the ose-rules plan system. The diagrams were left for
four review-sized units. In these paths, 126 Mermaid diagrams had no accessibility text, many used inline colours or
white text on light fills, and 35 conceptual diagrams were drawn with arrows in `text` fences.

## Change

- Every Mermaid diagram in the 162 frozen paths gains `accTitle` (its section heading) and `accDescr` (the
  relationships it shows).
- 35 conceptual text diagrams (flows, decision trees and layer chains) become Mermaid. Detail that does not fit a short
  label moves to prose after the diagram. The other 98 text blocks are literal trees, traces or templates and stay
  `text`.
- 25 fences that exist to show a mistake are re-fenced as literal `text`: FAIL examples, a bare `classDef` fragment and
  entity-code syntax demonstrations. The reader still sees the wrong source, and it no longer renders.
- Inline `style` lines become `classDef`s. Every class uses a palette fill, a different palette stroke and the text
  colour with the higher contrast: white on blue and black, black on every other fill. Off-palette spellings map to
  the matching palette hue.
- Long node and edge labels are split with `<br/>` or shortened to the 20-character strict limit.

## Verification

- The pinned `./rhino md mermaid validate`, run over a copy of the 162 paths with an `md-mermaid` section, checks 128
  diagrams with no findings. Before the change it reported 450 findings.
- rhino-cli `md mermaid validate` at the strict 20-character limit reports no violation in these paths.
- Prettier and markdownlint pass on all 91 changed files, and no file exceeds its word budget.
- The pre-commit and pre-push surfaces and all six CI groups pass through `./rhino gate run`. The diff, commit message,
  title and body were screened clean.

## Risk

- RHINO `md-mermaid` is not yet a declared gate here; a later diagram unit declares it. Until then the F# Mermaid gates
  stay authoritative.
- Some labels change from white to black text where white failed contrast, so those diagrams look different.
- Re-fenced wrong examples no longer render. They exist to show broken source, so the source is what readers need.


wahidyankf (commented):
Leak review of the pinned head: whole diff, title, and body read. The change edits Markdown governance documents only: it adds accessibility titles and descriptions to Mermaid diagrams, converts conceptual text diagrams to Mermaid with explanatory prose, re-fences deliberate wrong examples as literal text, and adjusts diagram colour classes and label line breaks. Every path is repository-relative. No real secret or credential, no real address or hostname, no property that belongs in environment storage, and no machine-specific absolute path.

PR #529, opened 2026-09-15
docs(repo): name the private sibling by role, not by repository

wahidyankf:
## Outcome

The repository names its private sibling by role instead of by repository, and no tracked file carries an internal
host.

## Why

The public-safety term set blocks the private repository's name and internal host names on every outbound surface.
Units 1a to 8 kept new text clean, but 312 tracked paths still named the sibling and 13 held internal hosts. This unit
removes the remainder so the whole tree scans clean.

## Change

- Prose says "the private sibling". Identifiers, file names and anchors use `private-sibling`, and shell and data
  snippets use `<private-sibling>`. Links into the private repository keep their label, lose their URL and say that
  they are not linked.
- 10 paths whose names carried the repository name are renamed with the slug, and every referrer follows.
- rhino-cli: the `Parity.fs` message, an `Env.fs` comment, the `iac-env-validation` step text and its bindings, the
  parity-manifest assertions and two test comments use the new wording. The parity manifest is regenerated.
- The agent source is edited, and the Codex and OpenCode adapters and skill mirrors are regenerated from it.
- Archived plans take the same mechanical substitution; their internal host names become `internal-host`.
- The ayokoding advanced-networking captures use RFC 5737 documentation addresses.

## Verification

- rhino-cli unit (746), integration (1414) and E2E (352) suites pass. Unit line coverage is 99.15%.
- Fantomas, FSharpLint, the F# analyzers and behaviour coverage for all four adapters pass.
- A rescan of the tracked tree with the workspace term set finds nothing; its negative and positive controls behave.
- Prettier and markdownlint pass on every changed Markdown file. The pinned RHINO validators pass, including word
  budgets, internal links and README indexes.
- The pre-commit and pre-push surfaces and all six CI groups pass through `./rhino gate run`. The added lines, new path
  names, commit message, title and body were screened clean.

## Risk

- Removed lines and old path names in this diff, and earlier history, still show the old wording; history is not
  rewritten.
- Archived plans change by substitution only. Links into renamed archived files are repointed in the same commit.
- The `iac-env-validation` step text and parity message change here first; the private sibling's copy follows in its own
  phase.


wahidyankf (commented):
Leak review of the pinned head: whole diff, title, and body read. The change replaces the private sibling's repository name with role wording, a slug or a placeholder across documentation, governance, agent and skill sources, generated adapters, archived plans, and rhino-cli F# sources, tests and Gherkin specifications; renames the paths whose names carried it; removes links into the private repository; replaces internal host names in archived plans with a neutral token; and rewrites lesson packet captures with RFC 5737 documentation addresses. Removed lines and old path names show text already in the public history; the added side introduces none. Every path is repository-relative. No real secret or credential, no real address or hostname, no property that belongs in environment storage, and no machine-specific absolute path.

PR #531, opened 2026-09-15
test(rhino-cli): tighten bindings and align Fantomas 8

wahidyankf:
## Outcome

Every rhino-cli scenario that Phase 13 added or modified now has, at each layer, a binding that fails when the behaviour
it names breaks. Fantomas is also aligned on the latest compatible stable release across local and CI execution.

## Why

A Gherkin implementation review of the Phase 13 scenarios found 11 bindings, in five findings, that could not fail on
their scenario's behaviour:

- The image-link scenario had no runner at any layer.
- The `--fail-kinds` integration binding audited synthetic findings, and its E2E Then could not tell a restricting flag
  from an ignored one.
- The integration split-remainder Then accepted any non-empty output.
- The E2E derived-files Then checked only that the gate succeeded.

## Change

- `MdSteps.fs`, `MdResourceSteps.fs` and `MdProcessSteps.fs` run `An image link to a missing file is reported`.
- `--fail-kinds`: the integration Given writes a real tree with one orphan and one ghost, and the When runs the audit.
  The unit and integration Thens check the parsed kinds and prove the other findings fail an unflagged run but not the
  flagged one. The E2E Then reruns the tree without its orphan, with and without the flag.
- `RustDelegationResourceSteps.fs` asserts the exact links report per documentation tree.
- `GateExecutionProcessSteps.fs` stubs RHINO with an argv recorder and requires `md mermaid validate --file a.md --file
  b.md`; an unstaged `c.md` with a diagram must not reach it.
- The local tool manifest pins Fantomas 8.0.0, and the setup action installs that exact manifest version instead of
  floating independently.
- A root F# editor config retains the established `cramped` bracket layout. The two nested backend configs use the
  supported auxiliary setting names, and the 19 remaining formatter changes complete the 336-file v8 migration.

The behavioural change remains confined to the eight test bindings. The other changes are formatter configuration or
Fantomas-only rewrites; no feature text or parity-manifest row changes.

## Verification

- Each tightened binding was run against a scratch-only production mutation that its old binding survives. Every
  mutated run went RED and every restored run went GREEN: M1, M2b, M3, M4 and M5 each record
  `old=GREEN new=RED restored=GREEN`; the broader M2 probe is retained as a rejected non-witness.
- rhino-cli unit (747), integration (1415) and E2E (353) suites pass. Unit line coverage is 7852/7919 (99.15%) against
  the 99% threshold.
- Fantomas, FSharpLint, the F# analyzers and behaviour coverage for all four adapters pass.
- All 336 tracked F# files pass Fantomas 8.0.0. Affected build, lint, and test pass across Crane, Organiclever, OSE, and
  RHINO; the final lint and test runs bypassed Nx cache.
- A re-review of the same scope reports 0 failing rows (previously 11).
- The pre-commit and pre-push surfaces and all six CI groups pass through `./rhino gate run`. The added lines, commit
  message, title and body were screened clean.

## Risk

- Fantomas 8 is a major formatter upgrade. The repository explicitly retains the prior bracket style, every tracked F#
  file passes the new formatter, and affected builds/tests/lints pass without cache. The E2E fail-kinds Then runs the
  CLI twice more per run, adding a few seconds to that suite.


wahidyankf (commented):
Leak review of the pinned head: whole diff, title, and body read. Eight rhino-cli F# step files make existing Gherkin scenarios fail when their behaviour breaks. The repair also pins Fantomas 8.0.0 consistently, adds F# formatting compatibility settings, and applies formatter-only rewrites to the remaining 19 files in the 336-file corpus. Every test fixture is synthetic and temporary; all configuration and source paths are repository-relative. No real secret or credential, no real address or hostname, no property that belongs in environment storage, and no machine-specific absolute path.

PR #537, opened 2026-09-16
fix(security): clean interrupted safety canaries

wahidyankf:
## Summary

- Keep the generated credential canary inside the preflight workspace whose exit trap removes it.
- Give the non-disclosure regression case a private temporary root so concurrent public-safety runs cannot observe or
  remove each other's canary workspace.
- Preserve the scanner's fail-closed behavior and non-disclosing diagnostics.

## New-code cost and benefit

The change adds no production dependency or long-lived process. It adds one cleanup assignment and isolates one test's
temporary state, preventing interrupted scans and concurrent test runs from leaving or racing over credential-shaped
artifacts.

## Test plan

- [x] Public-safety regression case `060-non-disclosing-output`
- [x] Complete repository gate
- [x] Exact inner `pre-push` registry
- [x] Real push hook: outer `public-safety` and `registry-pre-push`


wahidyankf (commented):
Focused leak-only review: **PASS**.

Reviewed the complete aggregate diff and delivery-controlled title/body from base
`926994d13becd0857c4dd93f6a355550aa7eb565` to exact head
`e88adc86edb090378e4ff9e1f785517a85f87bb2`. Found zero real secret or private values, zero protected environment
properties, and zero machine-specific absolute paths. Every other security and semantic concern was out of scope.

PR #539, opened 2026-09-17
feat(ose-id): local foundation for OSE ID (Plan 01)

wahidyankf:
## Outcome and Why

This PR delivers Plan 01 of the OSE ID initialization series: the local-only foundation for OSE
Identity/CIAM — four Nx projects (`ose-id-be`, `ose-id-be-e2e`, `ose-id-web`, `ose-id-web-e2e`), an
ASP.NET Core 10 / C# 14 backend behind a hexagonal/DDD boundary, PostgreSQL persistence via
SqlKata+Npgsql at runtime (EF Core confined to migration time only), independently observable
liveness/readiness/startup diagnostics, and an owned local-stack runner that starts and stops the
whole stack deterministically, including two-backend-instance and two-concurrent-run robustness
proofs. It answers the plan's stated problem: give every later OSE ID plan (local email accounts,
tenancy, OIDC, and so on) a production-disabled, fully local, testable foundation to build on, with
zero real user-facing identity surface shipped yet.

## Scope

- **In scope:** the four `ose-id-*` Nx projects; PostgreSQL migrations/schema/privileges;
  health/readiness/startup diagnostics; the local-stack runner (owned Postgres container, 1–2
  backend instances, web shell, deterministic startup and reverse-order cleanup, run-scoped
  isolation for concurrent runs); reserved local ports (`OSE_ID_WEB_PORT=3500`,
  `OSE_ID_BE_PORT=8501`, `OSE_ID_POSTGRES_PORT=5438`); archiving the plan folder to `plans/done/`;
  a handful of pre-existing in-branch-scope governance-gate defects fixed at root cause (a stale
  Rhino CLI parity manifest, the TestHost `integration-loopback` allowlist, missing per-app
  `LICENSE` files, and a `tsconfig.json` pollution bug in the local-stack runner).
- **Deliberately not in scope, and why:** baobab (explicitly excluded by the requester);
  account/sign-in/OIDC/company/tokens/passkeys/admin UI (later plans 02–09 in the same series);
  production deployment/Kubernetes (depends on the private sibling repo's infrastructure
  foundation); GraphQL/MCP delivery (architectural seam only, no implementation yet); two
  pre-existing, unrelated Nx-affected failures (`ayokoding-www-fe-e2e:test:e2e` content/DOM drift,
  `ose-www-be-e2e:test:e2e` build-parallelism contention) — confirmed via
  `git log <merge-base>..HEAD -- <path>` to be untouched by any commit on this branch.

## Summary

- Adds `apps/ose-id-be` (ASP.NET Core 10 / C# 14, hexagonal/DDD), `apps/ose-id-be-e2e`,
  `apps/ose-id-web` (Next.js), `apps/ose-id-web-e2e`.
- PostgreSQL migrations, schema, and privilege boundary; SqlKata+Npgsql at runtime, EF Core
  confined to migration time only (build-enforced absence of any EF runtime query path).
- Liveness/readiness/startup diagnostics, each independently observable.
- `apps/ose-id-be-e2e/scripts/local-stack.mjs`: one owned local-stack runner — starts Postgres,
  runs migrations, starts 1–2 backend instances, builds/starts the web shell, blocks until
  signaled, and stops everything in exact reverse order; two concurrent invocations never collide.
- Fixes a defect discovered mid-delivery: the runner's custom Next.js `distDir` caused Next's own
  TypeScript-setup verification to permanently leak two `tsconfig.json` `include` entries per run,
  which had already accumulated to 116 stale lines in committed history. Closed with cleanup-time
  pruning of only the current run's own entries under a lock file, a Prettier reformat pass to
  undo Next's own drive-by reformatting of the rest of the file, and a new regression test.
- Archives this plan from `plans/in-progress/` to `plans/done/2026-09-17__ose-id-init-01-foundation/`.

## Reading Guide

- **Start here:** `plans/done/2026-09-17__ose-id-init-01-foundation/README.md`, then
  `apps/ose-id-be-e2e/scripts/local-stack.mjs` and
  `apps/ose-id-be-e2e/steps/LocalStackRunnerProcess.cs` / `LocalStackRunnerTests.cs` for the
  runner and its robustness proofs.
- **Skip these paths:** `plans/done/2026-09-17__ose-id-init-01-foundation/evidence/**` (raw
  command-output logs, not narrative); generated mirrors (`.agents/`, `.opencode/`, `.codex/`)
  untouched by this PR.

- **Natural seam:** this is exactly Plan 01 of a nine-plan series — the smallest independently
  reviewable, locally verifiable slice (four projects, persistence, health, and a runner) with no
  dependency on any later plan in the series.
- **Production-deployable state:** safe to deploy immediately as-is; there is no production
  deployment target yet for this series (explicitly out of scope — see Non-Goals), and no
  user-reachable identity surface ships in this plan.
- **Feature flag lifecycle:** not applicable — nothing here is gated behind a flag; the entire
  delivered surface is local-only tooling, tests, and a backend/web pair with no external routing
  wired up yet.

## Cost/Benefit of Added Code

- **What it buys:** a real, testable, DDD-shaped backend + web pair with an owned, deterministic
  local dev/test stack — every later OSE ID plan can build directly on this instead of re-deriving
  persistence, health, and local orchestration from scratch.
- **What it costs to maintain:** four more Nx projects; one more local-only runner script and its
  C# robustness suite to keep passing under future Next.js/.NET upgrades.
- **Simpler alternative rejected, and why:** letting each test start its own ad hoc
  Postgres/backend/web was rejected — it would silently duplicate concurrency-safety logic across
  every test file, and it was exactly this shared-runner exercise, run repeatedly across a
  multi-day session, that surfaced the `tsconfig.json`-pollution bug fixed here — a bug no
  single test-file-scoped setup would have caught in one pass.

## Verification

- `rtk ./hippo run --class transactional --disk-path . -- npm exec nx -- run-many -t typecheck,lint,test:quick --projects=ose-id-be,ose-id-be-e2e,ose-id-web,ose-id-web-e2e` — green.
- `rtk ./hippo run --class transactional --disk-path . -- npm exec nx -- run ose-id-be-e2e:test:e2e` — 27/27 passed.
- `ose-id-be:test:unit` 96/96 passed, 99.58% Unit line coverage (≥99% required); `ose-id-web:test:unit` passed its enforced 99% Vitest line-coverage threshold.
- `repo-governance/workflows/gherkin-implementation-review.md` (`scope=changed`) — PASS: 86/87 rows PASS, 1 valid `@e2e-exempt` EXEMPT, 0 FAIL, backed by a 247-test runtime proof.
- `rtk ./hippo run --class transactional --disk-path . -- apps/rhino-cli/scripts/rhino-bin.sh gate run --surface=pre-push` — clean.
- Full detail and raw logs: `plans/done/2026-09-17__ose-id-init-01-foundation/delivery.md` Phase 7, `evidence/phase-7/`.

## Risk and Rollback

- **Risk:** low — everything shipped is local-only (no production route reaches this code), and
  the persistence/health/runner surfaces are each covered by Unit/Integration/E2E proofs plus the
  Gherkin implementation review.
- **Containment:** the two known-unrelated Nx-matrix failures (ayokoding-www, ose-www) are
  pre-existing and untouched by this branch, confirmed via git history.
- **Rollback:** revert the merge commit; nothing here is referenced by any other in-flight branch,
  and no production configuration or deployed service depends on it.

---

## Semantic Review Route Record (scout, cycle 1)

- **Pinned base / head:** base `main` @ `284504b689caf7e77dedd75c529f63a2069665df`; head
  `d89ec62ef144d74f7d7424ce7047cc7de499a9b4`. Merge-base `d9a832b6a49d7f5587dd63061bb45827a6a0b75e`
  (branch is 4 commits behind `main`; the reviewed diff is three-dot against the merge-base).
- **Frozen outcome / scope:** as stated in "Outcome and Why" and "Scope" above, at this head.
- **Classification evidence:** 383 files changed, +39,892 / -778 (three-dot). Extensions include 95
  `.cs`, 22 `.ts`, 5 `.tsx`, 1 `.fs`, 10 `.feature`, 65 `.md`, 4 `.yml`/1 `.yaml`. Not plans-only
  (executable source present). Security-sensitive paths touched: `.github/actions/setup-dotnet/action.yml`
  (CI), `repo-config.yml` (gate/allowlist registry), `apps/ose-id-be/.env.example`,
  `apps/ose-id-web/.env.example`, `.config/dotnet-tools.json`.
- **Risk tier:** `full` — over both thresholds (>50 lines, >20 files) **and** touches
  security-sensitive paths.
- **Route:** standard (not plans-only).
- **Selected lenses (9 of 9):** architecture (new hexagonal/DDD boundary + four Nx projects);
  docs-quality (65 `.md`, four new spec trees, plan archival); governance/rules-conformance
  (`repo-config.yml`, `repo-governance/**`, `docs/reference/**` normative edits); instruction-decay
  (new rule text and two self-serving exceptions carved in `target-naming-rules.md` and
  `mandatory-targets-cli-e2e.md`); integrity/test-gaming (`scripts/behaviour-coverage.mjs` extended
  by the same PR it scores, `integration-loopback` allowlist widened for `ose-id-be`, one
  `@e2e-exempt`, 99%+ coverage thresholds); business-logic/correctness (health/readiness/startup
  semantics, migration compatibility, capability-disabled policy); performance (local-stack runner,
  process lifecycle, E2E fixture/`dev`-server standard); security non-leak (runtime-mode fail-closed
  guard, route-disclosure guard, DB privilege separation, disabled-capability rejection);
  type-soundness (95 `.cs` + 27 `.ts`/`.tsx`).
- **Skipped lenses:** none. Content-Type Applicability Filter re-derived at this head: typed-language
  files present (types lens retained); test files and CI config present (integrity lens retained).
- **Out of scope for this pass:** the three leak predicates (secret-or-private-value,
  protected-environment-property, machine-specific-absolute-path) — owned by the separate
  authenticated `pr-leak-review` pass.
- **Current checks:** 17 passed, 0 failed at this head.
- **Settled history:** none. 0 reviews, 0 review comments, 0 issue comments on this PR; no prior
  cycle, no human dismissal, no fixer rejection to carry forward.
- **Probe class (cycle 1):** `claim-vs-artifact-truthfulness` — for each falsifiable capability claim
  this PR, its plan, and its specs make, does the shipped artifact at this head enforce it, or does
  only a test or a comment assert it? Not previously used on this PR.


wahidyankf (commented):
## Focused PR Leak Review (`pr-leak-review`, exact leak-only mode)

Scope: exactly three predicates only — secret-or-private-value, protected-environment-property,
machine-specific-absolute-path — across the complete aggregate base→head diff and delivery-controlled
PR metadata. No broad security review, no style, no other discipline's territory was performed or is
implied by this review.

**Result: findings (6)** — all `machine-specific-absolute-path`, all in raw build/tool-output evidence
files under `plans/done/2026-09-17__ose-id-init-01-foundation/evidence/`. No secret-or-private-value
and no protected-environment-property instance was found. This PR's earlier leaked absolute
home-directory path (commit `15d0aca7f`) was correctly redacted to `<home>` and does **not** recur;
these are separate, previously-unredacted instances of the same general defect class (a raw local
tool-output capture embedding a path rooted at a local installation prefix).

### Findings

1. **Category:** machine-specific-absolute-path
   **Location:** `plans/done/2026-09-17__ose-id-init-01-foundation/evidence/phase-2-gate/nx-quality.txt` — 8 occurrences (first at line 48)
   **Why:** Raw .NET analyzer-config log output embeds a literal absolute path rooted at
   a local Homebrew `dotnet` SDK installation prefix — the same category this repository's own
   convention names as a prohibited example (`/opt/homebrew/bin/node`) in
   "What Counts as Machine-Specific Information" → Absolute Local Paths.
   **Remediation:** Redact the SDK-install-prefix segment to a portable placeholder (e.g. the
   `<home>`-style redaction already applied elsewhere in this same PR), consistent across all
   occurrences in this file.

2. **Category:** machine-specific-absolute-path
   **Location:** `plans/done/2026-09-17__ose-id-init-01-foundation/evidence/phase-2-refactor/nx-quality.txt` — 8 occurrences (first at line 168)
   **Why:** Same pattern as Finding 1 — raw analyzer-config log line embedding the local Homebrew
   `dotnet` SDK install-prefix path.
   **Remediation:** Same as Finding 1.

3. **Category:** machine-specific-absolute-path
   **Location:** `plans/done/2026-09-17__ose-id-init-01-foundation/evidence/phase-2/nx-quality.txt` — 8 occurrences (first at line 316)
   **Why:** Same pattern as Finding 1.
   **Remediation:** Same as Finding 1.

4. **Category:** machine-specific-absolute-path
   **Location:** `plans/done/2026-09-17__ose-id-init-01-foundation/evidence/phase-3-quality-matrix/test-quick-ose-id-be-and-e2e.txt` — 4 occurrences (first at line 124)
   **Why:** Same pattern as Finding 1.
   **Remediation:** Same as Finding 1.

5. **Category:** machine-specific-absolute-path
   **Location:** `plans/done/2026-09-17__ose-id-init-01-foundation/evidence/phase-7/pre-push-gate.txt` — 4 occurrences (first at line 3241)
   **Why:** Same pattern as Finding 1.
   **Remediation:** Same as Finding 1.

6. **Category:** machine-specific-absolute-path
   **Location:** `plans/done/2026-09-17__ose-id-init-01-foundation/evidence/phase-7/pre-push-gate.txt` — line 664
   **Why:** A separate raw tool-output line in the same evidence file embeds an absolute macOS
   per-user temp-directory path (`$TMPDIR`-resolved, under `/var/folders/...`), which is unique to
   the local machine/user session that produced this evidence and is not reproducible across
   machines or CI.
   **Remediation:** Redact the temp-directory path segment to a portable placeholder; the
   surrounding bundling-tool message content can remain verbatim otherwise.

No candidate value is reproduced above; only category, location, and remediation are recorded, per
the workflow's sanitized-output contract.

plans/done/2026-09-17__ose-id-init-01-foundation/README.md
wahidyankf, line 1: ### P-F4 · MEDIUM · confidence 90

> _Anchored at file level: this file is a 100%-similarity rename in this PR, so GitHub exposes no diff hunk to attach a line comment to. The finding is at **line 3**._ — archived plan README still announces "In Progress … execution starts at Phase 0"

**Raised by**: `pr-review-docs-quality-maker` + `pr-review-governance-maker` — merged
(two independent specialists, one undisputed fact).

**What is wrong.** This PR archives the plan into `plans/done/`, and `delivery.md` records all
eight phases as `**Status**: Done`. Lines 3-4 of the README this PR is moving still read:

> **Status:** In Progress — first delivery in the OSE ID initialization chain. The lifecycle move to
> `plans/in-progress/ose-id-init-01-foundation/` has landed; execution starts at Phase 0.

Two things are stated that the same commit falsifies: the status, and a pointer to the
`plans/in-progress/` path this commit just vacated. "Execution starts at Phase 0" is the most
misleading part — a reader landing here from the PR body's own Reading Guide ("**Start here:**
`plans/done/2026-09-17__ose-id-init-01-foundation/README.md`") is told the work has not begun.

**Why MEDIUM and not CRITICAL.** Both specialists proposed CRITICAL at 92+ confidence. I checked
the rules they cited and they do not carry that weight:

- `repo-governance/conventions/structure/plans/starting-and-completing-work.md` lists six
  "Completing Work" steps (verify, resolve date, rename+move, update indexes, commit, archive) and
  **none** of them is "update the README status line". There is no MUST here to violate.
- `repo-governance/conventions/structure/plans/purpose-scope-and-overview.md:14` says the opposite
  of what a CRITICAL would need: *"File location (backlog/, in-progress/, done/) indicates status —
  no hidden metadata or databases required."* The authoritative status is the directory, and the
  directory is correct.
- The "established precedent" is real but not universal — most `plans/done/` READMEs do say
  Done/Complete, yet `plans/done/2026-03-29__skills-layer-implementation/`,
  `plans/done/2026-04-22__adopt-wahidyankf-web/` and `plans/done/2026-04-22__auth-register-login/`
  all still read "In Progress" on `main` today.

So: a real, self-contradicting documentation defect in the file the PR body tells reviewers to read
first — but nothing is blocked, and no MUST is breached. Per the Criticality Levels convention that
is MEDIUM ("suboptimal but functional"), not CRITICAL ("violations of MUST requirements").

**Fix**: update the status to Complete/Done with the completion date, and drop or past-tense the
`plans/in-progress/` sentence. The rest of the body is a legitimate historical record and should be
left alone.

**Refutation clause**: void if `grep -n 'In Progress' plans/done/2026-09-17__ose-id-init-01-foundation/README.md`
returns no match at the reviewed head.

---

Generated by AI

plans/done/2026-09-17__ose-id-init-01-foundation/tech-docs/006-api-contract-delta.md
wahidyankf, line 1: ### P-F9 · LOW · confidence 85

> _Anchored at file level: this file is a 100%-similarity rename in this PR, so GitHub exposes no diff hunk to attach a line comment to. The finding is at **line 39**._ — blanket "every response carries `X-Correlation-ID`" overstates the shipped behaviour

**Raised by**: `pr-review-docs-quality-maker`. Severity lowered from the raw MEDIUM — see below.

**What is wrong.** Line 39 states the guarantee without qualification:

> Every response carries a generated or validated `X-Correlation-ID`; no secret, connection string,
> database host, stack trace, or absolute machine path is returned.

The shipped behaviour is narrower: the header is emitted on the 7 documented operations'
correct-method responses and is deliberately **absent** on the framework's ordinary 404/405
fallback. A reader treating this document as the contract would expect a header the service does
not always send.

**Why LOW rather than MEDIUM.** This PR's own QA evidence already records the discrepancy, by line
number, at
`plans/done/2026-09-17__ose-id-init-01-foundation/evidence/phase-5/api-quality-gate/openapi-comparison.md:80-85`:

> `CorrelationId` header (`required: true`): held for all 7 operations' correct-method responses;
> **absent on the framework `405` fallback for the same 7 routes' wrong methods (AET-002)** — …
> the contract-wide prose in `tech-docs/006-api-contract-delta.md` (line 39/41) frames this as a
> blanket "every response" rule …

So this is a known, deliberately-taken and already-documented deviation, not an undetected false
claim — which is most of what separates this from a real contract defect. It is still worth a
one-line fix, because the evidence file that records it is one the PR body explicitly tells
reviewers to skip, and `openapi.yaml` — not this tech-doc — is the surface a consumer would trust.

**Fix**: scope the sentence, e.g. "Every response **from a documented operation** carries …", and
note the fallback exclusion.

**Refutation clause**: void if the service emits `X-Correlation-ID` on the not-found/wrong-method
fallback too. Executable check — start the local stack and
`curl -si -X POST http://127.0.0.1:8501/health/live | grep -i x-correlation-id`; this finding is
refuted if the header is present on that response. Equally refuted if line 39 already scopes the
claim to documented operations at the reviewed head.

---

Generated by AI

apps/ose-id-be-e2e/scripts/local-stack.mjs
```diff
@@ -0,0 +1,767 @@
...
+        }
+        reachedStageCount = 2;
+        log(runId, `backend ready (${owned.backends.map((backend) => backend.origin).join(", ")})`);
+
+        // Step 6: the web shell.
+        await buildWeb(runId);
+        webBuilt = true;
+        owned.web = startWeb(webPort, runId, nodeExecutable);
+        await waitForWebReady(owned.web.origin);
+        reachedStageCount = 3;
```
wahidyankf: ### P-F1 · CRITICAL · confidence 95 — cleanup skips the stage that was already spawned, leaking a process and its build output

**Raised by**: `pr-review-logic-maker` (correctness) + `pr-review-performance-maker` — merged.
One root cause, two consequences.

**What is wrong.** `cleanup()` decides what to tear down from `reachedStageCount` (line 660):

```js
const stages = ["postgres", "backend", "web"].slice(0, reachedStageCount).reverse();
```

But `reachedStageCount` is only incremented **after a stage's readiness wait succeeds**, while the
process is spawned **earlier in that same stage**. The window between "spawned" and "counted" is
not covered by cleanup:

- Line 741 `owned.web = startWeb(...)` spawns a **detached** web process → line 742
  `await waitForWebReady(...)` **throws on timeout** → line 743 `reachedStageCount = 3` never runs.
  Cleanup sees `2`, the stage list is `[backend, postgres]`, and `stopProcess(owned.web)` is never
  called. The web process survives the runner and keeps holding `webPort`.
- Lines 731-733: the backend loop pushes each spawned instance into `owned.backends` and *then*
  awaits readiness; `reachedStageCount = 2` is at line 735, after the whole loop. If **any**
  instance's readiness wait throws — including the only instance in a default `--instances=1`
  run — cleanup sees `1`, the stage list is `[postgres]`, and every already-spawned backend is
  left running. (The specialist reported this for `--instances=2` only; it is in fact reachable
  on the single-instance default path too.)

**Second consequence — leaked disk.** The directory removals live *inside* the same stage loop
(lines 672 and 674), so the same two paths leak build output on the same failure paths:

- `apps/ose-id-be/dist/local-stack/<runId>/` — written by `publishBackend()` at line 373, i.e.
  before the readiness wait that can throw.
- `apps/ose-id-web/.next/local-stack-runs/<runId>/` — written by `buildWeb()` before the same wait.

Neither is under `runDirectory` (an `mkdtemp` under `tmpdir()`, removed unconditionally at the end
of `cleanup()`), so the unconditional `rm(runDirectory)` does not cover them, and there is no
sweep or TTL for either path anywhere in the repo. Each failed run leaves a full Release publish
tree and a full Next build behind, permanently.

**Why this matters.** `specs/apps/ose/id-be/behaviours/foundation/local-stack.feature:13` states
the accepted criterion: *"stopping the runner leaves no owned process, container, network, volume,
or port reservation"*. A readiness timeout is a budgeted, ordinary outcome (`STARTUP_BUDGET_MS`
exists precisely because readiness can time out) — not an exotic path. And the leak is
self-propagating: the orphan keeps `webPort`/`OSE_ID_BE_PORT` bound, so the *next* run hits the
port-collision guard at the top of `main()` and refuses to start, in a way that looks like an
unrelated environment problem.

**No existing test covers this window.** `UnknownFixtureProfileFailsAfterReadinessAndCleansUpEverything`
(`steps/LocalStackRunnerTests.cs:66`) throws only after `reachedStageCount` has already reached 3,
and `TwoInstancesBothBackendsBecomeReadyAndBothStop` (line 111) is a success path. Both pass while
the defect is live.

**Reproduction / verification performed.** Evaluating the exact line-660 expression at each value
the counter can hold:

```
reachedStageCount=0 -> stages: []                       | web stopped? false | backend stopped? false
reachedStageCount=1 -> stages: [postgres]               | web stopped? false | backend stopped? false
reachedStageCount=2 -> stages: [backend, postgres]      | web stopped? false | backend stopped? true
reachedStageCount=3 -> stages: [web, backend, postgres] | web stopped? true  | backend stopped? true
```

Rows 1 and 2 are exactly the two states reachable when a readiness wait throws, and in both the
already-spawned process is excluded from teardown.

**Worth noting: this file already solved this exact problem once.** The `webBuilt` flag at line 686
is gated independently of `reachedStageCount`, with a comment explaining why —
*"Gated on the build actually running, not on `reachedStageCount`: `next build` already appended
this run's tsconfig entries even when the later start/readiness step is what ends up failing"*.
That is precisely the reasoning that process and directory teardown still need. Suggested shape:
track ownership, not stage progress — tear down anything present in `owned` (and any run-scoped
directory already created), rather than slicing a stage list by a success counter.

**Refutation clause**: void if `reachedStageCount` is assigned before, or independently of, the
readiness-wait calls. Executable check — `grep -n -B4 'reachedStageCount = [23]' apps/ose-id-be-e2e/scripts/local-stack.mjs`;
this finding is refuted if no `await waitFor*Ready` call appears between the spawn and the
assignment. Equally refuted by a test that kills or stalls the web process before readiness and
then asserts no listener remains on `webPort` and that both run-scoped directories are gone.

---

Generated by AI

repo-governance/development/infra/nx-targets/target-naming-rules.md
```diff
@@ -14,3 +14,8 @@ when_to_use: "Use when naming or reviewing an Nx target."
 - Use `test:quick` for the closed fast composition; never include Integration/E2E runtime.
 - Separate variants with a colon and use lowercase kebab-case within each segment.
 - Omit an inapplicable target; never provide an echo, no-op, sentinel, duplicate, or alias.
+- Exception: an E2E project with no app of its own may use `serve` for a target that orchestrates
+  multiple owned processes (for example a multi-process local stack with selectable fixture
+  profiles) for manual verification. This is not a `dev`/`start` alias — the project has neither —
+  and the E2E project explicitly needs a name distinct from `test:e2e`'s automated Gherkin run.
+  Example: `ose-id-be-e2e:serve`.
```
wahidyankf, line 21: ### P-F2 · HIGH · confidence 92 — the new `serve` exception was never propagated to the surfaces that still forbid `serve` outright

**Raised by**: `pr-review-instruction-decay-maker` (two raw findings, merged — one failure to run
the subject-scoped tidy).

**What is wrong.** This diff adds a sanctioned `serve` exception for E2E projects (lines 17-21,
worked example `ose-id-be-e2e:serve`) to the canonical naming-rules document. Two other live rule
surfaces still state a **flat, unqualified** prohibition and were not touched:

1. `.claude/skills/swe-developing-applications-common/SKILL.md:21` — *"Use canonical
   [Nx Target Standards] names (`dev`, `test:quick`, `start`, never `serve`/`test`)"*. This is a
   primary-binding skill that agents consult when scaffolding, and it cites `nx-targets.md` — the
   very document this diff amended — as its authority. Its `.agents/skills/` mirror carries the
   same text.
2. `repo-governance/development/infra/nx-targets/target-naming-canonical-names-e2e-and-utility.md:10`
   — *"Aliases (`serve`, `start:dev`, `unit-test`) are anti-patterns."* This is the **E2E and
   utility target reference**, i.e. the single most likely document an author of an E2E project
   consults, and it is a sibling file in the same `nx-targets/` directory as the file this diff
   edited.

**The rule this violates.** `repo-governance/workflows/rules/rules-propagation.md` Step 6 requires
that when a rule is written, *"every other surface stating its subject is brought into
agreement"*, with an explicit verdict (Keep/Amend/Merge/Delete/Relocate/Supersede) recorded for
each — and it warns to *"fix the class, not the sites one search happened to name"*. A rule change
that lands on one surface and leaves two contradicting it is the exact failure that step exists to
prevent.

**Consequence in plain terms.** An agent or contributor scaffolding the next `*-be-e2e` project
reads "never `serve`", refuses the target this PR just sanctioned, and either invents a
non-canonical name or reverts `ose-id-be-e2e:serve`. The repo now answers the same question two
different ways depending on which file you open first.

**Tool-verify correction to the raw findings.** The specialist also named
`.claude/skills/swe-developing-applications-common/reference/nx-monorepo-integration.md:23` and
`repo-governance/development/infra/nx-target-naming/principles-conventions-and-lifecycle-targets.md:40-41`.
I read both and **they need no change**: each is scoped to *aliasing* — "`dev` (not `serve`)",
"`start` for production server — never `serve`" — which is precisely the case the new exception
carves itself out of ("This is not a `dev`/`start` alias — the project has neither"). They are
consistent with the exception as written. Only the two surfaces listed above are genuinely
contradicted.

**Refutation clause**: void if the two named surfaces already carry the exception at this head.
Executable check — `grep -n 'serve' .claude/skills/swe-developing-applications-common/SKILL.md repo-governance/development/infra/nx-targets/target-naming-canonical-names-e2e-and-utility.md`;
refuted if the prohibition at each site is qualified (or cross-references this exception) rather
than absolute.

---

Generated by AI

apps/ose-id-be/tests/unit/Tests/ArchitectureBoundaryTests.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    [Fact]
+    public void Application_DeclaresOnlyTheDomainProjectReference()
+    {
+        string csproj = ReadCsproj("OseId.Application");
+
+        csproj
+            .Should()
+            .NotContain("<PackageReference", "a use case exposes no ASP.NET, EF, OpenIddict, GraphQL, or MCP type");
+        csproj.Should().Contain("""<ProjectReference Include="../OseId.Domain/OseId.Domain.csproj" />""");
```
wahidyankf, line 51: ### P-F3 · MEDIUM · confidence 88 — `Application_DeclaresOnlyTheDomainProjectReference` never checks "only"

**Raised by**: `pr-review-architecture-maker`. Category confirmed as **architecture** on
re-categorization — see the note at the end.

**What is wrong.** The test promises an exclusivity property in its name, and the class doc comment
above it claims the boundary is *"a build-enforced property, not a convention"*. The body is two
independent substring checks (lines 48-51):

```csharp
csproj.Should().NotContain("<PackageReference", ...);
csproj.Should().Contain("""<ProjectReference Include="../OseId.Domain/OseId.Domain.csproj" />""");
```

"Contains the Domain reference" and "contains no PackageReference" together do not imply "contains
**only** the Domain reference". A second `<ProjectReference>` — to a future `OseId.Infrastructure`,
`OseId.Host`, or `OseId.Migrator` — satisfies both assertions and the test stays green.

The reflection-based companion (`Application_AssemblyReferencesNoFrameworkOrDataOrTransportType`)
does not close the gap either: `ForbiddenAssemblyPrefixes` (lines 15-24) lists only third-party
framework prefixes (`Microsoft.AspNetCore`, `Npgsql`, `SqlKata`, `OpenIddict`, …) and no
repo-internal `OseId.*` sibling name.

**Why it matters now rather than someday.** Today the violation is masked by MSBuild's own
circular-reference detection — but that is an accident of the current topology (every sibling
already references Application transitively), not an assertion this suite makes. The protection
disappears the moment a project **without** a back-edge joins the graph, which is exactly what the
"GraphQL/MCP delivery — architectural seam only" work this PR's own Scope section anticipates would
add. The test would then silently stop guarding the thing it is named after.

**Suggested shape.** Bound the count rather than testing membership, e.g. assert that the number of
`<ProjectReference` occurrences in the csproj is exactly 1, or add `OseId.` sibling names (other
than `OseId.Domain`) to the forbidden-prefix list for the Application assembly.

**Re-categorization note.** Applying the boundary tie-breaker: no existing `repo-governance/` rule
states how strictly a hexagonal-boundary guard must be mechanically enforced, so step 1
(governance) does not apply; deciding "presence-based assertion vs count-bounded assertion" is a
genuinely new structural-enforcement tradeoff, so step 2 applies and this stays **architecture**,
not correctness. Per the tie-breaker, the resolution should be written down as a rule so the next
occurrence falls under step 1.

**Refutation clause**: void if the test counts or bounds `<ProjectReference` occurrences rather
than using independent `Contain`/`NotContain` calls, or if the forbidden-prefix list covers
repo-internal `OseId.*` siblings. Executable check — add a second `<ProjectReference` to
`apps/ose-id-be/src/OseId.Application/OseId.Application.csproj` pointing at a project with no
back-edge and run `ose-id-be:test:unit`; this finding is refuted if the suite fails.

---

Generated by AI

docs/reference/monorepo-structure.md
```diff
@@ -99,6 +99,11 @@ Flat structure - all apps at the same level, no subdirectories.
 - `organiclever-be-e2e` - Playwright BE E2E tests for organiclever-be
 - `roots-be` - Roots Islamic-practice tooling Go/Gin REST API backend (port 8402)
 - `roots-be-e2e` - Playwright BE E2E tests for roots-be
+- `ose-id-be` - OSE ID C#/ASP.NET Core REST API backend (port 8501)
+- `ose-id-be-e2e` - Reqnroll/xUnit BE E2E tests for ose-id-be (C#, not Playwright — see
+  `nx-targets/mandatory-targets-cli-e2e.md`)
```
wahidyankf: ### P-F5 · MEDIUM · confidence 90 — new cross-reference points at a path that does not exist from this file

**Raised by**: `pr-review-governance-maker`.

**What is wrong.** Line 104, added by this diff, sends the reader to:

> (C#, not Playwright — see `nx-targets/mandatory-targets-cli-e2e.md`)

Resolved from this file's own location that is `docs/reference/nx-targets/mandatory-targets-cli-e2e.md`,
and `docs/reference/nx-targets/` does not exist. The document actually lives at
`repo-governance/development/infra/nx-targets/mandatory-targets-cli-e2e.md`.

**Why the link gate did not catch it.** It is a bare backtick string, not Markdown link syntax, so
`md-links` never resolves it. A reader gets a dead pointer that no CI job will ever report.

**Consequence.** The one line in the whole document that explains *why* `ose-id-be-e2e` is the
only non-Playwright `*-e2e` project sends the reader nowhere — and this is precisely the
distinction a future contributor scaffolding an E2E project needs to find.

**Fix**: make it a real relative link, matching what this file already does for the parent
document: `[Mandatory Targets — CLI and E2E](../../repo-governance/development/infra/nx-targets/mandatory-targets-cli-e2e.md)`.

**Tool-verify correction to the raw finding.** The specialist supported this by claiming the same
file "already uses correct relative markdown links elsewhere for the same target doc (lines 358,
570)". That is not accurate — lines 358 and 570 link to the **parent** document
`../../repo-governance/development/infra/nx-targets.md`, not to this child file, which line 104 is
the only reference to. The in-file convention argument still holds (both existing references use
proper relative-link syntax rooted at `../../repo-governance/`), but the "same doc" part of the
claim does not, and the corrected version is what is posted here.

**Refutation clause**: void if the path resolves. Executable check — `ls docs/reference/nx-targets/mandatory-targets-cli-e2e.md`;
this finding is refuted if that file exists.

---

Generated by AI

wahidyankf (commented):
## PR Review — Single Pass

**Head SHA**: d89ec62ef144d74f7d7424ce7047cc7de499a9b4
**Base**: `main` at 284504b689caf7e77dedd75c529f63a2069665df
**Risk tier**: full
**Probe class**: claim-vs-artifact-truthfulness
**Specialists fanned out**: architecture, business-logic (correctness), docs-quality, governance, instruction-decay, integrity, performance, security, type-soundness
**Per-specialist raw findings**: architecture 1 · business-logic 1 · docs-quality 3 · governance 2 · instruction-decay 3 · performance 3 · integrity 0 (fired, clean) · security 0 (fired, clean, ordinary mode) · type-soundness 0 (fired, clean). Total raw 13 → 9 posted. No specialist was skipped by the Content-Type Applicability Filter this cycle.
**Diff coverage**: full diff (three-dot against merge-base `d9a832b6a49d7f5587dd63061bb45827a6a0b75e`; 383 files, +39,892/−778)
**Prior settled findings respected**: none (0 prior semantic reviews at scout time)

### This review is blocking despite its `COMMENT` state

GitHub refuses `REQUEST_CHANGES` on one's own pull request, and this review posts under the PR
author's `gh` identity. Blocking status is carried by severity: **1 CRITICAL and 1 HIGH must be
resolved or refuted before merge.** The PR is currently a **draft**, so nothing is being blocked
right now — but the CRITICAL is not a "before you undraft" nicety, it falsifies a shipped Gherkin
acceptance criterion.

### Severity counts

| CRITICAL | HIGH | MEDIUM | LOW |
| -------- | ---- | ------ | --- |
| 1 | 1 | 5 | 2 |

### Synthesis decisions worth knowing about

Three merges and one drop happened here, plus two corrections to specialists' own supporting
evidence. Stating them so the record is auditable rather than implied:

- **Merged (one root cause, two consequences)** — correctness and performance independently
  reported the same `reachedStageCount` gating defect in `local-stack.mjs` with different blast
  radii (leaked processes / leaked build directories). Posted once, as **P-F1**.
- **Merged (two specialists, one fact)** — docs-quality and governance independently reported the
  stale `In Progress` banner on the archived plan README. Posted once, as **P-F4**.
- **Merged (one unpropagated rule change)** — instruction-decay reported the un-propagated `serve`
  exception twice, split by surface type. It is one failure to run the subject-scoped tidy. Posted
  once, as **P-F2**.
- **Dropped by the reasonableness filter** — docs-quality's `apps/ose-id-be/README.md:43`
  "`capability_disabled`-shaped" finding. The suffix `-shaped` is load-bearing and correct: it
  claims the *problem-body shape*, not the code value, and line 48 of the same README writes
  "`capability_disabled` **code**" when it does mean the code. The distinction is deliberate, not
  accidental. Post-verification confidence fell below the 80 bar.
- **Severity lowered, CRITICAL → MEDIUM (P-F4)** — two specialists at 92+ confidence both proposed
  CRITICAL for the stale README banner. Agreement is not severity. I read the two rules they cited
  and neither supports it: `plans/starting-and-completing-work.md` lists six completion steps and
  *none* of them is "update the README status", and
  `plans/purpose-scope-and-overview.md:14` says outright that "File location (backlog/,
  in-progress/, done/) indicates status — no hidden metadata". There is also no unanimous precedent:
  `plans/done/2026-03-29__skills-layer-implementation/`,
  `plans/done/2026-04-22__adopt-wahidyankf-web/` and `plans/done/2026-04-22__auth-register-login/`
  all still say "In Progress" today. Real defect, MEDIUM severity.
- **Two specialist evidence claims corrected during tool-verify** — see P-F2 and P-F5; in both cases
  the core defect survived but a supporting claim did not, and the narrowed version is what is
  posted.

### Confirmed sound (no finding raised)

Integrity, security, and type-soundness each fired and returned clean **by execution and
exhaustive grep, not by reading claims**: the C#/Reqnroll behaviour-coverage extractor, the
`integration-loopback` allowlist widening, the 99.58% / 27-27 / 86-87 numbers, the
`RuntimeModeStartupGuard` / `RouteDisclosureGuard` fail-closed enforcement, the DB privilege
separation, and the "EF confined to migration time" claim all hold up. For a PR of this size that
is worth stating explicitly.

### Related, separate review track

An authenticated `pr-leak-review` (review-id `5230479866`) already posted findings on this same
head — 6 machine-specific absolute paths under
`plans/done/2026-09-17__ose-id-init-01-foundation/evidence/*.txt`. That is the leak-only track and
is **not** duplicated here. It still needs its own resolution.



### Where the nine findings live

Seven findings are line-anchored inline threads on this review. **P-F4** and **P-F9** could not be
line-anchored: both files are 100%-similarity renames in this PR, so GitHub exposes no diff hunk to
attach a line comment to. They are posted as **file-level review threads on this same head** —
equally resolvable, just not line-pinned:

- **P-F4** (MEDIUM) `plans/done/2026-09-17__ose-id-init-01-foundation/README.md` — [thread](https://github.com/wahidyankf/ose-public/pull/539#discussion_r4032753016)
- **P-F9** (LOW) `plans/done/2026-09-17__ose-id-init-01-foundation/tech-docs/006-api-contract-delta.md` — [thread](https://github.com/wahidyankf/ose-public/pull/539#discussion_r4032753468)

All nine are counted in the severity table and the machine record above.



---

Generated by AI

apps/ose-id-web/next.config.ts
```diff
@@ -0,0 +1,45 @@
...
+// copy), so Next's automatic root detection is ambiguous. Pin it explicitly.
+const workspaceRoot = path.resolve(process.cwd(), "../..");
+
+const nextConfig: NextConfig = {
+  outputFileTracingRoot: workspaceRoot,
+  // The local-stack runner (apps/ose-id-be-e2e/scripts/local-stack.mjs) may build and serve two
+  // independent runs from this same checkout at once; each sets this to its own run-scoped
+  // subdirectory of the already-ignored .next/ so a later build never overwrites an
+  // already-serving run's files. Absent, this is the ordinary shared .next/ every other build uses.
+  distDir: process.env.OSE_ID_WEB_DIST_DIR ?? ".next",
```
wahidyankf, line 29: ### P-F6 · MEDIUM · confidence 88 — run-scoping `distDir` also throws away Next's build cache, every run

**Raised by**: `pr-review-performance-maker`.

**What is wrong.** `distDir: process.env.OSE_ID_WEB_DIST_DIR ?? ".next"` lets the local-stack runner
point the build output at `.next/local-stack-runs/<runId>/`, a fresh directory per invocation. The
comment above explains the reason and the reason is sound — two concurrent runs must not overwrite
each other's served files.

The cost is that Next nests its **persistent build cache inside `distDir`**. Verified directly
against the installed Next 16.2.6 in this worktree:

```js
// node_modules/next/dist/build/index.js:171
const cacheDir = _path.default.join(distDir, 'cache');
```

A fresh `distDir` is therefore a cold cache. Every local-stack invocation pays a full production
build from scratch even when nothing in `apps/ose-id-web` changed since the last one — and this
PR's own E2E suite invokes the runner several times per run.

**The backend side of the same runner gets this right**, which is what makes this a defect rather
than an unavoidable tradeoff: `publishBackend()` run-scopes only the `-o` publish *output*, leaving
`obj/`/`bin/` shared, so incremental compilation still works. The web side scoped the
cache-bearing directory itself.

**On the fix — deliberately not prescribing a mechanism.** The specialist recommended redirecting
the cache via `NEXT_PRIVATE_CACHE_DIR`. I grepped `node_modules/next/dist` and **that variable
does not exist in Next 16.2.6**, so it is not an available fix and this review does not ask for it.
The finding is the cost, not a specific API: viable directions are run-scoping only the served
output while leaving the cache at a stable shared path, or accepting the cost with a recorded
rationale in the comment at lines 20-23 so the next reader knows it was weighed rather than missed.

**Refutation clause**: void if Next stores its build cache outside `distDir` at the pinned
version. Executable check — `grep -rn "join(distDir, 'cache')" node_modules/next/dist/build/index.js`;
this finding is refuted if no such join exists. Equally refuted by a timing measurement showing two
consecutive no-source-change local-stack invocations take comparable wall time.

---

Generated by AI

apps/ose-id-be-e2e/scripts/local-stack.mjs
```diff
@@ -0,0 +1,767 @@
...
+ * running instance is still serving from.
+ */
+function backendPublishDirectory(runId) {
+    return path.join(REPO_ROOT, "apps", "ose-id-be", "dist", "local-stack", runId);
+}
+
+async function publishBackend(runId) {
+    const project = path.join(REPO_ROOT, "apps", "ose-id-be", "src", "OseId.Host", "OseId.Host.csproj");
+    const output = backendPublishDirectory(runId);
+    const result = await runCommand("dotnet", ["publish", project, "-c", "Release", "-o", output], {
```
wahidyankf: ### P-F7 · MEDIUM · confidence 82 — concurrent runs share unscoped MSBuild intermediate state

**Raised by**: `pr-review-performance-maker`.

**What is wrong.** `publishBackend()` run-scopes the `-o` output directory but nothing else. The
project path, and therefore `obj/` and `bin/` for `OseId.Host.csproj`, is shared across every
concurrent invocation from the same checkout:

```js
const result = await runCommand("dotnet", ["publish", project, "-c", "Release", "-o", output], ...);
```

`dotnet publish` performs an **implicit restore**, which writes `obj/project.assets.json`. That
file sits at the root of the intermediate path and is not configuration-scoped, so two concurrent
publishes of the same project read and rewrite the same file. This is a documented hazard class,
independently re-verified for this review rather than taken on the specialist's word — see
[NuGet/Home#12973](https://github.com/NuGet/Home/issues/12973) on shared `project.assets.json` and
[dotnet/sdk#10046](https://github.com/dotnet/sdk/issues/10046) on its absolute-path coupling
between build and publish stages.

*(The raw finding cited `dotnet/sdk#16586`, `dotnet/sdk#52773` and `dotnet/msbuild#8030`. I could
not verify those three, so they are not cited here. Only the two links above were confirmed.)*

**Why this is not hypothetical for this PR.** `TwoConcurrentRunsUseIndependentRunIdsAndBothCleanUpFully`
(`apps/ose-id-be-e2e/steps/LocalStackRunnerTests.cs:155`) starts both runners before waiting on
either — its own comment says *"Both runs proceed genuinely concurrently"* — so the PR ships a test
that deliberately drives two simultaneous `dotnet publish` invocations against the same `obj/`.

**Consequence.** An intermittent `NETSDK1004`/assets-file failure in a test that advertises
concurrency safety, appearing as a flake. Under this repo's own rule that a flaky test is a defect
to fix at root cause, that is worth closing before it is observed rather than after.

**Confidence is 82, not higher, deliberately**: the race is real and the shared state is real, but
no failure has been observed here — a green run does not disprove an intermittent race, and equally
does not demonstrate one. Straightforward mitigations exist: a dedicated
`--property:BaseIntermediateOutputPath` per run, or an explicit `dotnet restore` before the
concurrent phase plus `--no-restore` on publish.

**Refutation clause**: void if the publish already scopes intermediate output per run. Executable
check — `grep -n 'BaseIntermediateOutputPath\|IntermediateOutputPath\|no-restore' apps/ose-id-be-e2e/scripts/local-stack.mjs`;
this finding is refuted if the publish invocation sets a per-run intermediate path or suppresses
the implicit restore. Also refuted if `OseId.Host.csproj` or `Directory.Build.props` already
redirects `BaseIntermediateOutputPath` per run.

---

Generated by AI

repo-governance/development/infra/nx-targets/mandatory-targets-cli-e2e.md
```diff
@@ -37,3 +37,20 @@ See `apps/organiclever-be-e2e/project.json` for a canonical product-app example.
 An E2E project also exposes `test:coverage:e2e` and `test:coverage:behaviour`. These validators
 statically prove adapter completeness and exemption validity without running Playwright. The owning
 corpus is an explicit Nx input of both coverage and runtime targets.
+
+### Dotnet/Reqnroll E2E Test Projects
```
wahidyankf, line 42: ### P-F8 · LOW · confidence 85 — the file's discovery metadata still says Playwright-only after this diff adds a non-Playwright section

**Raised by**: `pr-review-instruction-decay-maker`.

**What is wrong.** This diff adds `### Dotnet/Reqnroll E2E Test Projects` (lines 41-56) under the
existing `## E2E Test Projects` heading. Three pieces of framing above it were not updated and now
under-describe the file:

- Frontmatter `description` (line 2) — enumerates only "the install/test:e2e/test:e2e:ui/test:e2e:report
  targets required on `*-e2e` projects", which is the Playwright set; the new Reqnroll table
  explicitly *omits* `install` and `test:e2e:ui`/`test:e2e:report` and adds `test:coverage:e2e`,
  `test:coverage:behaviour` and `serve`.
- Frontmatter `when_to_use` (line 3) — *"Use when scaffolding a new CLI application or a new `*-e2e`
  **Playwright runner** project."*
- Line 19, directly under the `## E2E Test Projects` H2 the new subsection now sits inside —
  *"Playwright suites (`*-e2e`):"*.

**Why this is worth a line rather than nothing.** `when_to_use` is the routing surface — it is how
an agent decides whether to open this file at all. An agent scaffolding the next Reqnroll-based
`*-be-e2e` project reads "Playwright runner project", concludes this file does not apply, and never
reaches the section written specifically for it. The content is correct; only its findability is
stale.

**Fix**: widen `description` and `when_to_use` to cover both runner families, and qualify line 19
so "Playwright suites" scopes the table beneath it rather than the whole H2.

**Refutation clause**: void if the frontmatter or the line-19 framing already names a non-Playwright
E2E runner. Executable check — `sed -n '1,4p;17,20p' repo-governance/development/infra/nx-targets/mandatory-targets-cli-e2e.md`;
this finding is refuted if either `description`/`when_to_use` or the heading lead-in mentions
Reqnroll, dotnet, or otherwise admits a non-Playwright runner.

---

Generated by AI

wahidyankf (commented):
## Focused PR Leak Review (`pr-leak-review`, exact leak-only mode)

Scope: exactly three predicates only — secret-or-private-value, protected-environment-property,
machine-specific-absolute-path — across the complete aggregate base→head diff and delivery-controlled
PR metadata. No broad security review, no style, no other discipline's territory was performed or is
implied by this review.

This is a fresh full-diff pass at a new pinned head, not a reuse of the prior authenticated result
(review-id `5230479866`, head `d89ec62ef144d74f7d7424ce7047cc7de499a9b4`). Four commits have landed
since that head: `d72bf5e47` (the redaction fix), `fd0e4513a`, `0fed0f395`, and `013e6c15d`.

**Result: pass** — 0 findings in all three categories.

- **Prior 6 machine-specific-absolute-path findings verified resolved.** All five previously-flagged
  evidence files under `plans/done/2026-09-17__ose-id-init-01-foundation/evidence/` (`phase-2-gate/nx-quality.txt`,
  `phase-2-refactor/nx-quality.txt`, `phase-2/nx-quality.txt`, `phase-3-quality-matrix/test-quick-ose-id-be-and-e2e.txt`,
  `phase-7/pre-push-gate.txt`) now carry portable `<dotnet-sdk-prefix>`/`<tmp>` placeholders in place of
  the raw `/opt/homebrew/...` and `/var/folders/...` paths, matching the pre-existing `<home>` redaction
  style. A full-diff grep for `/Users/`, `/opt/homebrew`, `/var/folders`, `/private/(tmp|var)`, `/root/`,
  and `C:\Users` style paths across all 386 changed files returned zero real hits (one `/usr/local/bin/clang-format`
  string is the OpenAPI Generator tool's own generic example text in its log output, not a real
  machine-specific path from this environment).
- **The three post-redaction commits reviewed fresh** (not covered by any prior leak pass):
  `fd0e4513a` (`apps/ose-id-be-e2e/scripts/local-stack.mjs` + `LocalStackRunnerTests.cs` — ownership-based
  cleanup tracking, ephemeral/loopback ports only, one synthetic unreachable test DSN
  `Host=127.0.0.1;Port=1;...;Password=ose_id_test` matching its own test-name convention), `0fed0f395`
  (`ArchitectureBoundaryTests.cs` — test logic only, no paths/values), and `013e6c15d` (documentation/governance
  text only, plus the two `swe-developing-applications-common` skill mirrors). No candidate in any of the
  three.
- **No secret-or-private-value.** No private-key block, no AWS/API-key-shaped credential, no
  high-entropy token outside known-public identifiers (git commit SHAs, a source-file SHA-256 in the
  Rhino CLI parity manifest, xUnit test-run session IDs).
- **No protected-environment-property.** The only `.env`-shaped diffs are two new `.env.example`
  files (`apps/ose-id-be/.env.example`, `apps/ose-id-web/.env.example`) using `Password=change-me`
  placeholders and a `localhost` dev connection string — explicitly permitted example content, and
  this PR's own scope statement confirms no production surface exists yet for this plan.
- **PR title/body inspected** — narrative only (reserved local ports, commit SHAs, file paths, prose);
  no candidate value.

No candidate value is reproduced above; only category, location, and remediation status are recorded,
per the workflow's sanitized-output contract. Every other security and semantic concern was out of
scope for this pass.

plans/done/2026-09-17__ose-id-init-01-foundation/README.md
wahidyankf: Fixed in 013e6c15d. Corrected the two false specifics (dead `plans/in-progress/` pointer and "execution starts at Phase 0") rather than flipping to "Complete" as suggested — at fix time Phase 7 (this review) was still open, so "Complete" would have been a new false claim. Now reads "Phases 0-6 are complete and Phase 7 ... is underway." Will do the final Complete flip once merge and Phase 8 actually finish.

plans/done/2026-09-17__ose-id-init-01-foundation/tech-docs/006-api-contract-delta.md
wahidyankf: Fixed in 013e6c15d. Scoped the sentence to "Every response from a documented operation carries..." and noted the deliberate 404/405 fallback exclusion.

apps/ose-id-be-e2e/scripts/local-stack.mjs
wahidyankf: Fixed in fd0e4513a. cleanup() now derives its stage list from actual ownership (owned.postgres/owned.backends/owned.web) instead of reachedStageCount, with build-output removal gated on flags that flip before the work that creates each directory. Added a deterministic fixture-profile test seam (no timing races) and two regression tests (BackendReadinessFailureStillStopsTheSpawnedBackendAndRemovesItsRunOutput, WebReadinessFailureStillStopsTheSpawnedWebProcessAndRemovesBothRunOutputs) — both proven red without the fix (orphaned process + leaked directory confirmed by hand) and green with it. Full LocalStackRunnerTests suite: 29/29 passing.

repo-governance/development/infra/nx-targets/target-naming-rules.md
wahidyankf: Fixed in 013e6c15d. Propagated the `serve` exception to the two genuinely-contradicted surfaces you identified: `.claude/skills/swe-developing-applications-common/SKILL.md` (mirror regenerated via `npm run generate:bindings`, confirmed identical text) and `target-naming-canonical-names-e2e-and-utility.md`. Left the two aliasing-scoped docs untouched per your correction.

apps/ose-id-be/tests/unit/Tests/ArchitectureBoundaryTests.cs
wahidyankf: Fixed in 0fed0f395. Added a count assertion (exactly one ProjectReference) to Application_DeclaresOnlyTheDomainProjectReference. Verified the gap was real and the fix closes it: temporarily added a second ProjectReference to a sibling with no back-edge — the suite passed without this fix and failed with it, on the new assertion specifically.

docs/reference/monorepo-structure.md
wahidyankf: Fixed in 013e6c15d. Replaced the bare backtick reference with a real relative link: `[Mandatory Targets — CLI and E2E](../../repo-governance/development/infra/nx-targets/mandatory-targets-cli-e2e.md)`.

apps/ose-id-web/next.config.ts
wahidyankf: Addressed in 013e6c15d via the documented-rationale option you offered rather than a cache-relocation mechanism: confirmed `NEXT_PRIVATE_CACHE_DIR` doesn't exist in this Next version as you found, and a shared cache directory would reintroduce the concurrent-write hazard distDir scoping exists to avoid. Added an in-line comment recording the tradeoff was weighed, not missed.

apps/ose-id-be-e2e/scripts/local-stack.mjs
wahidyankf: Fixed in fd0e4513a. Scoped via `--property:ArtifactsPath=<run-scoped dir>` rather than the suggested `BaseIntermediateOutputPath` — verified BaseIntermediateOutputPath is actually wrong for this project graph (a command-line property propagates to every referenced project, so Host/Domain/Application/Infrastructure would collide on one shared intermediate directory and the build fails with CS0579). ArtifactsPath propagates the same way but gives each project its own obj/bin subdirectory. Verified clean publish and full suite (29/29) including TwoConcurrentRunsUseIndependentRunIdsAndBothCleanUpFully.

repo-governance/development/infra/nx-targets/mandatory-targets-cli-e2e.md
wahidyankf: Fixed in 013e6c15d. Widened frontmatter `description`/`when_to_use` to cover both runner families and qualified the "Playwright suites" lead-in with a pointer to the Dotnet/Reqnroll subsection.

wahidyankf (commented):
## Focused PR Leak Review (`pr-leak-review`, exact leak-only mode)

Scope: exactly three categories only — secret-or-private-value, protected-environment-property,
machine-specific-absolute-path — inspected across the full three-dot base→head diff (base `main`
@ `d576c60f4ee3a72118fb7091193530e933ccc38a`, head @ `0fae918c4fe929890c2a0d5c0e6164cebbb55752`;
386 changed files) plus delivery-controlled PR title/body. No other security or semantic concern
was in scope for this pass.

This is a fresh pass at a new head: the branch was brought forward with a merge from `origin/main`
since the prior authenticated pass (review-id 5231367941, pre-merge head
`013e6c15deff332edfe912dbdbdcec65c007db07`). The merge's own base is now this PR's current base
(`main` @ `d576c60f4ee3a72118fb7091193530e933ccc38a`), so the five newly-merged main commits
(`.claude/hooks/require-hippo-boundary.sh`/`.test.sh`, `.claude/settings.json`,
`.codex/hooks.json`, `.opencode/opencode.json`, `docs/how-to/setup-development-environment.md`,
`plans/ideas/**`, `repo-config.yml`, `repo-governance/**` docs, `scripts/public-safety/**`,
`social-media-posts/linkedin/**`) land inside the merge-base and are byte-identical between base
and head at this head — verified directly (`git diff <base> <head> -- <path>` = 0 lines for each)
rather than inferred from diff-header absence. `repo-config.yml` does carry 85 lines of real
diff at this head, but that diff is this PR's own governance-config content (pre-commit HIPPO
guard, `integration-loopback` allowlist entry for `ose-id-be`, `csharpier` command form), not
main-only content; it was inspected and is clean.

Result: **pass** — no finding in any of the three categories.

Coverage notes: 134 new files under
`plans/done/2026-09-17__ose-id-init-01-foundation/evidence/**` (incl. 55 PNG screenshots) were
inspected; sampled screenshots across all five tester subfolders (design/exploratory/manual-
browser/usability, light/dark, multiple breakpoints) render only the static local status-page
copy with no browser chrome, URL bar, or diagnostic path visible. The one binary-flagged source
file (`apps/ose-id-be/tests/unit/Tests/CorrelationIdTests.cs`) was decoded directly: the binary
marker is a literal NUL byte inside an intentional `[InlineData]` rejection-fixture string, not
encoded/obscured content. All Postgres passwords/connection strings introduced by the local-stack
runner and its C# robustness tests are per-run `randomBytes`/`Guid.NewGuid()`-generated ephemeral
credentials for an owned, loopback-only container (one is a deliberately syntactically-valid-but-
unreachable placeholder on port 1, documented as such in-line) — synthetic fixtures, not real
secrets. `.env.example` additions use `change-me` placeholders only. `package-lock.json` additions
are standard public-npm-registry resolutions with no registry-auth tokens. No `/Users/`, `/home/
<user>`, Windows-drive, `/private/var`, or other machine-specific absolute path pattern was found.
No production/staging hostname, credential, or protected property was found; the only prod-*/
`oseplatform.com`/`organiclever.com` references are pre-existing, already-public documentation
entries in `docs/reference/web-sites.md`.

PR #545, opened 2026-09-18
fix(governance): restore consumer governance baseline

wahidyankf:
Cost: one configuration exclusion and seven MIT license files. Benefit: restores current consumer governance checks without changing the RHINO release or pin. Validation: targeted governance, native CI, and diff public-safety passed.


wahidyankf (commented):
Focused leak-only review passed. Scope was limited to secrets/private values, protected environment properties, and machine-specific absolute paths; all other security and semantic concerns are out of scope.

PR #546, opened 2026-09-18
fix(governance): exclude local memsearch

wahidyankf:
## Summary

- Exclude the local `.memsearch` store from repository-wide scanning.
- Preserve the store itself; this change updates only tracked consumer policy.

## Cost / benefit

Cost: one configuration line. Benefit: governance scans remain deterministic when local memory tooling is present.

## Verification

- Guarded `repo-config validate`
- Guarded CI `governance` group
- Guarded full native CI vector


wahidyankf (commented):
Focused pr-leak-review passed for the pinned current head. Scope was limited to secrets/private values, protected environment properties, and machine-specific absolute paths; every other security and semantic concern was out of scope. Counts: secret-or-private-value 0; protected-environment-property 0; machine-specific-absolute-path 0.

wahidyankf (commented):
Focused pr-leak-review passed for the pinned current head. Scope was limited to secrets/private values, protected environment properties, and machine-specific absolute paths; every other security and semantic concern was out of scope. Counts: secret-or-private-value 0; protected-environment-property 0; machine-specific-absolute-path 0.

PR #547, opened 2026-09-18
Document ose-rules adoption by main commit

wahidyankf:
## Outcome

The public-safety fixture documentation now identifies its verified full source commit on the ose-rules published
`main`. It no longer treats an ose-rules release as the transfer boundary.

## Rule propagation

- Normalized rule: describe the adopted fixture by its full source commit on published `main`, while preserving
  byte-identity, upstream-first ownership, and explicit consumer re-adoption.
- Destination: `scripts/README.md`, the existing local owner of the fixture provenance and update path.
- Enforcement: unenforced by decision. Cross-repository provenance is verified during explicit adoption; adding a
  release, pin, drift ledger, or automatic synchronization gate would recreate the rejected coupling.
- Supersession: replaces the single release-based sentence; no instruction entry is evicted.
- Sibling obligation: none. This is ose-public-specific documentation for its own adopted copy.

## Scope and non-goals

Only the active provenance paragraph changes. The public-safety files and their digest/byte-identity contract remain
untouched. Historical plans and commits are not rewritten, and no automatic update mechanism is introduced.

## Proof

- `diff -qr` between both repositories' `scripts/public-safety/` trees at source commit
  `0d38b3e387b0d73e4bff3b8ccafabab6bbce8ccd`
- guarded Prettier check for `scripts/README.md`
- guarded Markdown lint for `scripts/README.md`
- `git diff --check`
- direct outbound preflight for the changed file, branch, commit message, and this pull-request text
- commit and pre-push gate registries

## Reading order

Read the changed paragraph under “The public-safety gate”; the remainder is unchanged context.

## Cost and benefit

There is no new runtime code, dependency, or recurring compute. The benefit is exact, reviewable provenance; the cost is
that a future explicit re-adoption must update the full SHA after byte identity is reverified.

## Risk and rollback

The risk is a stale or mistyped SHA. The pre-edit cross-repository comparison verified the stated commit and byte
identity. Reverting this commit restores the previous wording without changing the fixture itself.

## Trunk state

Merged `main` remains releasable; this documentation-only change has no reachable application or API behaviour.


wahidyankf (commented):
Focused leak-only review: PASS.

Reviewed the complete aggregate diff plus delivery-controlled title and body at the exact pinned head. Findings:

- secret or private value: 0
- protected environment property: 0
- machine-specific absolute path: 0

Every other security and semantic concern was out of scope for this review.

